PR #20, opened 2021-03-12
Don't use target blank for external links

kordianbruck:
Its annoying by breaking navigating back/forward and people should usually use their middle mouse to open a new tab.


joschahenningsen (approved):
Thank you sooo much!!!! ^_^

PR #59, opened 2021-04-17
Cut VoDs into lecture units

joschahenningsen:
todo:

- [x] allow watching different versions (camera, presentation...) of a unit
- [x] allow embedding a unit
- [x] allow disabling original lecture after creating units
- [x] enhance overall ux (especially on unit creation)
- [x] while we're at it: remove legacy player, there were no complaints about the other one


joschahenningsen:
LGTM

PR #74, opened 2021-05-02
Makefile: improve version target

max-ae:


Makefile
```diff
@@ -14,8 +14,7 @@ clean:
 	rm -fr web/node_modules
 
 version:
-	echo -n "hash=" > hash.env; \
-	git log -n 1 --no-merges --pretty=format:%h >> hash.env
+	echo "hash=$$(git rev-parse --short HEAD)\c" > hash.env
```
joschahenningsen: Neat! The server usually merges changes into some environment specific local changes. Can this be modified to skip merge commits?

joschahenningsen (changes requested):

Makefile
max-ae: In that case, the easiest and most elegant way to do this is to use `origin/HEAD` instead of `HEAD`. This assumes a remote reference exists when the command is run, but I guess this is a given in this case.
I updated the PR to reflect this.

max-ae: As a result, the latest revision of this repository will be used, ignoring any local changes.

joschahenningsen (approved):
Awesome, thanks!

PR #76, opened 2021-05-03
Bundle version revision into executable

max-ae:
Followup to #74 

This PR makes changes to bake the revision at build right into the executable instead of using an env-file that could very well change at runtime. 

One question remains: does the server merge local env commits into the local repository before or after the `make bundle` target? If before, `HEAD` in the Makefile again needs to be changed to `origin/HEAD`, otherwise fine as-is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,7 +104,6 @@ web/assets/css-dist
 
 # secrets
 variables-backend.env
-hash.env
 rebuild.sh
 
 tumlive.service
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 all: npm_dependencies go_dependencies bundle
 
+VERSION := $(shell git rev-parse --short HEAD)
+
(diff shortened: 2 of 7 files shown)


joschahenningsen:
Nice, Appreciate it!

> does the server merge local env commits into the local repository before or after the make bundle target? 

Before.

Imo, the version tag variable should live in the tools package in case I need to access it from eg. the dao at some point. (There already is a variable for it in tools.Cfg, I just didn't use it for some reason)

joschahenningsen (approved):
Until I removed dependencies from tools to dao this is perfectly fine :) Thanks for helping out!

max-ae:
Nice! 
I just did not get the chance to change `HEAD` to `origin/HEAD` 😅 As-is, this will now again use the revision of the local merge commit...

joschahenningsen:
Just changed it, thanks for the heads up

PR #84, opened 2021-06-03
web/template/watch.gohtml: Remove player padding-top.

ro-i:
Hi :)

The `pt-20` class adds a rather huge padding-top (5rem) above the player
which is especially inconvenient when using the "big picture mode".
This commit removes the `pt-20` class from the enclosing div.

Feedback would be greatly appreciated, thank you for your work! :heart:


joschahenningsen (approved):
Yikes, messed up with the padding there, thanks for noticing and helping out!
I can deploy this tomorrow

PR #88, opened 2021-06-17
web/template/login.gohtml: Add autofocus for username field.

ro-i:
Hi :)

Can we have `autofocus` for the username input field? It would save one click :)
I think it's supported quite well: https://www.w3schools.com/tags/att_input_autofocus.asp (IE 11?)
And I think there are no [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus) here?

Thanks for reviewing :)

### DIFF
--- a/web/template/login.gohtml
+++ b/web/template/login.gohtml
@@ -15,7 +15,7 @@
             <div class="my-5 text-sm">
                 <label for="username" class="hidden">Username</label>
                 <input type="text" name="username" id="username"
-                       placeholder="LRZ ID (eg. ga10abc) or Email" autocomplete="off" required/>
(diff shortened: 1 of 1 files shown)


joschahenningsen (approved):
Sure thing, my password manager will love this addition! :sparkles: 

> I think it's supported quite well: https://www.w3schools.com/tags/att_input_autofocus.asp (IE 11?)

I'm not concerned about IE 11, this is a website for compsci students, if they decide to use it they're on their own :P (the 0.02% of folks who actually used IE this week were bots probably)

Concerning accessibility: The label for the input is "username" and most people clicked login before accessing this page so it should be fine.

PR #132, opened 2021-10-16
Add ability to resume video at previous time

### LINKED ISSUE #35: Continue playback where you left off & mark videos as seen
TUM-Live should remember where students stopped watching a video and continue there when it's played again. Should be easy using webstorage or by just saving it to the database.

alexanderstephan:
**Description**

This PR enables users to continue watching a VOD where they previously left off, fixing #35. We do this by saving a timestamp using `video.js` and associate it with a user and a stream id. This will be done for every VOD the user has watched.

 To achieve this a new table in the database was created where the corresponding timestamp is saved and also retrieved later on when the video player is loaded initially.
 
 **Tasks**
 
- [x] Implement database scheme and communication via REST calls.
- [x] Retrieve watch progress from database when video player is loaded.
- [x] Set the time in the player.
- [x] Clean up code and names
- [x] Add comments
- [x] Make setting the video time faster if possible 
- [x] Testing
 
**Testing**

I tested everything locally using Firefox 93.0. In a follow up, more formal tests will be introduced, but merging this soon is also important as students can profit from the feature.


api/progress.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"TUM-Live/dao"
+	"TUM-Live/tools"
+	"github.com/gin-gonic/gin"
+	"net/http"
+)
+
+func configProgressRouter(router *gin.Engine) {
+	router.POST("/api/progressReport", saveProgress)
+	router.POST("/api/progressRequest", fetchProgress)
```
joschahenningsen: Semantically this should be a GET request, but I think the progress should be fetched when the site is loaded (`GET /w/GBS/123`) and passed to your videjs plugin through the DOM
This saves an extra HTTP request and there won't be any weird jumps in case the request is delayed.

model/progress.go
```diff
@@ -0,0 +1,7 @@
+package model
+
+type StreamProgress struct {
+	Progress float64
+	StreamID uint `gorm:"primaryKey"`
+	UserID uint `gorm:"primaryKey"`
```
joschahenningsen: It's not entirely clear why this needs to be a primary key (I see why it does but it could use a helpful comment)

web/template/watch.gohtml
```diff
@@ -238,6 +238,10 @@
         }
     });
 
+    {{if $stream.Recording}}
+        player.watchProgress({{$stream.Model.ID}});
```
joschahenningsen: This is where the current progress would be passed

web/ts/TUMLiveVjs.ts
```diff
@@ -121,6 +123,81 @@ const skipSilence = function (options) {
...
+            if (initialized) {
+                return;
+            }
+            this.currentTime(this.progressRatio * this.duration());
+            initialized = true;
+        });
+
+        // iPhone/iPad need to play the video first, so they depend on a different event
+        // More info: https://www.w3.org/TR/html5/embedded-content-0.html#mediaevents
+        this.on("canplaythrough", () => {
```
joschahenningsen: Amazing catch :)

web/ts/TUMLiveVjs.ts
```diff
@@ -121,6 +123,81 @@ const skipSilence = function (options) {
...
+            const duration = this.duration();
+            const progress = ctime / duration;
+
+            postData("/api/progressReport", {
+                "streamID": streamID,
+                "progress": progress
+            }).then(r => {
+                    if (r.status !== 200) {
+                        console.log(r);
+                    }
```
joschahenningsen: It would be nice to exponentially back off when the request fails, this way I can control possible load issues using traefik rate limits 
e.g.:
10 seconds -> request -> fail -> 20 seconds -> fail -> 40 seconds ...

web/template/watch.gohtml
joschahenningsen: Please don't initialize this if the user is not logged in.

dao/progress.go
```diff
@@ -0,0 +1,24 @@
...
+		Columns:   []clause.Column{{Name: "stream_id"}, {Name: "user_id"}}, // key column
+		DoUpdates: clause.AssignmentColumns([]string{"progress"}), // column needed to be updated
+	}).Create(&model.StreamProgress{
+		Progress: progress,
+		StreamID: streamID,
+		UserID:   userID,
+	})
+}
+
+func LoadProgress(userID uint, streamID uint) float64 {
```
joschahenningsen: For good measure this could return an error so the `/w/GBS/123` handler can take care of setting the progress to 0 (if not found) or logging it in case of any other error. See https://gorm.io/docs/error_handling.html#ErrRecordNotFound

joschahenningsen (changes requested):
Overall very nice and works as expected :+1: Thanks!
Just a few nits

api/progress.go
alexanderstephan: This makes sense :+1: Better loading times and less traffic is always good.

dao/progress.go
alexanderstephan: Good idea, this should improve robustness. :+1:

model/progress.go
alexanderstephan: Yes, I also got confused and thought using foreign keys was possible when coming back for refactoring. I will add a comment.

web/ts/TUMLiveVjs.ts
alexanderstephan: Agreed, this should drastically increase robustness.

api/progress.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/gin-gonic/gin"
+	log "github.com/sirupsen/logrus"
+	"net/http"
+)
+
+func configProgressRouter(router *gin.Engine) {
+	router.POST("/api/progressReport", saveProgress)
+}
+
+type ProgressRequest struct {
```
joschahenningsen, line 16: I'm not a good example, but exported fields, methods, functions, and structs should have a comment documenting what they do. (I marked them with a .)

dao/progress.go
```diff
@@ -0,0 +1,24 @@
+package dao
+
+import (
+	"TUM-Live/model"
+	"gorm.io/gorm/clause"
+)
+
+func SaveProgress(progress float64, userID uint, streamID uint) (err error) {
```
joschahenningsen, line 9: .

dao/progress.go
```diff
@@ -0,0 +1,24 @@
...
+		DoUpdates: clause.AssignmentColumns([]string{"progress"}),          // column needed to be updated
+	}).Create(&model.StreamProgress{
+		Progress: progress,
+		StreamID: streamID,
+		UserID:   userID,
+	}).Error
+	return err
+}
+
+func LoadProgress(userID uint, streamID uint) (streamProgress model.StreamProgress, err error) {
```
joschahenningsen, line 23: .

model/progress.go
```diff
@@ -0,0 +1,8 @@
+package model
+
+type StreamProgress struct {
```
joschahenningsen, line 4: .

web/ts/TUMLiveVjs.ts
```diff
@@ -121,6 +118,75 @@ const skipSilence = function (options) {
...
+        // Triggered on pause and skipping the video
+        this.on('pause', () => {
+            reportProgress(this.currentTime());
+            clearInterval(timer);
+        })
+
+        // Triggered when the video has no time left
+        this.on('ended', () => {
+            // Maybe we want to set a flag here in the future to mark the video as watched
+            reportProgress(0);
```
joschahenningsen: Let's not reset the progress here, I'd like to keep it (even if it approaches 100%) to indicate that the video is watched through. This way we can mark them as viewed in the frontend or let users filter their lectures in the future (perhaps something for a follow-up?)

`setProgress()` would then need to check the progress and start at 0 if it's close to the end (>95% maybe?)

joschahenningsen (changes requested):
Amazing, thanks for applying my suggestions, now starting a video somewhere in the middle feels really smooth :) 
I'm just nit-picking at this point but other than the last comments this is ready.

joschahenningsen (approved):
Amazing, works great. Thanks a lot! :partying_face:

PR #134, opened 2021-10-18
Detect Light/Dark Theme

### LINKED ISSUE #133: Toggle Light/Dark Mode Based on System Theme
Many websites and apps [1] have added the ability to not only switch manually between a dark and light theme, but also to adapt automatically to the system theme. TUM-Live should add this as an option.

**Solution**
The existing theme-switcher button should be extended to cycle between light mode, dark mode and auto mode (follow system theme). In light or dark mode, the behaviour should be as before. In auto mode, TUM-Live should check the preferred theme upon loading and then listen for updates, updating the displayed theme as needed.

The following issue describes adding the same feature to VSCode, the link points specifically to a comment suggesting a clean way of implementing theme preference: https://github.com/microsoft/vscode/issues/61519#issuecomment-558738593. The solution bases off of using `window.matchMedia('(prefers-color-scheme: dark)')`

**More info**
- Chromium-based snap apps, eg. Google Chrome on Ubuntu, don't update `prefers-color-scheme` properly (see https://bugs.chromium.org/p/chromium/issues/detail?id=998903). Automatic switching would not work in this case.

- Documentation/examples: https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Testing_media_queries

**Implementation**
I've gone ahead and implemented this, PR incoming.

***
[1] examples: github, VSCode

just-max:
Add support for automatically switching between light and dark modes based on preferred colour scheme as reported by browser.
Resolves #133.

Behaviour:
- default to dark theme with no automatic switching
- the scheme switcher cycles between light, dark, and auto modes
- the scheme switcher shows which mode is active by displaying a sun (light mode), moon (dark mode) or [adjust](https://fontawesome.com/v5.15/icons/adjust?style=solid) (auto mode) icon

Testing:
I've "tested" this by making the appropriate changes in the Firefox developer tools... Is it even possible to run a local instance without the secrets?

Next steps:
It might be worth checking the browser support for the MediaQuery API. Specifically, `MediaQueryList.addEventListener` isn't supported in IE and would need to be replaced by `MediaQueryList.addListener`, which is deprecated in most browsers.

### DIFF
--- a/api_grpc/worker.go
+++ b/api_grpc/worker.go
@@ -16,8 +16,10 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/protobuf/types/known/timestamppb"
+	"io/ioutil"
 	"net"
 	"net/http"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


jpbernius:
I would very much like this, but I would vote to use the system setting as a default, not a dark theme.

just-max:
Yeah, that would just be a matter of changing one boolean. I did it this way to make the default behaviour the same as before but I would also prefer auto as default.

just-max:
For now, I wrote something similar into a userscript: https://greasyfork.org/en/scripts/444353-tum-live-auto-dark

joschahenningsen:
Hi @just-max 

I'm super sorry I let this go stale :/ I'd still like to merge this feature but I'm afraid this PR is a little outdated by now. I had a look at https://tailwindcss.com/docs/dark-mode#toggling-dark-mode-manually, imo their code is very clean and i love the look of the preference switch in the top right corner. 

image

Also they recommend putting the theme related code into the DOM to avoid flickering when the page is loaded. 

If you want you can pick this up, otherwise I'll implement it :)

just-max:
Hey! No worries, it was just a feature I found myself missing :)

I could put something together.
- For the dropdown I would adapt the notification dropdown (for consistency)
- FA Icons: I would use `sun`, `moon`, and `desktop`
- I like how simple the implementation is too, but it means that anyone who has used the site before will not benefit from the automatic theme changing. I would suggest changing the local storage key to something else (`theme`) and migrating appropriately.

joschahenningsen:
Sounds great! :100:

PR #153, opened 2021-11-23
Improve model and dao

alexanderstephan:
This is a small PR that contains a bunch of little improvements to the database or the dao.
- Introduces usage of gorms [Count](https://gorm.io/docs/advanced_query.html#Count).
- Removes code duplication in `GetStudentActivityCourseStats` in `dao/statistics.go` by using a boolean argument.
- Uses MySQL `DATE_FORMAT` to avoid manually building date strings.
- Introduces a `not null` constraint where it made sense to me. Of course, this is up to debate and can and should be adjusted during review.
- Moves time calculation that do not depend on state in the database query itself. This also avoids some fault prone time calculations.
- Adds additional, hopefully sensible default values where possible.


model/chat.go
```diff
@@ -7,10 +7,11 @@ import (
...
-	StreamID uint
-	Admin    bool
-	SendTime time.Time
+
+	UserID   string    `gorm:"not null"`
+	UserName string    `gorm:"not null"`
+	Message  string    `gorm:"not null"`
+	StreamID uint      `gorm:"not null"`
+	Admin    bool      `gorm:"not null"`
+	SendTime time.Time `gorm:"not null"`
```
joschahenningsen: While you're at it, can you remove sendTime and check that we just use gorms `createdAt`?

model/chat.go
```diff
@@ -7,10 +7,11 @@ import (
...
 type Chat struct {
 	gorm.Model
-	UserID   string
-	UserName string
-	Message  string
-	StreamID uint
-	Admin    bool
-	SendTime time.Time
+
+	UserID   string    `gorm:"not null"`
```
joschahenningsen: Did you check if gorm applies these changes to an existing database? What would happen with `null` values?
If not we might need a migration. I did something similar here: https://github.com/TUM-Dev/Campus-Backend/tree/main/server/backend/migration If you want to check it out. 
It basically creates another table that lists which migrations have been applied to the database and runs them in case they are new. Reference: https://github.com/go-gormigrate/gormigrate

model/stream-unit.go
```diff
@@ -10,11 +10,12 @@ import (
 
 type StreamUnit struct {
 	gorm.Model
-	UnitName        string
+
+	UnitName        string `gorm:"not null"`
```
joschahenningsen: I can see this being `null`, most lecturers don't bother thinking about naming their lectures.

model/stream.go
```diff
@@ -11,21 +11,21 @@ import (
 type Stream struct {
 	gorm.Model
 
-	Name             string
+	Name             string `gorm:"not null"`
```
joschahenningsen: This could be null as well

joschahenningsen (changes requested):
Nice improvements! ~~Most of the fields that you annotated with `not null` should have a default value though as they currently fall back to null (usually interpreted as 0 or false in gorm)~~ Edit: should be fine if some of those queries would fail. Just a few defaults might be required.

model/chat.go
alexanderstephan: Sure, no problem!

model/stream-unit.go
alexanderstephan: Ok, will remove this then. :+1:

model/chat.go
alexanderstephan: Yes, the updates applied automatically as there is an automigration happening when starting up the server. Unfortunately, it cannot resolve some conflicts, e.g. when a column that was annotated with non-null still contains null values this causes error. If we want to delete incorrect null entries, we will have to do it manually.

model/worker.go
```diff
@@ -6,7 +6,7 @@ type Worker struct {
 	WorkerID string `gorm:"primaryKey"`
 	Host     string
 	Status   string
-	Workload int // How much the worker has to do. +1 per silence detection job, +2 per converting job, +3 per streaming job
+	Workload uint // How much the worker has to do. +1 per silence detection job, +2 per converting job, +3 per streaming job
```
joschahenningsen, line 9: This requires a change in 
https://github.com/joschahenningsen/TUM-Live/blob/02400f7406905c59c12731369edc1283e25c19ce/api_grpc/worker.go#L189

api/chat.go
```diff
@@ -53,7 +53,13 @@ func configGinChatRouter(router *gin.RouterGroup) {
 		if chat.Msg == "" || len(chat.Msg) > maxMessageLength {
 			return
 		}
-		if dao.IsUserCooledDown(fmt.Sprintf("%v", tumLiveContext.User.ID)) {
+		isCooledDown, err := dao.IsUserCooledDown(fmt.Sprintf("%v", tumLiveContext.User.ID))
+		if err != nil {
+			sentry.CaptureException(errors.New("could not determine whether user has cooled down"))
+			ctx.(*gin.Context).AbortWithStatus(http.StatusInternalServerError)
```
joschahenningsen: This could kill the WebSocket. I think, in case of an error we should either just return or wait for #173 to be merged and send back a "something went wrong" message. 
But since this should be a very rare edge case a return should be fine

api/chat.go
```diff
@@ -53,7 +53,13 @@ func configGinChatRouter(router *gin.RouterGroup) {
 		if chat.Msg == "" || len(chat.Msg) > maxMessageLength {
 			return
 		}
-		if dao.IsUserCooledDown(fmt.Sprintf("%v", tumLiveContext.User.ID)) {
+		isCooledDown, err := dao.IsUserCooledDown(fmt.Sprintf("%v", tumLiveContext.User.ID))
+		if err != nil {
+			sentry.CaptureException(errors.New("could not determine whether user has cooled down"))
```
joschahenningsen: Logrus can log the error and send an event to sentry as well (if configured).

```suggestion
			log.WithError(err).Error("could not determine whether user is cooled down")
```

dao/server-notification.go
```diff
@@ -33,6 +34,10 @@ func DeleteServerNotification(notificationId string) error {
...
 //CreateServerNotification creates a new ServerNotification
 func CreateServerNotification(notification model.ServerNotification) error {
-	err := DB.Create(&notification).Error
-	return err
+	if notification.Expires.Before(notification.Start) {
+		return errors.New("expiration before start is invalid")
+	} else {
+		err := DB.Create(&notification).Error
+		return err
+	}
```
joschahenningsen: Could this be a [BeforeCreate hook](https://gorm.io/docs/hooks.html)?

dao/statistics.go
```diff
@@ -68,25 +68,14 @@ func GetCourseStatsHourly(courseID uint) ([]Stat, error) {
...
-		FROM stats
-        	JOIN streams s ON s.id = stats.stream_id
-		WHERE (s.course_id = ? OR ? = 0) AND stats.live = 0
-		GROUP BY year(stats.time), week(stats.time)
-		ORDER BY stats.time;`,
-			courseID, courseID).Scan(&res).Error
-		return res, err
-	}
+		courseID, courseID, live).Scan(&res).Error
+	return res, err
```
joschahenningsen, line 78: That's way better. What did I think here? :smile:

joschahenningsen (changes requested):
Amazing, just a few last notes, then we're good to go

api/chat.go
alexanderstephan: Ah, I always forget about that function... will change that!

api/chat.go
alexanderstephan: I just merged with #173 from `dev`, we should be good to go now.

joschahenningsen (approved):
Tested this on a copy of the prod database and everything looks good :) great work, could you just resolve the conflict with `dev` please?

joschahenningsen:
Migration worked flawlessly, thanks again!

PR #154, opened 2021-11-23
Introduce staticcheck as a static analysis tool

alexanderstephan:
**Description:**

This PR adds a new static analysis tool, adds pre-commit hooks and fixes some other small code issues.

**Changes**

- Adds Github action for [staticcheck](https://staticcheck.io/docs/running-staticcheck/github-actions/). It is pretty powerful and probably the best static analysis tool for golang at the moment. Right now `ST1000` (Incorrect or missing package comment) is disabled as adding comments for each exported method is out of scope. `ST1003` (Poorly chosen identifier) is also disabled and might enforce a different naming style (e.g. Id -> ID).
- I also added that `dependabot`  now also updates GitHub actions.
-  I also added pre-commit hook configuration  that runs `staticcheck`, `gofmt` (go's built-in formatter) and `goimports`. This can be used when developing on a branch to assure code quality before creating an actual PR.
- I fixed all issues that were discovered by using the newly introduced tools.
- I added some additional small things while browsing the code like introducing constants for magic numbers.

**Testing:**

I tested the hooks locally, compared the GitHub action result with local results and tested whether the server still works.


joschahenningsen (approved):
Wow, there really are a few funny lines in my code :D
Thanks a lot!

api/chat.go
```diff
@@ -15,13 +15,15 @@ import (
...
 )
 
 var m *melody.Melody
-var stats = map[string]int{}
-var statsLock = sync.RWMutex{}
+
+const maxParticipants = 10000
+const maxMessageLength = 200
+
+// TODO: Use stats
```
joschahenningsen: We don't need this todo, stats are now inferred from wsHub.go

alexanderstephan: Ah, that makes sense. I'll remove it :+1:

PR #168, opened 2021-12-04
Refactor and lint typescript

alexanderstephan:
This PR is based partly on https://github.com/joschahenningsen/TUM-Live/pull/155, which it is not likely to get merged at the moment as the front end framework is probably going to change soon.

It adds the following stylistic improvements (not exhaustive):
- Consistent usage of quotes
- Always use semi colons at the end of a line
- Use `const` whenever possible
- Consistent use of spaces
- Enforces new lines at end of files
- I changed some names to be more expressive.

It also incorporates some refactorings  (not exhaustive): 
-  E.g. `setIntervall` is also used  for the `skipSilence` Plugin now.
- I extracted expressions when the line length limit was reached.
- In `scheduling.ts` I used an arrow function in line 72 so we don't have to alias `this`.

Eslint was extended by plugging in prettier. I had to disable some linting rules due to imports not working and usage of the code within the HTML, which is not detected right now. The linter now runs via GitHub actions as well.


joschahenningsen (approved):
This is so much better, thanks! :100: I hope a lot of this was done automatically :D

joschahenningsen (approved):
Amazing!

If you enable eslint in goland you should get a file `.idea/jsLinters/eslint.xml`[1]. Commiting this would be nice as well

[1]:
```
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="EslintConfiguration">
    <option name="fix-on-save" value="true" />
  </component>
</project>
```

alexanderstephan:
Ah, I didn't even know that goland has eslint integration, thanks for the recommendation! I will definitely add this :+1:

joschahenningsen (approved):
:heart:

PR #170, opened 2021-12-05
Make streams visible that are starting soon

alexanderstephan:
**Motivation**:
I experienced that lecture that use OBS start the lectures just in time or even after they should start. By providing a visual representation which streams start soon along with a corresponding link to waiting room, it's much more comfortable to open the stream. Otherwise one needs to refresh the page until the stream becomes visible.

**Design:**
The start page has now an overview of today's lectures, where links to the waiting rooms are highlighted and linked accordingly.
I went for a minimalistic approach now.
image

 Right now the individual course site shows upcoming streams.

The state is also visible in the course overview and is also a link to the waiting room.

image

**I would be happy about feedback regarding the design.** :grin:


model/stream.go
```diff
@@ -45,6 +45,11 @@ func (s Stream) IsPast() bool {
 	return s.End.Before(time.Now())
 }
 
+// IsComingUp returns whether the stream begins within an hour
+func (s Stream) IsComingUp() bool {
+	return !s.IsPast() && s.Start.Before(time.Now().Add(time.Hour))
```
joschahenningsen: I think 30 minutes should be enough

web/assets/css/main.css
```diff
@@ -46,6 +46,10 @@
         height: 3px;
     }
 
+    /* Waiting room pill color */
+    .bg-wait {
+        background-color: darkorange;
+    }
```
joschahenningsen: We should consistently use tailwinds coloring system: https://tailwindcss.com/docs/customizing-colors

model/stream.go
joschahenningsen: A lecture should also be coming up when the start date is passed, the end date isn't and the stream is not yet live (e.g. prof is late).
Also, this function could check whether the stream is live or a recording (recordings can be in the future for preuploaded videos)

model/course.go
```diff
@@ -69,6 +69,16 @@ func (c Course) IsLive() bool {
 	return false
 }
 
+// IsNextLectureStartingSoon checks whether the course has a lecture that starts soon
+func (c Course) IsNextLectureStartingSoon() bool {
+	for _, s := range c.Streams {
```
joschahenningsen, line 77: You can use the courses `getNextLecture` method I think

joschahenningsen (changes requested):
This is a great idea!

You iterate over all courses and a couple of their streams a few times (when preparing the page and when rendering it). 
Since the courses on the start page are already ordered by date we can avoid another section IMO and be a little more subtle by displaying a link to the upcoming lecture next to the course. This way we could just use the one loop in the template.

How does something like this look? (perhaps a little more colorful/eye-catching?)
image

The waiting-room page contains a timer that counts down to 10 minutes before the stream is scheduled (that's when it actually starts). This needs to be disabled for self streamers, as they can start whenever they want.

web/assets/css/main.css
alexanderstephan: Okay, I just liked that the orange is really vibrant here, but e.g. #F59E0B from the palette also seems really good.

model/stream.go
alexanderstephan: 30 Minutes should be plenty :+1:  The idea was to have a better overview of todays lectures in general, but that should be probably done differently.

Ah, I didn't think about pre-uploaded videos, thanks for the hint.

model/course.go
alexanderstephan: Ah, that's way cleaner!

alexanderstephan:
I was not sure if iterating over course lists has an performance impact here. Putting the text in the existing course overview seems cleaner to me, but I was unsure about the visibility. Like you said we definitely need to add some colors in this case.

I initially had a timer like this, but then it would not be really possible with `gohtml` as we want the timer to update.

web/assets/css/main.css
joschahenningsen: Feel free to extend the palette :)

joschahenningsen:
> I was not sure if iterating over course lists has an performance impact here

Nothing significant, but whatever we can avoid is surely not bad :shrug: 

> I initially had a timer like this, but then it would not be really possible with gohtml as we want the timer to update.

Do you mean the "in x minutes" timer? In that case this https://codepen.io/harsh/pen/KKdEVPV could be cool :)

model/course.go
alexanderstephan: `getNextLecture` was broken, but it now should be fixed

alexanderstephan:
Yeah, I meant the minutes timer. I have no updated everything for the new timer. It loads now way faster than before. :tada:

.idea/jsLibraryMappings.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="JavaScriptLibraryMappings">
-    <file url="PROJECT" libraries="{jquery}" />
+    <file url="PROJECT" libraries="{jquery}"/>
+    <includedPredefinedLibrary name="Node.js Core"/>
```
joschahenningsen:
```suggestion
```

model/stream.go
```diff
@@ -41,10 +41,30 @@ type Stream struct {
 	StreamName       string
 }
 
+// IsRegularStream returns whether the stream is a scheduled stream in a lecture hall
+func (s Stream) IsRegularStream() bool {
```
joschahenningsen: I'd call this something like `isNotSelfStream`

web/assets/css/main.css
```diff
@@ -236,6 +236,15 @@ input[type='checkbox']:checked + label {
...
 
+.timer {
+    display:flex;
+    font-variant-numeric: tabular-nums; /* Otherwise the counter will "jump" */
+}
+
+.timer span + span:before {
+    content:":"
+}
+
```
joschahenningsen: You can use the classes `flex`, `tabular-nums` and `after:content-[':']`(?) (https://tailwindcss.com/docs)

web/template/course-card.gohtml
```diff
@@ -1,21 +1,29 @@
 {{define "course-card"}}{{- /*gotype: TUM-Live/model.Course*/ -}}
 <div class="pt-1 group course{{.Model.ID}}">
     <div class="flex items-center">
-        <i class="fancyeye opacity-0 group-hover:opacity-100 transition-colors duration-200 dark:hover:text-white hover:text-black text-gray-500 mr-2 cursor-pointer"
+        <i class="fancyeye opacity-0 group-hover:opacity-100 transition-colors duration-200 dark:hover:text-whitehover:text-black text-gray-500 mr-2 cursor-pointer"
```
joschahenningsen:
```suggestion
        <i class="fancyeye opacity-0 group-hover:opacity-100 transition-colors duration-200 dark:hover:text-white hover:text-black text-gray-500 mr-2 cursor-pointer"
```

web/ts/global.ts
```diff
@@ -105,6 +105,64 @@ function initHiddenCourses() {
     }
 }
 
+// Adapted from https://codepen.io/harsh/pen/KKdEVPV
+function timer(expiry: string, regular: boolean, leadingZero: boolean) {
+    return {
+        expiry: new Date(expiry),
+        remaining: null,
+        init() {
+            const offsetInMinutes = regular ? 10 : 0;
```
joschahenningsen: For some reason, the timer on the start page has an offset of 20 minutes instead of 10, any idea why? :thinking:

web/template/course-card.gohtml
```diff
@@ -1,21 +1,29 @@
...
                 Live now.
-            {{else}}
-                {{if .HasNextLecture}}
-                    {{$next := .GetNextLectureDate}}Next Lecture: {{printf "%v %02d. %02d:%02d" $next.Month $next.Day $next.Hour $next.Minute}}
+            {{else if .HasNextLecture}}
+                {{if .IsNextLectureStartingSoon}}
+                    <div class="timer text-5 font-sans font-light text-sm" x-data="timer('{{.GetNextLectureDateFormatted}}', '{{.GetNextLecture.IsRegularStream}}', false)" x-init="init();">
+                        In&nbsp;<span x-text="time().minutes"></span>&nbsp;Minutes •&nbsp;
+                    </div>
+                    <a class="text-5 font-sans text-sm" href="/w/{{.Slug}}/{{.GetNextLecture.Model.ID}}">
```
joschahenningsen:
```suggestion
                    <a class="text-5 font-sans text-sm underline" href="/w/{{.Slug}}/{{.GetNextLecture.Model.ID}}">
```

web/template/course-card.gohtml
```diff
@@ -1,21 +1,29 @@
...
+            {{/*Order of conditions is important*/}}
             {{if .IsLive}}
                 Live now.
-            {{else}}
-                {{if .HasNextLecture}}
-                    {{$next := .GetNextLectureDate}}Next Lecture: {{printf "%v %02d. %02d:%02d" $next.Month $next.Day $next.Hour $next.Minute}}
+            {{else if .HasNextLecture}}
+                {{if .IsNextLectureStartingSoon}}
+                    <div class="timer text-5 font-sans font-light text-sm" x-data="timer('{{.GetNextLectureDateFormatted}}', '{{.GetNextLecture.IsRegularStream}}', false)" x-init="init();">
+                        In&nbsp;<span x-text="time().minutes"></span>&nbsp;Minutes •&nbsp;
```
joschahenningsen: If the stream should already have started but didn't, this should display something like "starting soon" instead of "in -10 minutes".
Also if there's only one minute left it would be nice if this was "in 1 minut**e**" (no plural)

model/course.go
```diff
@@ -80,33 +93,35 @@ func (c Course) NumStreams() int {
...
-			earliestLecture = s.Start
-		}
-	}
-	return earliestLecture
-}
-
-func (c Course) GetNextLecture() *Stream {
+// GetNextLecture returns the next lecture of the course
+func (c Course) GetNextLecture() Stream {
+	var earliestLecture Stream
```
joschahenningsen, line 108: Not sure if this is related to this change, but on the start page I now see courses without a next lecture on the top: 
image

joschahenningsen (changes requested):
Very nice! I like the design, just a few smaller remarks :)

model/stream.go
alexanderstephan: The name is not ideal. I think it would be even better to invert the logic here and call `isSelfStream` instead. Otherwise it gets confusing when we negate the variable again.

web/assets/css/main.css
alexanderstephan: Thanks, the first two work. However, `after:content-[':']` does not work from my testings. My solution would be to just insert the `:` manually in the template, I think using CSS for this is not necessary.

web/template/course-card.gohtml
alexanderstephan: Thanks, good catch! :)

web/ts/global.ts
alexanderstephan: Not really, but moving the time calculation outside of the `init` function somehow solves it.

web/template/course-card.gohtml
alexanderstephan: Right, that should make it feel more clickable :+1:

web/template/course-card.gohtml
alexanderstephan: Oh yeah, forgot about the singular case :see_no_evil: I thought it could be useful to see the overdue time, but I agree that its pretty ugly this way, I will fix that.

model/course.go
alexanderstephan: It only works with the hold `GetNextLectureDate` somehow and for me it only happens for the Public Courses. Maybe I can find the bug, but otherwise we can just stick with old method.

alexanderstephan: There seems to be a sorting issue, possibly related with the database retrieval. I stuck with the existing method for now.

joschahenningsen (approved):
I'm super happy with this, thanks! :)

PR #173, opened 2021-12-06
Server chat notifications

joschahenningsen:
When logging in to a chatroom a user should receive a welcome message that reminds them to stay on topic and be nice. Also there should be another message indicating that lecturers turned off anonymous Chats.

image

### DIFF
--- a/api/wsHub.go
+++ b/api/wsHub.go
@@ -48,7 +48,7 @@ var connHandler = func(s *melody.Session) {
 	if tumLiveContext.Course.ChatEnabled {
 		sendServerMessageWithBackoff(s, uid, tumLiveContext.Stream.ID, "Welcome to the chatroom! Please be nice to each other and stay on topic if you want this feature to stay active.", TypeServerInfo)
 	}
-	if !tumLiveContext.Course.AnonymousChatEnabled {
+	if !tumLiveContext.Course.AnonymousChatEnabled && tumLiveContext.Course.ChatEnabled {
 		sendServerMessageWithBackoff(s, uid, tumLiveContext.Stream.ID, "The broadcaster disabled anonymous messaging for this stream.", TypeServerWarn)
 	}
 }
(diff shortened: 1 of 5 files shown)


api/wsHub.go
```diff
@@ -34,6 +41,48 @@ var connHandler = func(s *melody.Session) {
...
 		log.WithError(err).Error("can't write initial stats to session")
 	}
+	var uid uint = 0
+	if tumLiveContext.User != nil {
+		uid = tumLiveContext.User.ID
+	}
+	if tumLiveContext.Course.ChatEnabled {
+		sendServerMessageWithBackoff(s, uid, tumLiveContext.Stream.ID, "Welcome to the chatroom! Please be nice to each other and stay on topic if you want this feature to stay active.", TypeServerInfo)
+	}
+	if !tumLiveContext.Course.AnonymousChatEnabled {
```
alexanderstephan: Not sure if this can happen, but it could prevent some weird behavior in case of errors.
```suggestion
	if !tumLiveContext.Course.AnonymousChatEnabled && tumLiveContext.Course.ChatEnabled {
```

tools/cache.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/dgraph-io/ristretto"
+	"time"
+)
+
+var cache *ristretto.Cache
+
+func initCache() {
+	c, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 1e6,     // number of keys to track frequency of (1M).
+		MaxCost:     1 << 29, // 1 << 30 == 1/2GB Cost of cache
```
alexanderstephan: Shouldn't this refer to `MaxCost`?
```suggestion
		MaxCost:     1 << 29, // 1 << 29 == 1/2GB Cost of cache
```

alexanderstephan (changes requested):
Looking really good! :rocket: Just some minor comments.

joschahenningsen:
Nice catches, thanks!

alexanderstephan (approved):

PR #180, opened 2021-12-09
Variable ingest server output url

joschahenningsen:
This enables us to specify where the hls stream is located depending on the server the stream is ingested to. 

This makes test environments possible and is the first step to streaming on an independent infrastructure

Related worker pr: https://github.com/joschahenningsen/TUM-Live-Worker-v2/pull/4


joschahenningsen:
just merged https://github.com/joschahenningsen/TUM-Live-Worker-v2/pull/4 This should be good to go

api_grpc/worker.go
```diff
@@ -447,9 +449,15 @@ func notifyWorkersPremieres() {
 		}
 		workerIndex := getWorkerWithLeastWorkload(workers)
 		workers[workerIndex].Workload += 3
+		ingestServer, err := dao.GetBestIngestServer()
+		if err != nil {
+			log.WithError(err).Error("Can't find ingest server")
```
alexanderstephan, line 454: Maybe we should stop here when an error happens?

alexanderstephan (changes requested):
Looks really good, seems to match the implementation in the worker. :+1:  Just a minor comment.

api_grpc/worker.go
joschahenningsen: You're right, we absolutely should. Pushed a fix

alexanderstephan (approved):
Looks perfect to me now! :+1:

PR #183, opened 2021-12-12
use viper & yaml for configurations

joschahenningsen:


joschahenningsen:
ping @mono424 @MatthiasReumann @alexanderstephan With this you should be able to remove the environment variables from your run configuration.

alexanderstephan (approved):
Tested out locally, works perfectly! :rocket:  Code changes look also really good :+1:

PR #187, opened 2021-12-19
End streams early

### LINKED ISSUE #171: End streams early
Often, lecturers either end their lectures early or don't show up. The streams continues nonetheless, this eats resources (computing power and disk space) 

Every live stream needs a button for admins and lecturers that allows them to end the stream. This should include a prompt whether or not to discard the recording as well (in case no one was there we don't need to create a vod.)

alexanderstephan:
This PR resolves #171 and enables lecturers and admins to end streams in a lecture hall before the official deadline. Before, streams would be active until the next stream in the lecture has scheduled, this takes up resources that could be used better elsewhere.

### Changes:

- Update the design of the admin stream buttons
- The stop button sends a request to the backend via REST api, which then notifies the worker via grpc
- Streams now have a ended state
- With this ended state the watch page was a bit refined by using more dynamic display of a timer, e.g. it's now detected when a stream ends and automatically switches to the timer when within 24h frame preceding the start of stream.
- When ending a stream there is the option to discard the VoD, in this case there will be now transcoding and publishing which also saves time, in case the VoD is not needed.

###  Before merging:

- [x] If https://github.com/joschahenningsen/TUM-Live-Worker-v2/pull/5 gets merged, we should restore `go.mod` and `go.sum` to point to the right branch. I had to change because this PR depends on a specific branch in the worker.

## Screenshots:

**The new admin buttons**:

image

**The prompt for discarding the VoD**:

image


api_grpc/worker.go
```diff
@@ -315,7 +317,7 @@ func (s server) NotifyStreamStarted(ctx context.Context, request *pb.StreamStart
 		default:
 			dao.SaveCOMBURL(&stream, request.HlsUrl)
 		}
-		api.NotifyViewersLiveState(stream.Model.ID, true)
+		//api.NotifyViewersLiveState(stream.Model.ID, true)
```
joschahenningsen: Bummer, I guess we could move this grpc stuff into the API package to resolve this.

api_grpc/worker.go
```diff
@@ -419,6 +421,7 @@ func NotifyWorkers() {
 			}
 			workerIndex := getWorkerWithLeastWorkload(workers)
 			workers[workerIndex].Workload += 3
+			workersForStream[streams[i].ID] = append(workersForStream[streams[i].ID], workers[workerIndex])
```
joschahenningsen: This needs to be persisted in the database. Otherwise, streams can't be stopped if TUM Live is restarted during a stream.

api_grpc/worker.go
```diff
@@ -440,6 +443,49 @@ func NotifyWorkers() {
...
+		resp, err := client.RequestStreamEnd(context.Background(), &req)
+
+		// Was ending the stream successful?
+		if err != nil || !resp.Ok {
+			log.WithError(err).Error("Could not end stream!")
+			_ = conn.Close()
+			return
+		}
+
+		_ = conn.Close()
```
joschahenningsen: closing the connection shoud be in a defer statement like this

```go
defer func(c *grpc.ClientConn) {
	if err := c.Close(); err != nil {
		log.WithError(err).Error("Could not close connection to worker")
	}
}(conn)
```

Since those should be avoided in a loop, the "send request" functionality should be in its own method.

(I might have been a bad example before :P)

api_grpc/worker.go
```diff
@@ -440,6 +443,49 @@ func NotifyWorkers() {
...
+
+	}
+	// All workers for stream are assumed to done
+	// TODO: Remove individual workers in NotifyStreamFinished?
```
joschahenningsen: Assuming all are done is just fine :) But again, this should happen in the database.

web/template/about.gohtml
```diff
@@ -20,12 +20,13 @@
     <div class="border-gray-500 border-t-2 mt-4 mb-4"></div>
     <h1 class="text-1">Übertragung von Veranstaltungen</h1>
     <p>
-    Wir übertragen gern alle Veranstaltungen, die sich an TUM-Studierende oder Mitglieder richten.<br>
-    Hier gibt es zwei Möglichkeiten:
+        Wir übertragen gern alle Veranstaltungen, die sich an TUM-Studierende oder Mitglieder richten.<br>
```
joschahenningsen: There seem to be some commits from another pr, can you merge `dev`?

web/template/watch.gohtml
```diff
@@ -115,16 +120,25 @@
...
+                    <a href="?restart=1" title="Issues with the stream? Restart it."
+                       class="bg-indigo-500 inline-block text-center text-white w-24 font-bold hover:bg-indigo-700 rounded cursor-pointer p-2 mr-3 text-sm">
+                        <i class="fas fa-redo mr-1"></i>Restart
+                    </a>
+                    <a @click="fetch(`/api/stream/${streamID}/end`)" href="#"
+                       title="Already finished? End the stream early."
+                       class="bg-red-500 inline-block text-center w-24 text-white hover:bg-red-700 font-bold rounded cursor-pointer p-2 text-sm">
+                        <i class="fas fa-power-off mr-1"></i>Stop
+                    </a>
+                    <p class="text-4 my-3 font-bold">Presets</p>
```
joschahenningsen: this looks nice :fire:

joschahenningsen (changes requested):
Looks good so far :) 
Just a few thoughts

api_grpc/worker.go
alexanderstephan: I had the same idea, but I was not sure about it. I guess a package for one file is not optimal anyways.

api_grpc/worker.go
alexanderstephan: I only wanted to get it working right now, once it works on the test server I will adapt this.

api_grpc/worker.go
alexanderstephan: Ah, right. I tend to forget using `defer` :see_no_evil:

web/template/about.gohtml
alexanderstephan: I think it was some stashing issue unrelated to the merge, I already just fixed it locally. I will push it now!

joschahenningsen:
The action problem seems to be because of the merge conflicts: https://github.community/t/run-actions-on-pull-requests-with-merge-conflicts/17104

This is super stupid, maybe on push behaves differently?

alexanderstephan:
Yes, on push seems to be the only solution. This is kind of a bummer because I can't avoid a merge conflict here due to the changed worker branch in `go.mod`. Maybe we can change it, at least until this PR is merged?

joschahenningsen:
I'm sorry for causing a merge conflict here, you can just copy the changes from a03313d43e0171fb5b6fefbc975a108ee73db4ad into the new worker grpc file.

alexanderstephan:
No problem, I just merged again!

joschahenningsen:
This looks good, I'd like to merge it later :tada: 
I just merged your other PR and tagged https://github.com/joschahenningsen/TUM-Live-Worker-v2 with `v1.0.0`, can you update the version in `go.mod`?

joschahenningsen (approved):

alexanderstephan:
Ok, I think this ready now!

joschahenningsen:
can you run another `go mod tidy`? then we're ready to go :)

joschahenningsen:
Amazing :partying_face:

PR #193, opened 2021-12-20
Stream Management - Perform bulk actions (delete)

### LINKED ISSUE #190: Add "bulk actions" to stream management
Editing streams can currently only be done one stream at a time. It would be nice to have a checkbox next to every stream in the admin tab and be able to apply bulk actions. 
E.g. or now: select multiple streams and click on a `delete selected` button instead of deleting them one by one.

**Todo:**
- [ ] #194
- [ ] Update titles
- [ ] Change lecture hall
- [ ] change start/end time
- [ ] move to day of week

@mono424 this might be something for now because streaming on the test server is not yet possible :)

mono424:
Enabled selecting multiple lectures/videos to delete them all at once.


joschahenningsen:
Looks good, surprisingly little code as well 🥳 thanks! 
Can we make the delete lectures endpoint accept an array of ids in it's body instead of just one id from the url? Then we'd avoid spamming a bunch of consecutive requests for essentially one job.
Also I would expect the checkboxes to be on the left hand side, but that's just a matter of taste.

mono424:
Actually thought already about both, endpoint makes totally sense! I will add that! Further I like the checkboxes aligned to the delete button as you can see better they are related, but can also move them to the left as you wish :)

mono424:
Isn't. it wired if the checkbox is somehow directly below the filter checkbox even though they have no link? 
Bildschirmfoto 2021-12-21 um 12 34 43

api/courses.go
```diff
@@ -353,21 +353,42 @@ type renameLectureRequest struct {
...
+		stream, err := dao.GetStreamByID(context.Background(), streamID)
+		if err != nil || stream.CourseID != tumLiveContext.Course.ID {
+			c.AbortWithStatus(http.StatusForbidden)
+			return
+		}
+		streams = append(streams, stream)
+	}
+
+	for _, stream := range streams {
+		stream.Model.DeletedAt = gorm.DeletedAt{Time: time.Now()} // todo ?!
```
joschahenningsen: You don't need to set this, that's what gorm internally does when you call `Delete()`

joschahenningsen (changes requested):
This looks great :+1: one bulk request for a bulk actions seems like the way to go :)
How about moving the filter to the right and the bulk action (checkbox and button) to the left? 

But other than the comment on gorm, this looks good!

api/courses.go
mono424: Nice thanks, I will make the changes ASAP! :)

mono424: What do you think now ? :)
Bildschirmfoto 2021-12-31 um 16 48 37

joschahenningsen:
Thanks, this looks good :) Can you just roll back the diff in `package-lock.json` (and ideally use npm 8.x)?

mono424:
> Thanks, this looks good :) Can you just roll back the diff in `package-lock.json` (and ideally use npm 8.x)?

Done, also updated npm locally to 8.3 :)

joschahenningsen:
Great! Thanks a lot :fire: 

How about adding a "change lecture-hall" bulk action next? :)

joschahenningsen (approved):

PR #195, opened 2021-12-21
fix activity overview statistics

joschahenningsen:
There are a few problems with the stat calculation system:

- When there are weeks with e.g. VoD views but no live views, painting them both in one graph breaks it
- stats are not reliably sorted 
- some weeks appear duplicated. 

Worst case, it looks like this:
image

This PR makes sure, weeks with gaps between stats are filled with 0 stats, additionally,the MAX/SUM behaviour is fixed (currently the VoD view count incorrectly uses MAX instead of SUM)

Result: 
image


alexanderstephan (commented):
Nice fix! :+1:  Just some small things that I noticed.

Wouldn't there be a gap in years with 53 weeks right now? Maybe we could calculate the weeks similar to this: https://www.socketloop.com/tutorials/golang-calculate-how-many-weeks-left-to-go-in-a-given-year). It doesn't look too bad - but then again, setting it to 52 should work for now.

Also, I noticed a strange gap at the end of live statistics locally:
image
I poked around a little bit, but couldn't find the issue. Probably a problem on my end, since my statistics are completely out of sync, but I just wanted to have it mentioned here.

joschahenningsen:
> Also, I noticed a strange gap at the end of live statistics locally

That's probably because your db dump doesn't contain live views from later than week 48.

https://github.com/jinzhu/now looks fun! Also it's written by the same dude who made gorm :tada:

joschahenningsen:
I also added test automation to gh actions here

alexanderstephan (approved):
Looks great! :rocket: Also really nice addition of tests including the GitHub action.

If we should have some more tests in the future, we could consider adding this to `.pre-commit-config.yml`, but right now it doesn't really make sense to me.

PR #196, opened 2021-12-29
Set fixed go hook revision

alexanderstephan:
Fixes a warning for `pre-commit` regarding use of a moving branch. The author of the go hook scripts, see https://github.com/Bahjat/pre-commit-golang, set the revision to master in the template, but setting to a fixed revision should be more stable and remove the associated warning.


joschahenningsen (approved):
:+1:

PR #198, opened 2021-12-31
closable chat and error on disconnect

### LINKED ISSUE #174: Show "Connection to chat lost" indicator
When the WebSocket connection dies for some reason there should be a small indicator like `connection to chat lost, reconnecting...`

### LINKED ISSUE #175: Allow hiding chat
As the chat tends to be a little off-topic from time to time, students who want to focus should be able to hide it.

MatthiasReumann:
- Closes Issues #174 and #175 
- With the tests I made the chat window didn't show on mobile/smaller width devices, fixed that
- Chat State (Open/Close) stored in LocalStorage
- UI improvements


joschahenningsen:
This looks great! 
can the send button be next to the chat input when not on mobile? 
image

web/ts/watch.ts
```diff
@@ -26,6 +26,10 @@ function startWebsocket() {
     if (cf !== null && cf != undefined) {
         (document.getElementById("chatForm") as HTMLFormElement).addEventListener("submit", (e) => submitChat(e));
     }
+    ws.onopen = function (e) {
+        hideDisconnectMsg();
+    };
```
joschahenningsen: I'd initialize this hidden so the red message doesn't flicker up every time a page is opened.

web/ts/watch.ts
```diff
@@ -129,4 +138,12 @@ function submitChat(e: Event) {
     return false; //prevent form submission
 }
 
+function showDisconnectMsg() {
+    document.getElementById("disconnectMsg").style.display = "block";
+}
+
+function hideDisconnectMsg() {
+    document.getElementById("disconnectMsg").style.display = "none";
+}
```
joschahenningsen: Nit: You can add and remove the `hidden` & `block` classes here

web/template/partial/stream/chat.gohtml
```diff
@@ -1,14 +1,36 @@
 {{define "chat"}}
     {{- /*gotype: TUM-Live/web.WatchPageData*/ -}}
     {{$stream := .IndexData.TUMLiveContext.Stream}}
-    <div id="chatWrapper"
-         class="hidden md:flex flex-col w-full md:h-full border-t-2 mt-2 md:mt-0 md:border-t-0 md:dark:border-l dark:border-secondary dark:bg-secondary-lighter bg-gray-100 md:w-2/6 lg:w-1/6 text-1 pt-3">
-        <div id="chatBox" class="overflow-scroll overflow-x-hidden overflow-y-scroll" style="flex: 2;">
+    <script src="/static/assets/ts-dist/chat.js?v={{.IndexData.VersionTag}}"></script>
+    <div x-data="{show: false}" x-init="show = initChat();" id="chatWrapper"
```
joschahenningsen: This should help with the flickering when loading the website (alpine hides the chat until it's done initializing the page: https://alpinejs.dev/directives/cloak)

```suggestion
    <div x-cloak x-data="{show: false}" x-init="show = initChat();" id="chatWrapper"
```

joschahenningsen (changes requested):
Overall I'm super happy with this, great job :)

Just one more thing I noticed: when loading the page, the last message is not visible because it's hidden behind the input box. This probably happens because the input becomes visible (through alpine) after the chat was scrolled down by js. Maybe the scrolling should be taken care of in alpine? (I'm not seeing this issue every time)

Could you also resolve the conflicts with `dev` (see #217)

web/template/partial/stream/chat.gohtml
MatthiasReumann: Thanks! Works like a charm.

joschahenningsen (approved):
This looks super nice :)

PR #213, opened 2022-01-04
Fix warnings and deprecations after updating tailwindcss

joschahenningsen:


joschahenningsen:
the remaining failing static check will be fixed in #215

web/assets/css/main.css
```diff
@@ -224,7 +224,6 @@ input[type='checkbox']:checked + label {
 }
 
 .dark .fc-scroller {
-    @apply scrollbarThin;
```
alexanderstephan, line 227: I just wondered why @apply scrollbarThin is a circular dependency. There seem to be some false positives in this regard. Is this correct a warning or also a false positive?

Maybe https://github.com/tailwindlabs/tailwindcss/pull/6588 would change something about this warning (not sure)

alexanderstephan (commented):
Awesome! Just one minor note I where I was a bit confused. Maybe I am missing something regarding the dependency.

joschahenningsen:
Yeah, seems to be a false positive to me as well.
But removing the apply directive here doesn't make the calendar look bad, so I'd just circle back to this with the next tw release.

PR #215, opened 2022-01-04
fix grpc deprecations

joschahenningsen:
in response to:

```
Error: api_grpc/worker.go:416:79: grpc.WithInsecure is deprecated: use insecure.NewCredentials() instead. Will be supported throughout 1.x.  (SA1019)
Error: api_grpc/worker.go:463:78: grpc.WithInsecure is deprecated: use insecure.NewCredentials() instead. Will be supported throughout 1.x.  (SA1019)
```


joschahenningsen:
Other failing static checks/lint issues will be fixed in #213

alexanderstephan (approved):
Looks good to me! :+1:  The API change does not seem critical, but the earlier we migrate the better!

PR #217, opened 2022-01-07
Improve timer for waiting rooms

alexanderstephan:
This should improve the way the timer is handled and displayed.

I moved the formatting and waiting room specific implementation in the gohtml template.
This also enables a timer implementation that is a bit cleaner.
This way we can also set custom text like `Starting Soon` depending on the timer, without having to wrap or modify the timer. I did not know until recently that this was possible with alpine.js.

Besides, I added a little icon to the link, so it is a bit more visible.

How it looks:
image


joschahenningsen (approved):
Code seems much cleaner this way, thanks! :)

PR #229, opened 2022-01-11
Delete LectureHalls

### LINKED ISSUE #223: Delete Lecture halls

mono424:
I added the global "Delete"-Function, to send delete requests to the backend. If not wanted, this could be changed to a post request with a different URL for example "lectureHalls/:id/delete" but I liked delete request more :)

### DIFF
--- a/dao/lecture_halls.go
+++ b/dao/lecture_halls.go
@@ -46,7 +46,7 @@ func GetLectureHallByID(id uint) (model.LectureHall, error) {
 }
 
 func DeleteLectureHall(id uint) error {
-	return DB.Delete(model.LectureHall{}, "id = ?", id).Error
+	return DB.Delete(&model.LectureHall{}, id).Error
 }
 
 func SaveLectureHallFullAssoc(lectureHall model.LectureHall) {
--- a/dao/lecture_halls.go
+++ b/dao/lecture_halls.go
@@ -45,6 +45,10 @@ func GetLectureHallByID(id uint) (model.LectureHall, error) {
 	return lectureHall, err
 }
 
+func DeleteLectureHall(id uint) error {
+	return DB.Delete(model.LectureHall{}, "id = ?", id).Error
+}
+
 func SaveLectureHallFullAssoc(lectureHall model.LectureHall) {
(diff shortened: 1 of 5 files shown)


web/template/admin/admin_tabs/lecture_halls.gohtml
```diff
@@ -21,7 +21,12 @@
             pwrCtrlIp: '{{$lectureHall.PwrCtrlIp}}',
             id: '{{$lectureHall.ID}}',}"
              class="w-full border-2 dark:border-secondary border-gray-200 rounded mb-6">
-            <div class="dark:bg-secondary bg-gray-100 font-thin text-xl text-3 p-2 px-6">{{$lectureHall.Name}}</div>
+            <div class="dark:bg-secondary bg-gray-100 font-thin text-xl text-3 p-2 px-6">
+                {{$lectureHall.Name}}
+                <i class="fas fa-trash hover:text-1 px-2"
+                   title="Delete Lecture"
```
joschahenningsen:
```suggestion
                   title="Delete Lecture Hall"
```

dao/lecture_halls.go
```diff
@@ -45,6 +45,10 @@ func GetLectureHallByID(id uint) (model.LectureHall, error) {
 	return lectureHall, err
 }
 
+func DeleteLectureHall(id uint) error {
+	return DB.Delete(&model.LectureHall{}, id).Error
```
joschahenningsen: What happens with streams that are scheduled for the lecture hall we're about to delete? IMO, streams.lecture_hall_id should be set to `null` here. e.g. using DB.Raw. Also, camera presets for the lecture hall could be deleted

joschahenningsen (changes requested):
Looks good @mono424  :) 
Just a quick thought.

dao/lecture_halls.go
mono424: true! thanks for the catch! 👍

mono424: @joschahenningsen I think this should be fine now, used `DB.Exec` instead of `DB.Raw` :)

joschahenningsen (approved):
Looks good, thanks! :)

PR #230, opened 2022-01-11
Update course page

MatthiasReumann:
- Hidable Planned Lectures 
- Sortable VoDs
- Next Lecture Field (Notification Effect when Coming Up or Live)
- If Lecture is not coming up or live then the next lecture field date will be 'Today <time>', 'Tomorrow <time>', or '01.01.2022 <time>
- A bit of a redesign
- Changed the header and semester selection color to white in light-mode

### DIFF
--- a/model/stream.go
+++ b/model/stream.go
@@ -73,6 +73,11 @@ func (s Stream) IsStartingInMoreThanOneDay() bool {
 	return s.Start.After(time.Now().Add(48 * time.Hour))
 }
 
+// IsPlanned returns whether the stream is planned or not
+func (s Stream) IsPlanned() bool {
+	return !s.Recording && !s.LiveNow && !s.IsPast() && !s.IsComingUp()
+}
+
 type silence struct {
 	Start uint `json:"start"`
 	End   uint `json:"end"`
--- a/model/stream.go
+++ b/model/stream.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"encoding/json"
+	"fmt"
 	"github.com/microcosm-cc/bluemonday"
 	"github.com/russross/blackfriday/v2"
 	"gorm.io/gorm"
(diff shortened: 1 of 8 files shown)


model/stream.go
```diff
@@ -108,3 +114,17 @@ func (s Stream) FriendlyDate() string {
...
+	starty, startm, startd := s.Start.Date()
+
+	if nowy == starty && nowm == startm && nowd == startd {
+		return "Today,"
+	} else if tmrwy == starty && tmrwm == startm && tmrwd == startd {
+		return "Tomorrow,"
+	}
+
+	return fmt.Sprintf("%v %02d.", s.Start.Month(), s.Start.Day())
+}
```
joschahenningsen: Using jinzhu/now could improve readability: 
```suggestion
func (s Stream) FriendlyNextDate() string {
	if now.With(s.Start).EndOfDay() == now.EndOfDay() {
		return "Today,"
	}
	if now.With(s.Start).EndOfDay() == now.With(time.Now().Add(time.Hour*24)).EndOfDay() {
		return "Tomorrow,"
	}
	return s.Start.Format("Mon, January 02.")
}
```
Also, this adds the day of the week to the date.

web/template/planned_course_list.gohtml
```diff
@@ -0,0 +1,57 @@
+{{define "planned_course_list"}}
```
joschahenningsen, line 1: A placeholder for when there are no upcoming streams would be cool

web/template/vod_course_list.gohtml
```diff
@@ -0,0 +1,82 @@
+{{define "vod_course_list"}}
```
joschahenningsen, line 1: And a placeholder for 'no vods yet' as well

joschahenningsen (changes requested):
Love it! just a nit or two :)

web/template/header.gohtml
```diff
@@ -1,5 +1,5 @@
 {{define "header"}}
-    <nav class="flex z-40 lg:z-50 sticky top-0 w-full items-center justify-between dark:bg-secondary bg-secondary-light p-3 h-20 z-40">
+    <nav class="flex z-40 lg:z-50 sticky top-0 w-full items-center justify-between dark:bg-secondary bg-white shadow p-3 h-20 z-40">
```
joschahenningsen, line 2: IMO this is _very_ bright, but I don't really mind either

model/stream.go
MatthiasReumann: That's great. Thanks!

joschahenningsen: Oh and this function is exported, it should have some documentation (shame on me for not always doing this myself :D)

web/template/header.gohtml
MatthiasReumann: I kinda like it that way. (However, I use light mode everywhere :D)

joschahenningsen: You're the pro then 😀

joschahenningsen (approved):
:beers:

PR #233, opened 2022-01-17
Dynamic Websocket URL

joschahenningsen:
This allows the websocket to connect to whatever server is currently used.
if the request URL would be `http://localhost:8081/...` the websocket would connect to `ws://localhost:8081...`
on `https://live.rbg.tum.de/...` it would be `wss://live.rbg.tum.de/...` and on 
`https://test.live.mm.rbg.tum.de/...` it would be `wss://test.live.mm.rbg.tum.de/...`

### DIFF
--- a/web/ts/watch.ts
+++ b/web/ts/watch.ts
@@ -19,7 +19,7 @@ let retryInt = 5000; //retry connecting to websocket after this timeout
 const pageloaded = new Date();
 
 function startWebsocket() {
-    const wsProto = window.location.protocol.indexOf("https") !== -1 ? `wss://` : `ws://`;
+    const wsProto = window.location.protocol === "https:" ? `wss://` : `ws://`;
     const streamid = (document.getElementById("streamID") as HTMLInputElement).value;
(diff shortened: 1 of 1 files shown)


web/ts/watch.ts
```diff
@@ -19,8 +19,9 @@ let retryInt = 5000; //retry connecting to websocket after this timeout
 const pageloaded = new Date();
 
 function startWebsocket() {
+    const wsProto = window.location.protocol.indexOf("https") !== -1 ? `wss://` : `ws://`;
```
mono424: Just an idea but you could make it more simple: `window.location.protocol == "https:"`.
https://developer.mozilla.org/en-US/docs/Web/API/Location/protocol

joschahenningsen: You're right, that makes much more sense, I changed it :)

alexanderstephan (approved):
Looks good to me! Really nifty. :+1: Maybe we can reuse something like this in the future.

PR #238, opened 2022-01-19
Refactor index.go

### LINKED ISSUE #191: Refactor index.go data operations into dao
Loading the courses in [index.go](https://github.com/joschahenningsen/TUM-Live/blob/dev/web/index.go) could be faster.

Prepare the data in the dao accordingly to avoid sorting and filtering in `index.go`.

MatthiasReumann:
- Split up function into smaller ones 
- Added function to DAO for public and loggedIn courses with caching
- For 'LoadLivestreams' I added the function 'GetCurrentLiveNonHidden' which checks if the course associated with the stream isn't hidden. (Joins in DAO functions?)


model/course.go
```diff
@@ -158,4 +158,9 @@ func (c Course) GetRecordings() []Stream {
...
 	}
 	return recordings
-}
\ No newline at end of file
+}
+
+// IsHidden returns true if visibility is set to 'hidden' and false if not
+func (c Course) IsHidden() bool {
+	return c.Visibility == "hidden"
+}
```
joschahenningsen, line 176: So much better than checking this in the .gohtml :100:

web/index.go
```diff
@@ -29,62 +100,51 @@ func MainPage(c *gin.Context) {
...
-	var tumLiveContext tools.TUMLiveContext
-	tumLiveContextQueried, found := c.Get("TUMLiveContext")
-	if found {
-		tumLiveContext = tumLiveContextQueried.(tools.TUMLiveContext)
-		indexData.TUMLiveContext = tumLiveContext
-	} else { //TODO log
-		c.AbortWithStatus(http.StatusInternalServerError)
-		return
-	}
+func (d *IndexData) SetYearAndTerm(c *gin.Context) {
```
joschahenningsen, line 117: Can you add some comments to the new exported functions?

dao/courses.go
```diff
@@ -137,6 +157,14 @@ func GetCourseById(ctx context.Context, id uint) (course model.Course, err error
 	return foundCourse, dbErr
 }
 
+func GetNonHiddenCourseById(ctx context.Context, id uint) (course model.Course, err error) {
+	var foundCourse model.Course
+	dbErr := DB.Preload("Streams.Stats").Preload("Streams.Files").Preload("Streams", func(db *gorm.DB) *gorm.DB {
```
joschahenningsen: I think this doesn't need to preload the stats, as the function is not used in the admin tab.

joschahenningsen: BTW, where is it used?

dao/courses.go
```diff
@@ -96,15 +96,35 @@ func GetPublicCourses(year int, term string) (courses []model.Course, err error)
...
+		return cachedCourses.([]model.Course), err
+	}
+	var publicCourses []model.Course
+
+	err = DB.Preload("Streams", func(db *gorm.DB) *gorm.DB {
+		return db.Order("start asc")
+	}).Find(&publicCourses,
+		"(visibility = 'public' OR visibility = 'loggedin') AND teaching_term = ? AND year = ?", term, year).Error
+	if err == nil {
+		Cache.SetWithTTL(fmt.Sprintf("publicAndLoggedInCourses%v%v", year, term), publicCourses, 1, time.Minute)
```
joschahenningsen: nit: I did this wrong as well and this works just fine, but the right format string would be `publicAndLoggedInCourses%d%s`

joschahenningsen (changes requested):
This is much better!

dao/courses.go
MatthiasReumann: Nowhere actually.😅 Interesting that the test didn't fail for the unused function.

joschahenningsen: It's exported (capitalized), so the linter assumes we might want other projects to import it :smile:

joschahenningsen (approved):
Thanks :)

PR #247, opened 2022-01-21
fix panic on admin page when not logged in

joschahenningsen:
Currently, when not logged in, access to pages that are only accessible by admins of courses (e.g. `http://localhost:8081/admin/course/53`) cause nil value dereferences on `User` which causes a panic. This fixes the faulty middleware.

### DIFF
--- a/tools/middlewares.go
+++ b/tools/middlewares.go
@@ -137,7 +137,7 @@ func AdminOfCourse(c *gin.Context) {
 		return
 	}
 	tumLiveContext := foundContext.(TUMLiveContext)
-	if tumLiveContext.User.Role != model.AdminType && tumLiveContext.User.Model.ID != tumLiveContext.Course.UserID {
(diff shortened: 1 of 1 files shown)


tools/middlewares.go
```diff
@@ -168,16 +168,6 @@ func Admin(c *gin.Context) {
...
-func Worker(c *gin.Context) {
-	if wid := c.Param("workerID"); wid != "" {
-		_, err := dao.GetWorkerByID(c, wid)
-		if err == nil {
-			return
-		}
-	}
-	c.AbortWithStatus(http.StatusForbidden)
-}
-
```
joschahenningsen, line 180: This has been unused for a while

alexanderstephan (approved):
Yikes, just tried the bug on the testserver, did not look good in the logs :sweat: . Fix looks good, also nice to get rid of the dead code! :+1:

PR #248, opened 2022-01-21
don't  scroll down chat when scrolled up

### LINKED ISSUE #241: Chat scrolls automatically when in the not at the bottom
**Describe the bug**
When the chat isn't scrolled to the most bottom message and a new message appears, the chat jumps to the bottom.

**To Reproduce**
View any live chat.
Scroll to the middle.
Write a chat message with another window.
Result: The chat scrolls to the bottom to the new message

**Expected behavior**
When scrolled to a message, the chat should stay at that message and should not scroll down when a new one appears.
That makes it really hard to read past chat messages, because one always has to scroll up again.

**Desktop:**
 - OS: MacOS
 - Browser: Firefox
 - Version: 96.0.2

joschahenningsen:
fixes #241 

Desired behaviour: 

On New message:
- If the chat has not been scrolled up (or only a few pixels), the chat scrolls down like before
- if the chat has been scrolled up, a blue indicator is shown and the chat does not scroll down. 

When the indicator is shown:
- clicking the indicator scrolls down the chat and the indicator hides
- scrolling down manually also hides the indicator

https://user-images.githubusercontent.com/44805696/150636383-6371f2c9-02fd-4f2a-b172-86fa0c8bd9c6.mp4

### DIFF
--- a/web/ts/watch.ts
+++ b/web/ts/watch.ts
@@ -18,6 +18,16 @@ let ws: WebSocket;
 let retryInt = 5000; //retry connecting to websocket after this timeout
 const pageloaded = new Date();
 
+function scrollChat() {
+    const c = document.getElementById("chatBox");
+    if (c.scrollHeight - c.scrollTop <= c.offsetHeight + 150) {
+        c.scrollTop = c.scrollHeight;
+    } else {
+        console.log("not at bottom, skipping scroll");
+        // todo: add some visual indicator that the chat is not at the bottom and there are new messages
+    }
+}
+
 function startWebsocket() {
     const wsProto = window.location.protocol === "https:" ? `wss://` : `ws://`;
     const streamid = (document.getElementById("streamID") as HTMLInputElement).value;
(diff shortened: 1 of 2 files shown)


web/ts/watch.ts
```diff
@@ -18,6 +18,34 @@ let ws: WebSocket;
 let retryInt = 5000; //retry connecting to websocket after this timeout
 const pageloaded = new Date();
 
+function initChatScrollListener() {
+    const chatBox = document.getElementById("chatBox") as HTMLDivElement;
+    if (chatBox === undefined || chatBox === null) {
```
alexanderstephan: I think this could be `if (!chatBox) {...}` for a bit more generic check, but this is a matter of taste

alexanderstephan (approved):
Works great! Tested on Firefox 96.02, Opera 82.0.4227.43-1 and Chromium 96.0.4664.110 on https://test.live.mm.rbg.tum.de/w/tests/199.
Just a minor note on the code.

alexanderstephan:
I think there is a `!` in the if statement missing :see_no_evil:

joschahenningsen:
image

probably shouldn't be too proud of this :D

PR #251, opened 2022-01-22
Prevent event-stacking of progressReports by using setTimeout + recursion

alexanderstephan:
Some browsers don't seem to respect `setIntervall` that is currently used. There was a fix added that should prevent progress reports from being sent too often. This actually did not work that well, so from my testings progress reports are sometimes not sent during watching the stream.

The issue was probably due to the event stacking property that `setIntervall` has (https://developer.mozilla.org/en-US/docs/Web/API/setInterval#ensure_that_execution_duration_is_shorter_than_interval_frequency).

If the browser has problems with sending the reports, the events can accumulate over time and then fire all at once, once resources are available again.

 Implementing it with `setTimeout` should be more robust and avoids the before mentioned problem.

I deployed this on the test server https://test.live.mm.rbg.tum.de/w/tests/199 for easy testing. To test this, just reload without stopping the video after like 30s e.g.

### DIFF
--- a/web/ts/TUMLiveVjs.ts
+++ b/web/ts/TUMLiveVjs.ts
@@ -147,7 +147,6 @@ const watchProgress = function (streamID: number, lastProgress: float64) {
         let timer;
         let iOSReady = false;
         let intervalMillis = 10000;
-        let lastReport = 0; // timestamp of last report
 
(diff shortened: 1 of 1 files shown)


joschahenningsen (approved):
Thanks a ton!

PR #258, opened 2022-01-23
Fix ts deployment

joschahenningsen:
My bad, I broke the workflow last second :D https://github.com/joschahenningsen/TUM-Live/actions/runs/1736543074


joschahenningsen:
Finally. @alexanderstephan
Now as soon as the action removes the label, the server is deployed.

alexanderstephan (approved):
Deployment and healthcheck seem to behave as expected. :rocket:  Also, clean solution for the hash. :+1:

PR #239, opened 2022-01-20
Hamburger Menu

MatthiasReumann:
Closes #237 (Add hamburger menu on mobile devices)

PR #259, opened 2022-01-24
Fix VoD Ordering

### LINKED ISSUE #240: Only months change order
If you change the order of the lectures, it only changes the order of the months but not of the lectures within.
No matter if the months are ascending or descending, the newest lecture of a month is always on top.

Expected behavior:
The lectures within a month should also change order.
Ascending Order
Descending Order

MatthiasReumann:
Closes #240


joschahenningsen (changes requested):
Looks good!
This seems to contain changes from #239 though, could you update your branch?

MatthiasReumann:
> Looks good!
> This seems to contain changes from #239 though, could you update your branch?

Ah, crap. Seems like I created the new branch on the hamburger one. Any idea on how to solve that? 😅

joschahenningsen:
I think `git merge dev` on your branch should do

joschahenningsen (approved):
Ahh already merged, thanks!

PR #263, opened 2022-01-25
make chat colorful

### LINKED ISSUE #256: Make chat names distinguishable
When there are many people in a chat room some people might have the same (first) name. This makes it hard to account a message to a person when there is name conflict.

For example like Twitch we could assign every name with a random color, may it be really subtle/pale. Maybe a small icon could also be an option.

We could leave it like this. It looks definitely cleaner without any colors, like it is now. Alternatively, one could append the first letter of the last name.

**Additional context**
How [Twitch](https://twitch.tv) solves this:
image

joschahenningsen:
Fixes #256 

This adds one of 8 colors to every username (unfortunately only for future messages but hey...)

IMO the colors look great in both, light and dark mode:

light | dark 
--- | ---
image | image

### DIFF
--- a/model/chat.go
+++ b/model/chat.go
@@ -4,9 +4,12 @@ import (
 	"database/sql"
 	"errors"
 	"gorm.io/gorm"
+	"strconv"
 	"strings"
 	"time"
 )
+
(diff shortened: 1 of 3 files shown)


mono424:
Colors are looking dope! 🔥

PR #266, opened 2022-01-26
Add last page redirect

### LINKED ISSUE #253: Go back to last visited page after logging in
When logging-in the user is always redirected to the start page again, it would be cool, if the previous page would load again after login.

Automatically redirect the user to the last previously visited page.

MatthiasReumann:
- Closes #253 
- Split up the login function

### DIFF
--- a/web/user.go
+++ b/web/user.go
@@ -15,24 +15,67 @@ import (
 func LoginHandler(c *gin.Context) {
 	username := c.Request.FormValue("username")
 	password := c.Request.FormValue("password")
-	if u, err := dao.GetUserByEmail(context.Background(), username); err == nil {
-		// user with this email found.
-		if match, err := u.ComparePasswordAndHash(password); err == nil && match {
-			s := sessions.Default(c)
-			s.Set("UserID", u.ID)
-			s.Set("Name", u.Name)
-			_ = s.Save()
-			if c.Request.FormValue("return") != "" {
-				red, err := url.QueryUnescape(c.Request.FormValue("return"))
-				if err == nil {
-					c.Redirect(http.StatusFound, red)
-					return
-				}
-			}
-			c.Redirect(http.StatusFound, "/")
+
+	var data *sessionData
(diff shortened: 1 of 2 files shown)


web/user.go
```diff
@@ -15,24 +15,67 @@ import (
...
+		startSession(c, data)
+		c.Redirect(http.StatusFound, getRedirectUrl(c))
+		return
+	}
+
+	if data = loginWithTumCredentials(username, password); data != nil {
+		startSession(c, data)
+		c.Redirect(http.StatusFound, getRedirectUrl(c))
+		return
+	}
```
joschahenningsen: `startSession` redirects, then you redirect here again.

web/user.go
```diff
@@ -15,24 +15,67 @@ import (
...
+			return &sessionData{u.ID, u.Name}
 		}
 	}
+
+	return nil
+}
+
+// loginWithTumCredentials Try to login with tum credentials
+// Returns pointer to sessionData if successful and nil if not
+func loginWithTumCredentials(username, password string) *sessionData {
```
joschahenningsen: This should return `(sessionData, error)` with error as an indicator for success/failure

joschahenningsen (changes requested):
Nice refactoring! Just a few notes that came into my mind

web/user.go
```diff
@@ -15,24 +15,67 @@ import (
...
+	s.Set("UserID", data.userid)
+	s.Set("Name", data.name)
+	_ = s.Save()
+	if c.Request.FormValue("return") != "" {
+		red, err := url.QueryUnescape(c.Request.FormValue("return"))
+		if err == nil {
+			c.Redirect(http.StatusFound, red)
+			return
+		}
+	}
```
joschahenningsen: This should probably be part of `getRedirectUrl`

joschahenningsen (approved):
Works as expected :ok_hand:

PR #268, opened 2022-01-26
buffer stream progresses

joschahenningsen:
This adds a buffer that saves all incoming stream progresses. These are then flushed to the database at most once every 5 seconds.

### DIFF
--- a/api/progress.go
+++ b/api/progress.go
@@ -15,35 +15,35 @@ var progressBuff *progressBuffer
 
 // progressWorker contains progresses to be written to the database
 type progressBuffer struct {
-	Lock       sync.Mutex
-	Progresses []model.StreamProgress
+	lock       sync.Mutex
+	progresses []model.StreamProgress
 	interval   time.Duration
 }
 
 func newProgressBuffer() *progressBuffer {
 	return &progressBuffer{
-		Lock:       sync.Mutex{},
-		Progresses: []model.StreamProgress{},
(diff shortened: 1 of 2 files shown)


api/progress.go
```diff
@@ -2,13 +2,65 @@ package api
...
 	"github.com/gin-gonic/gin"
 	log "github.com/sirupsen/logrus"
 	"net/http"
+	"sync"
+	"time"
 )
 
+var progressBuff *progressBuffer
+
+// progressWorker contains progresses to be written to the database
```
alexanderstephan:
```suggestion
// progressBuff contains progresses to be written to the database
```

api/progress.go
```diff
@@ -2,13 +2,65 @@ package api
...
+	defer b.lock.Unlock()
+	if len(b.progresses) == 0 {
+		return nil
+	}
+	err := dao.SaveProgresses(b.progresses)
+	b.progresses = []model.StreamProgress{}
+	return err
+}
+
+// Run flushes the progress buffer every interval
```
alexanderstephan:
```suggestion
// run flushes the progress buffer every interval
```

alexanderstephan (changes requested):
Code looks good to me, everything worked beautifully when testing this! Just some nits regarding the comments, then this is good to go, I think!

alexanderstephan (approved):
Looks good to me, sorry for the typo!

PR #271, opened 2022-01-27
Unify and align design and layout of common pages

### LINKED ISSUE #249: Uniform admin page layout
**Describe the bug**
When going through the different tabs headlines are and top design are not fixed and changing for each admin tab.

**To Reproduce**
1. Go to "Admin"
2. Browse through tabs

**Expected behavior**
Spacing and design is consistent throughout admin tabs.

**Screenshots**
This should probably look more like image below for example.
image
image

### LINKED ISSUE #252: Fix live statistics button alignment
**Describe the bug**
The statistics button is placed below the download button in the admin course overview.

**To Reproduce**
1. Go to https://test.live.mm.rbg.tum.de/admin/course/2
2. Expand the latest VoD.
3. Notice wrong alignment 

**Expected behavior**
The live statistics button should always be beneath the download button.

**Screenshots**
image

 - OS: Manjaro
 - Browser Firefox
 - Version 96.0.2

alexanderstephan:
This PR:
- Fixes a bug where the side bar highlights the Course Imports and Server Statistics tab, where only Server Statistics would be shown. Example for the bug:
image
The fix:
image
Also note that there is a redesigned download button.

- Fixes odd alignment in the watch page by adding consistent spacing, for example weird spacing between the viewer count and items, and addition of a missing seperator symbol. Example for the previously buggy alignment:
image
How it looks now:
image

- Redesigns **all** the admin tabs. It also unifies the layout by using a similar structure and design elements for each of the admin tabs.

Most important design changes (to keep it to a fair length, I omitted the direct comparisons, these can be seen in the test server:

image

image
I want to highlight that I moved all the buttons to the bottom, so it is more intuitive to use. Before it was distributed in the upper row, the very right corner and a small reload button under Presets.

image

**What is not done in this PR:**

- Mobile alignment. It is assumed that a normal desktop PC is used for the backend. Creating consistent layouts for mobile devices for every page is out of scope and not really a use case at the moment (e.g. lecture vertical scrollable lecture hall fields on smaller screens).
- Tailwind components: Ideally every common element should have its own tailwind component, also out of scope.
- Sorting the tailwind classes. This will be done in a follow up with a specific linter plugin.

While this is still far from perfect, I hope this makes UI/UX quite a bit better.


joschahenningsen (approved):
This looks much better than before (and like a ton of work!) Thanks :) 
I'll follow up on the token page :)

PR #275, opened 2022-01-29
allow links in messages

### LINKED ISSUE #273: Allow links in chat
links posted to the chat should be clickable HTML `a` tags

joschahenningsen:
This PR also introduces a way of manually migrating the database (in case it's needed in the future sometime)

Since allowing HTML in user-generated messages is a major security concern, I'd appreciate it if y'all could have a careful look at the HTML sanitation and maybe suggest a few more test cases :)

image

Fixes #273


MatthiasReumann:
Looks good. However, do we even want to allow "normal" (and especially anonymous) users to send links or should it be restricted to admins/lecturers?

joschahenningsen:
> Looks good. However, do we even want to allow "normal" (and especially anonymous) users to send links or should it be restricted to admins/lecturers?

Let's merge this as is and revisit this in case someone misbehaves or complains.

PR #277, opened 2022-01-31
like messages and order by likes

### LINKED ISSUE #276: Like messages
Messages should be likeable and sortable by likes

joschahenningsen:
https://user-images.githubusercontent.com/44805696/151780377-ec09ddd1-ded7-46df-9c83-93174ee2b86e.mp4


api/chat.go
```diff
@@ -45,61 +45,124 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
-			ReplyTo:  replyTo,
-		}
-		err := dao.AddMessage(&chatForDb)
+		var req wsReq
+		err := json.Unmarshal(msg, &req)
 		if err != nil {
-			if errors.Is(err, model.ErrCooledDown) {
-				sendServerMessage("You are sending messages too fast. Please wait a bit.", TypeServerErr, s)
-			}
+			log.Warn(err)
```
alexanderstephan: Maybe we should send a more verbose sentry exception here or even abort with an HTTP status code?

api/chat.go
```diff
@@ -45,61 +45,124 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+			handleLike(tumLiveContext, msg)
 		}
 	})
 }
 
+func handleLike(ctx tools.TUMLiveContext, msg []byte) {
+	var req likeReq
+	err := json.Unmarshal(msg, &req)
+	if err != nil {
+		return
```
alexanderstephan, line 70: Should be consistent with error handling of the other `json.Unmarshal`

api/chat.go
```diff
@@ -45,61 +45,124 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+		log.WithError(err).Error("error getting num of chat likes")
+		return
+	}
+	broadcast := gin.H{
+		"likes": req.Id,
+		"num":   numLikes,
+	}
+	broadcastBytes, err := json.Marshal(broadcast)
+	if err != nil {
+		log.WithError(err).Warn("Can't marshal like message")
```
alexanderstephan: Should also be consistent to Unmarshal operations

dao/chat.go
```diff
@@ -2,8 +2,48 @@ package dao
...
+	var numLikes int
+	err := DB.Raw("SELECT COUNT(*) FROM chat_user_likes WHERE chat_id = ?", chatID).Scan(&numLikes).Error
+	return numLikes, err
+}
+
+// GetChats returns all chats for the stream with the given ID. Number of likes are inserted and the user's like status is determined
+func GetChats(userID uint, streamID uint) ([]model.Chat, error) {
+	var chats []model.Chat
+	err := DB.Preload("Replies").Preload("UserLikes").Find(&chats, "stream_id = ?", streamID).Error
+	if err == nil {
```
alexanderstephan: Maybe we could unnest this by returning in case of `err != nil`.
```suggestion
	if err != nil {
	    return nil, err
	}
```

dao/chat.go
```diff
@@ -2,8 +2,48 @@ package dao
...
+		for i := range chats {
```
alexanderstephan:
```suggestion
		for _, chat := range chats {
```
Matter of taste, but I think it is safer to omit the index. Rest should be updated to the new variable.

api/chat.go
```diff
@@ -45,61 +45,124 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
 		if err != nil {
-			if errors.Is(err, model.ErrCooledDown) {
-				sendServerMessage("You are sending messages too fast. Please wait a bit.", TypeServerErr, s)
-			}
+			log.Warn(err)
 			return
 		}
-		if broadcast, err := json.Marshal(chatForDb); err == nil {
-			broadcastStream(tumLiveContext.Stream.ID, broadcast)
+		if req.Type == "message" {
```
alexanderstephan: Maybe we can use a `switch` here and handle an unknown message type in the `default` case?

api/chat.go
```diff
@@ -45,61 +45,124 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
 	Anonymous bool   `json:"anonymous"`
 	ReplyTo   int64  `json:"replyTo"`
 }
+
+type likeReq struct {
+	wsReq
+	Id uint `json:"id"`
+}
+
 type ChatRep struct {
```
alexanderstephan: Not related to this PR, but it would be good, if `ChatRep` had a comment and or a more verbose name.

dao/chat.go
```diff
@@ -2,8 +2,48 @@ package dao
...
+	if errors.As(err, &mysqlErr) && mysqlErr.Number == 1062 { // 1062: duplicate entry -> message already liked -> remove
+		return DB.Exec("DELETE FROM chat_user_likes WHERE user_id = ? AND chat_id = ?", userID, chatID).Error
+	}
+	return err // some other error
+}
+
+// GetNumLikes returns the number of likes for a message
+func GetNumLikes(chatID uint) (int, error) {
+	var numLikes int
+	err := DB.Raw("SELECT COUNT(*) FROM chat_user_likes WHERE chat_id = ?", chatID).Scan(&numLikes).Error
```
alexanderstephan: Maybe we can use https://gorm.io/docs/advanced_query.html#Count here?

web/ts/watch.ts
```diff
@@ -16,8 +16,36 @@ export class Watch {
...
+export function likeMessage(id: number) {
+    ws.send(
+        JSON.stringify({
+            type: "like",
+            id: id,
+        }),
+    );
+}
+
+export function sortMessages(messages) {
```
alexanderstephan: This function should have a return type, I think

web/ts/watch.ts
```diff
@@ -69,9 +109,15 @@ export function startWebsocket() {
                 window.dispatchEvent(new CustomEvent("pauseend"));
             }
         } else if ("server" in data) {
+            const scroll = shouldScroll();
+            console.log("scroll: ", scroll);
             const serverElem = createServerMessage(data);
             document.getElementById("chatBox").appendChild(serverElem);
-            scrollChatIfNeeded();
+            if (scroll) {
+                setTimeout(scrollChat, 100); // wait a bit to make sure the message is added before scrolling
```
alexanderstephan: Maybe we could extract `100` into a constant and call it `scrollDelay`.

web/ts/watch.ts
```diff
@@ -56,6 +95,7 @@ export function startWebsocket() {
     };
 
     ws.onmessage = function (m) {
+        console.log("got message");
```
alexanderstephan: Do we need this here?

web/ts/watch.ts
```diff
@@ -30,14 +58,25 @@ export function initChatScrollListener() {
...
-        window.dispatchEvent(new CustomEvent("messageindicator", { detail: { show: true } }));
+    return c.scrollHeight - c.scrollTop <= c.offsetHeight;
+}
+
+function showNewMessageIndicator() {
+    window.dispatchEvent(new CustomEvent("messageindicator", { detail: { show: true } }));
+}
+
+export function scrollChat() {
+    console.log("scrollChatIfNeeded");
```
alexanderstephan: Just marking this in case it was forgotten to remove it. Otherwise it's fine.

web/ts/watch.ts
```diff
@@ -126,6 +180,7 @@ export function createServerMessage(msg) {
 export function sendMessage(message: string, anonymous: boolean, replyTo: number) {
     ws.send(
         JSON.stringify({
+            type: "message",
```
alexanderstephan: Maybe we could create an `enum` of message types that can be used across this file.

alexanderstephan (changes requested):
I love this, didn't expect to see this feature so quickly! Maybe we can test this on saturday for our GBS course review!

One problem I see here is that the counter is the button at the same time. We could maybe separate the like from the counter itself or make it more obvious that this is also the like button.

Also, Reply displays a hint when hovering, Like does not at the moment, this should also be added for consistency. E.g. "Click here to like this message".

Otherwise, everything worked as expected. :+1:
Code looks really nice overall, just a few nits.

api/chat.go
joschahenningsen: Warn should be a good error level here, as users can potentially send whatever invalid data they want. I made it more verbose though :+1:

api/chat.go
joschahenningsen: :+1: I made this a switch statement and log the invalid request type as `default`

api/chat.go
joschahenningsen: :+1:

api/chat.go
joschahenningsen: Yup, this should be `Error` :)

api/chat.go
joschahenningsen: It's not used anymore, I removed it.

dao/chat.go
joschahenningsen: Definitely :+1:

dao/chat.go
joschahenningsen: No idea why I did this this way :D

dao/chat.go
joschahenningsen: Since I'm assigning here and the chat is not a pointer, I'd like to keep the index here.

web/ts/watch.ts
joschahenningsen: `Array.prototype.sort()` sorts in place. Should we add `void` to such functions?

web/ts/watch.ts
joschahenningsen: Yup, thanks :)

web/ts/watch.ts
joschahenningsen: :+1:

web/ts/watch.ts
joschahenningsen: :+1:

web/ts/watch.ts
joschahenningsen: :+1:

api/chat.go
alexanderstephan: Okay, perfect :+1:

web/ts/watch.ts
alexanderstephan: From what I've seen for example at matrix `void` is usually added here to be more explicit and safe, but it is not really needed.

joschahenningsen:
Would you prefer a `thumb-up`?

image

web/ts/watch.ts
joschahenningsen: I added `void`, since we have a bunch of such methods, do you think it would make sense to apply this everywhere or enforce it with eslint? @alexanderstephan

alexanderstephan: Yes, that would be nice, see https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-function-return-type.md e,g.
But we could add that as well in a follow up.

alexanderstephan (approved):
I think this looks great, I especially really like that there is no number when there is no like.

I think this is good to go, but one last thing that could be considered is to remove the icon and just call it "Like" so it's consistent with "Reply". I am fine with either, though. YouTube does it how it is right now, so it's definitely good already!

Example:
image

joschahenningsen:
> I think this looks great, I especially really like that there is no number when there is no like.
> 
> I think this is good to go, but one last thing that could be considered is to remove the icon and just call it "Like" so it's consistent with "Reply". I am fine with either, though. YouTube does it how it is right now, so it's definitely good already!
> 
> Example:
> image
> 

Yeah that looks better, I was too tired this morning so I overread your suggestion, gonna apply it though 👍

PR #281, opened 2022-02-01
Mobile Semester Selection

### LINKED ISSUE #267: Mobile Semester Selection
The semester selection on mobile devices is not optimal.

image

Move this area to the side navigation.

Own button on the bottom right of the mobile page which opens a dialog.

MatthiasReumann:
- Closes #267 
- Removed chat typescript code and replaced it with Alpine persist
- Something bugged me about the course page in dark mode and I finally figured it out and fixed it. The colours were inverted! Now the background is darker than the foreground, as it is in light mode.
localhost_8081_ (2)
localhost_8081_ (3)
localhost_8081_ (1)
ddd


joschahenningsen:
This looks amazing! Could we use something smaller for Desktop as well? The semester list will eventually outgrow all screen sizes.

Alpine has a `@click.outside` event listener: https://alpinejs.dev/start-here#listening-for-a-click-outside maybe this can be used to dismiss the selector when tapping somewhere outside the menu?

joschahenningsen:
The semester selection looks amazing! I don't see the blur effect from your screenshots, but that's fine by me. 

> Something bugged me about the course page in dark mode and I finally figured it out and fixed it. The colors were inverted! Now the background is darker than the foreground, as it is in light mode.

The dark mode version of the course page has another background color than all other pages now, I thought the version before was nice. If this should be changed it should be consistent across all pages. But again, I think the current background is good :)

MatthiasReumann:
> The semester selection looks amazing! I don't see the blur effect from your screenshots, but that's fine by me.
> 
> > Something bugged me about the course page in dark mode and I finally figured it out and fixed it. The colors were inverted! Now the background is darker than the foreground, as it is in light mode.
> 
> The dark mode version of the course page has another background color than all other pages now, I thought the version before was nice. If this should be changed it should be consistent across all pages. But again, I think the current background is good :)

Could be that the css property is not supported by all browsers (I used Tailwinds backdrop-blur class)

I've also updated the dark mode to something similar to before again (Background is back to normal!). In a side by side comparison with the current live version I think the updated one is much better.

joschahenningsen (approved):
> Could be that the css property is not supported by all browsers (I used Tailwinds backdrop-blur class)

The backdrop is gonna be a nice addition for everyone not on firefox then :D 

This looks great, thanks!

PR #282, opened 2022-02-01
Delete Course Button

### LINKED ISSUE #228: Delete courses
(and their lectures)

MatthiasReumann:
- Closes #228 

I've noticed that after a course is deleted it is still shown on the left menu (It is however deleted). Any idea why this could be the case? (Caching?)


joschahenningsen:
Could be caching, yeah. I usually call `Cache.Clear()` after significant changes like this one. Does that help?

MatthiasReumann:
> Could be caching, yeah. I usually call `Cache.Clear()` after significant changes like this one. Does that help?

Yup. Solved it, thanks!

api/courses.go
```diff
@@ -576,6 +577,21 @@ func createCourse(c *gin.Context) {
...
+func deleteCourse(c *gin.Context) {
+	foundContext, exists := c.Get("TUMLiveContext")
+	if !exists {
+		sentry.CaptureException(errors.New("context should exist but doesn't"))
+		c.AbortWithStatus(http.StatusInternalServerError)
+		return
+	}
+
+	tumLiveContext := foundContext.(tools.TUMLiveContext)
+	fmt.Println(*tumLiveContext.Course)
```
joschahenningsen:
```suggestion
```

joschahenningsen: This print statement doesn't seem to add much value. But a little
```go
log.WithFields(log.Fields{
	"user": tumLiveContext.User.ID,
	"course": tumLiveContext.Course.ID,
}).Info("Delete Course Called")
```
Would be nice

joschahenningsen (changes requested):
Looks good, just a thought

joschahenningsen (approved):
Sorry for the late review, looks good, thanks!

PR #298, opened 2022-02-07
Recurring stream option on create stream

mono424:


api/courses.go
```diff
@@ -465,37 +463,46 @@ func createLecture(c *gin.Context) {
...
-	// add file if premiere
-	if req.Premiere || req.Vodup {
-		lecture.Files = []model.File{{Path: fmt.Sprintf("%s/%s", premiereFolder, premiereFileName)}}
+	// Add start date as first event
+	req.DateSeries = append(req.DateSeries, req.Start)
+
+	for _, date := range req.DateSeries {
+		endTime := date.Add(time.Minute * time.Duration(req.Duration))
+
+		lecture := model.Stream{
```
joschahenningsen, line 473: In case there are multiple streams created at once, can you do this:

- add a new field to `model.Stream`, `SeriesIdentifier string`, with a default value of `null`
- generate a random string using `uuid.NewV4().String()`
- set this random string on all lectures in an event series

This way we'll be able to update all events that belong to a series in bulk.

joschahenningsen (changes requested):
This looks great so far, especially the UI is much more intuitive now! 
Just a thought on saving the events.

api/courses.go
mono424: Current Approach also adds a identifier to single entries. Would you prefer on single entries null as identifier?

mono424:
Made selection of the dates more visible ;)
Bildschirmfoto 2022-02-10 um 23 17 57

api/courses.go
joschahenningsen: Single entries should have a null identifier IMO

api/courses.go
```diff
@@ -437,20 +435,20 @@ func createLecture(c *gin.Context) {
...
-			return
-		}
-		reqFile, _ := req.File.Open()
-		_, err = io.Copy(file, reqFile)
-		if err != nil {
-			log.WithError(err).Error("Can't write file for premiere")
-			c.AbortWithStatus(http.StatusInternalServerError)
-		}
-		_ = file.Close()
+		//// Copy file to shared storage
```
MatthiasReumann, line 438: I would argue if the commented isn't needed right now it can be deleted or refactored into a separate function.

MatthiasReumann (commented):
Looks great :) There is just one tiny detail I can criticise.

api/courses.go
mono424: Totally agree! :)

mono424: @joschahenningsen do you need that comment to implement the new upload function? Otherwise I would remove it 👍

joschahenningsen (approved):
Looks Perfect :ok_hand:

api/courses.go
joschahenningsen: will follow up on this

PR #300, opened 2022-02-09
Add panic button

### LINKED ISSUE #250: Panic button on watch page
There is a large amount of streams on TUMlive. This makes it hard to track all the different streams.
We could also include a button on the viewer side that sends a similar message after e.g. 10% of the viewers complained about issues.

We could introduce an additional button on the watch page that triggers a message that would get send to a Matrix Channel stating the ID of a stream and includes a link, so issues could be checked out.

We could use https://matrix.org/docs/guides/usage-of-matrix-bot-sdk to send the message.

alexanderstephan:
This PR adds a panic button, that sends a message to a matrix room with useful information about the stream that has problems.

**Todos**:
- [x] Create config for matrix
- [x] Design GUI in watch page for sending the message
- [x] Write small matrix library for simple, unencrypted message exchange
- [x] Revert watch page
- [x] Use bluemonday for formatting
- ~Potentially write some small tests where possible~ Due to the current structure not really useful without providing secrets

**Design**:
image


tools/bot/bot.go
```diff
@@ -0,0 +1,65 @@
...
+func (c *BotInfo) BotUpdate(info BotInfo) error {
+	// Currently only supported for lecture hall streams
+	if info.LectureHall == "" {
+		return errors.New("sending bot messages is not supported for selfstreams")
+	}
+	err := c.BotMessaging.SendBotMessage(info)
+	if err != nil {
+		return err
+	}
+	return nil
```
Garfield96:
```suggestion
	return err
```

tools/bot/bot.go
```diff
@@ -0,0 +1,65 @@
...
+
+// getMessageText generates an unformatted issue text.
+func getMessageText(botInfo BotInfo) string {
+	return "🚨 Technical problem:" + "\n" +
+		"Course name: " + botInfo.CourseName + "\n" +
+		"Lecture hall: " + botInfo.LectureHall + "\n" +
+		"URL: " + botInfo.StreamUrl + "\n" +
+		"Lecture hall: " + botInfo.LectureHall + "\n" +
+		"CombinedIP: " + "http://" + botInfo.CombIP + "\n" +
+		"CameraIP: " + "http://" + botInfo.CameraIP + "\n"
```
Garfield96: Is it possible to use http**s** instead?

joschahenningsen: Unfortunately not :( This could be made configurable in the `LectureHall` at some point though

config.yaml
```diff
@@ -32,5 +32,12 @@ auths:
   smpPassword: "password"
   pwrCrtlAuth: "user:password"
   camAuth: "user:password"
+matrix:
+  username: "rbgbot"
+  password: "password"
+  homeserver: "matrix.org"
+  roomId: "roomid"
+
```
joschahenningsen: Just a thought on future extensibility

 
Something like this:
```suggestion
alerts:
  matrix:
   username: "rbgbot"
   password: "password"
   homeserver: "matrix.org"
   roomId: "roomid"
```

With a corresponding config struct like this one:
```go
Alerts struct {
		Matrix *struct {
			Username   string `yaml:"username"`
			Password   string `yaml:"password"`
			Homeserver string `yaml:"homeserver"`
			RoomID     string `yaml:"roomId"`
		} `yaml:"matrix"`
		Email *struct {
			Address string `yaml:"address"`
			Name  string `yaml:"name"`
		} `yaml:"email"`
	} `yaml:"alerts"`
```
would allow optional other strategies for future use. (Note that email is a pointer which defaults to nil if not configured)

joschahenningsen: Alerts could then be a pointer as well? Correct me if I'm wrong though :D

alexanderstephan: Nice idea, structuring the config like this is really a lot cleaner. :+1:
Yes, a pointer seems to work in this case.

api/stream.go
```diff
@@ -130,6 +133,47 @@ func pauseStream(c *gin.Context) {
...
+		sentry.CaptureException(errors.New("context should exist but doesn't"))
+		c.AbortWithStatus(http.StatusInternalServerError)
+		return
+	}
+	tumLiveContext := foundContext.(tools.TUMLiveContext)
+	stream := tumLiveContext.Stream
+
+	lectureHall, err := dao.GetLectureHallByID(stream.LectureHallID)
+	if err != nil {
+		sentry.CaptureException(err)
```
joschahenningsen, line 153: Logging an error here is sufficient, logrus will report it to sentry

api/stream.go
```diff
@@ -130,6 +133,47 @@ func pauseStream(c *gin.Context) {
...
+		c.AbortWithStatus(http.StatusInternalServerError)
+	}
+	course, err := dao.GetCourseById(c, stream.CourseID)
+	if err != nil {
+		log.WithError(err).Error("Can't send bot message")
+		sentry.CaptureException(err)
```
joschahenningsen, line 158: Here as well

api/stream.go
```diff
@@ -130,6 +133,47 @@ func pauseStream(c *gin.Context) {
...
+		CombIP:      lectureHall.CombIP,
+		CameraIP:    lectureHall.CameraIP,
+	}
+	// Set messaging strategy as specified in strategy pattern
+	botInfo.SetMessagingMethod(&bot.Matrix{})
+	err = botInfo.BotUpdate(botInfo)
+
+	if err != nil {
+		log.WithError(err).Error("Can't send bot message")
+		sentry.CaptureException(err)
```
joschahenningsen, line 174: And here

api/stream.go
```diff
@@ -130,6 +133,47 @@ func pauseStream(c *gin.Context) {
...
+		c.AbortWithStatus(http.StatusInternalServerError)
+	}
+	course, err := dao.GetCourseById(c, stream.CourseID)
+	if err != nil {
+		log.WithError(err).Error("Can't send bot message")
+		sentry.CaptureException(err)
+		c.AbortWithStatus(http.StatusInternalServerError)
+	}
+	streamUrl := tools.Cfg.WebUrl + "/w/" + course.Slug + "/" + strconv.Itoa(int(stream.ID))
```
joschahenningsen: fmt.Sprintf could be used to avoid string conversion here

tools/bot/matrix.go
```diff
@@ -0,0 +1,147 @@
...
+		return "", err
+	}
+	response, err := http.Post(loginUrl, contentType, bytes.NewBuffer(loginRequest))
+	if err != nil {
+		return "", err
+	}
+	if response.StatusCode != http.StatusOK {
+		return "", fmt.Errorf(fmt.Sprintf("received status code %d instead of %d.", response.StatusCode, http.StatusOK))
+	}
+	responseBody, err := io.ReadAll(response.Body)
```
joschahenningsen: You can use jsons Decoder to unmarshal the reader https://pkg.go.dev/encoding/json#NewDecoder

tools/bot/matrix.go
```diff
@@ -0,0 +1,147 @@
...
+	Password string `json:"password"`
+}
+
+// MessageResponse is the response for a message in our case
+type MessageResponse struct {
+	EventID string `json:"event_id"`
+}
+
+// LoginResponse is the response for a login request
+type LoginResponse struct {
```
joschahenningsen: Do we need to export all these types?

tools/bot/matrix.go
```diff
@@ -0,0 +1,147 @@
...
+	"strconv"
+)
+
+// Matrix strategy
+type Matrix struct {
+}
+
+// Message represents a Matrix message event that includes html formatting as specified
+// in https://spec.matrix.org/v1.2/client-server-api/#mroommessage-msgtypes
+type Message struct {
```
joschahenningsen: These strategy specific stuff should either be in a matrix package or be renamed: 

```suggestion
type MatrixMessage struct {
```

tools/bot/matrix.go
```diff
@@ -0,0 +1,147 @@
...
+	matrixMessageJSON, err := json.Marshal(matrixMessage)
+	if err != nil {
+		return err
+	}
+	err = sendMessageRequest(url, bytes.NewBuffer(matrixMessageJSON))
+	return err
+}
+
+// sendMessageRequest sends a PUT request to url with a given body
+func sendMessageRequest(url string, body io.Reader) error {
```
joschahenningsen: Can this be a method on the matrix type to abstract it out of the package root? 

```suggestion
func (m Matrix)sendMessageRequest(url string, body io.Reader) error {
```

tools/bot/matrix.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	response, err := client.Do(request)
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf(fmt.Sprintf("received status code %d instead of %d.", response.StatusCode, http.StatusOK))
+	}
+	return err
+}
+
+// getAuthToken retrieves a single use token for the next message sent to the server.
+func getAuthToken() (string, error) {
```
joschahenningsen: Might also be something like

```suggestion
func (m Matrix)getAuthToken() (string, error) {
```

joschahenningsen (changes requested):
Looks good already, a few comments though

tools/bot/matrix.go
alexanderstephan: I renamed it, since the name of a struct should not contain the name of the file, but yeah, this might cause conflicts at some point. Calling it `MatrixMessage` should be fine for now, I guess. :+1:  Also, the other `struct` is still called `Matrix`, soo... :sweat_smile:

api/stream.go
alexanderstephan: `fmt.Sprintf` is pretty bloated due to reflections and therefore a lot slower, but since it only used once here, it should be good  :+1:

joschahenningsen: Didn't do any benchmarks on this yet, nice to know, thanks! This is not a hot path though, so imo optimising for readability is fine

api/stream.go
```diff
@@ -130,6 +133,47 @@ func pauseStream(c *gin.Context) {
 	}
 }
 
+// reportStreamIssue sends a notification to a matrix room that can be used for debugging technical issues.
+func reportStreamIssue(c *gin.Context) {
```
joschahenningsen, line 136: We should also make sure the stream is actually live before sending a message

tools/bot/matrix.go
alexanderstephan: Not really anymore. I actually don't see a use case since refactoring everything matrix related into one file. This can be unexported now.

api/stream.go
alexanderstephan: There might be cases where a stream completely crashed or was quit by accident that might be dealt with in the future. But right now it isn't be possible yet, so also checking this for sanity is a good idea. :+1:

joschahenningsen: You're right, this is a legitimate reason. How about just checking that the stream date is today?

alexanderstephan: That sounds reasonable!

api/stream.go
alexanderstephan: Yes, readability is way more important here.

web/template/watch.gohtml
```diff
@@ -193,8 +193,13 @@
                             </div>
                         </section>
                     </span>
+                    <button @click="fetch(`/api/stream/${streamID}/issue`)"
+                            title="Issues with the stream? Send a report to the RGB."
+                            class="bg-amber-500 inline-block float-right text-center w-auto text-white font-bold hover:bg-amber-700 dark:hover:bg-amber-700 rounded cursor-pointer p-2">
+                    <i class="fas fa-exclamation-triangle mr-1"></i>Report Technical Problem
+                    </button>
```
joschahenningsen: Just one more thing I noticed; can we hide this when alerts are not configured?

alexanderstephan: Yes, that's a good idea!

joschahenningsen (approved):
Looks great, thanks!

PR #306, opened 2022-02-15
self announce worker at TUM-Live using token

joschahenningsen:
Related PR: https://github.com/joschahenningsen/TUM-Live-Worker-v2/pull/23 (needs merge first)

This PR allows workers to automatically join without manual actions from an admin. This is especially useful for throwing a bunch of worker containers into a cluster without manually managing them.

---

How it works:

- At startup, TUM-Live generates a token that can be used to register workers
- When a worker starts, it sends a message to TUM-Live, containing its hostname and the token
- If the token is correct, TUM-Live determines if the worker on that host already exists. 
  - If yes, TUM-Live responds with the workers secret key, which is used for all authentication on subsequent calls
  - If not, TUM-Live inserts the worker into the database and generates said secret. The response looks the same to the worker.


config.yaml
```diff
@@ -1,36 +1,37 @@
...
-  userDn: "uid=%s,ou=people,dc=example,dc=de"
+  basedn: dc=example,dc=de
+  password: secret
+  url: ldap://abc.de
+  user: user
+  userdn: uid=%s,ou=people,dc=example,dc=de
+lrz:
+  email: erika.mustermann@example.com
+  name: Erika Mustermann
+  phone: "0123456789"
```
alexanderstephan, line 26:
```suggestion
  phone: 0123456789
```

config.yaml
```diff
@@ -1,36 +1,37 @@
...
-  smpUser: "username"
-  smpPassword: "password"
-  pwrCrtlAuth: "user:password"
-  camAuth: "user:password"
-ingestBase: "rtmp://ingest.tum.live/"
-cookieStoreSecret: "put 40 random characters here"
\ No newline at end of file
+  mass: /srv/ceph/live
+  static: /var/www/static
+workerToken: 2657575f-40e4-4278-9ca1-f3a5ed667cff
```
alexanderstephan, line 37: Just to make sure, this an example?

web/template/admin/admin_tabs/workers.gohtml
```diff
@@ -34,22 +37,48 @@
...
-                    <i class="fas fa-plus dark:disabled:bg-gray-600 w-4 mr-2 transform hover:text-purple-500 hover:scale-110 text-gray-500 disabled:text-gray-300" :disabled="hostname==''"
-                            @click="secret===''?fetch('/api/worker', {method:'POST', body:JSON.stringify({'hostname': hostname})}).then(r=>r.json()).then(d=>secret=d.workerID):window.location.reload()"
-                    >
-                    </i>
-                </td>
-            </tr>
             </tbody>
         </table>
     </div>
+    <h3 class="text-3">Add a worker</h3>
```
alexanderstephan:
```suggestion
    <h3 class="text-3">How to add a worker</h3>
```
Maybe I am stupid, but I thought I could click something here. Maybe this avoids confusion?

alexanderstephan (changes requested):
Looks really good to me so far! I really like the added docs in the worker page. :rocket: Otherwise, the token management should get a consistent look at some point, but for now this is more than fine. :+1:

When checking this out locally, I ran into a issue where the dependencies had versions for go `1.16`, which made me unable to sync. Running `go mod tidy -compat=1.17` fixed this for me. I assume compatibility with `1.17` is sufficient?

Otherwise just some minor comments. :+1:

config.yaml
joschahenningsen: Yeah that's what my local instance generated, figured this would be fine to commit

config.yaml
joschahenningsen: This file was generated by viper (that's also why it's sorted now), the phone number is a string because it could contain something like "+49"

config.yaml
alexanderstephan: Ah, makes sense!

joschahenningsen:
> I assume compatibility with 1.17 is sufficient?

Yes, that sounds good, I'll look into it :)

config.yaml
alexanderstephan: Isn't everything a string here anyways? I checked the annotation, and it seemed to annotated as string anyways.  But not a big deal anyways.

joschahenningsen: Not entirely sure tbh, but since viper will change this again when the config is written I'd keep this as it

joschahenningsen:
I tidied the go module and also upgraded some dependencies while I was at it.

alexanderstephan (approved):
Perfect, thanks! :100:

PR #308, opened 2022-02-19
Fix download popup positioning

### LINKED ISSUE #307: 'Download Video' popup gets displayed at the right of the screen
**Describe the bug**
Download popup after clicking on the download icon gets displayed at the right hand side of the viewport, instead of below the icon / mouse like in a previous version.

**To Reproduce**
1. Go to a VOD
2. Click on the download icon
3. See error

**Screenshots**
Bildschirmfoto 2022-02-18 um 17 44 16

 - OS: macOS 12.1
 - Browser: Safari 15.2, Chrome 96.0.4664.110

MatthiasReumann:
- Closes #307 
- Added `dark:` classes for popup

Since I can't manage to checkout properly `web/template/partial/stream/download.gohtml` is the only file with changes on my side 😅


alexanderstephan (commented):
Thanks, looks and works good! I think I messed this up. :see_no_evil:  Maybe you can commit it directly by just pasting the changes in the relevant file in the GitHub editor?

alexanderstephan:
Alternatively, you should be able to do `git reset --hard origin/dev` and then you could just add your commit again. Then you can force push. Maybe a even better solution. :+1:

joschahenningsen (approved):
:ok_hand: :beers:

PR #310, opened 2022-02-27
Moderate-able chat

### LINKED ISSUE #293: Moderate Chat
- [x] message should be deletable by admins
- [ ] There should be an option to send all messages to moderators (admins) first and wait for their approval before broadcasting them.

MatthiasReumann:
- Closes #293
- Fixed the 'Send' button (I forgot to set the 'type' attributes properly, so a click on the send button wouldn't trigger a submit 😅)

https://user-images.githubusercontent.com/29633518/156117499-93036150-8b33-4a62-b048-2e757709bd4d.mov


web/template/admin/course_settings.gohtml
```diff
@@ -65,6 +65,15 @@
                         {{if not .}}checked{{end}}>
                 <span class="ml-2">Allow Anonymous Messages</span>
             </label>
+            <label class="inline-flex items-center pl-2" for="enChatMod">
+                <input :disabled="!enChat"
+                       class="w-auto form-checkbox"
+                       id="enChatMod" name="enChatMod"
+                       type="checkbox"
+                       {{if and . .ModeratedChatEnabled}}checked{{end}}
+                        {{if not .}}checked{{end}}>
```
alexanderstephan, line 74:
```suggestion
                       {{if not .}}checked{{end}}>
```

api/chat.go
```diff
@@ -118,6 +120,31 @@ func handleDelete(ctx tools.TUMLiveContext, msg []byte) {
 	broadcastStream(ctx.Stream.ID, broadcastBytes)
 }
 
+func handleApprove(ctx tools.TUMLiveContext, msg []byte) {
```
alexanderstephan, line 123: I think the error messages are outdated in this method.

model/chat.go
```diff
@@ -52,6 +52,9 @@ type Chat struct {
 	Admin            bool   `gorm:"not null;default:false" json:"admin"`
 	Color            string `gorm:"not null;default:'#368bd6'" json:"color"`
 
+	Visible   sql.NullBool `gorm:"not null;default:true" json:"-"`
```
alexanderstephan, line 55: Maybe this could use a comment why we need both `Visible` and `IsVisible`.

web/ts/watch.ts
```diff
@@ -154,3 +160,11 @@ export function sendMessage(message: string, anonymous: boolean, replyTo: number
         }),
     );
 }
+
+export async function fetchMessages(id: number) {
```
alexanderstephan, line 164: Way more readable to move this here :+1:

web/template/admin/course_settings.gohtml
```diff
@@ -65,6 +65,15 @@
...
+                <span class="ml-2">Activate Moderate Chat</span>
```
alexanderstephan: For consistency we could also say "Enable" here since we used it already for the chat and so on.
```suggestion
                <span class="ml-2">Enable Moderated Chat</span>
```

alexanderstephan (changes requested):
Works as expected for me, using the moderation feature feels really good. :+1: Code looks good to me overall. I added a few small comments.

alexanderstephan (approved):
Looks good, thanks! :heart:

PR #320, opened 2022-03-13
Improve watch page alignment

alexanderstephan:
**Initial problem**:
The download button was misaligned for smaller screens. Previous fixes killed the download icon in the course edit section.

image

**Description**:

- Properly aligns the download button.

- Moves the "stream actions" downwards when viewing the site on smaller screens. This prevents collision with longer lecture titles.
- Improves the design of the player a little bit by removing the margin around the player. This is still not perfect as the player is not 16:9 when the chat is open. 

**Screenshots**:
image
image


joschahenningsen (approved):
Nice, thanks! 🍻

PR #321, opened 2022-03-13
Make theater mode default

alexanderstephan:
**Description:**

To avoid unused space within the watch page, this PR sets the theater mode as a default for the player.

`{if $course.ChatEnabled}}md:w-4/6{{end}}` was removed since it is overwritten by the current theater mode implementation anyways and didn't have any noticeable effect.

**Steps for Testing:**

- Check that untoggling still works
- Check that it behaves correctly on different aspect ratios
- Check that it works on streams with a Chat

### DIFF
--- a/web/template/watch.gohtml
+++ b/web/template/watch.gohtml
@@ -28,14 +28,14 @@
                 To view this video please enable JavaScript.
             </p></noscript>
         <div id="watchContent"
-             class="transition-all motion-reduce:transition-none duration-300 ease w-full pt-3 px-2 md:px-8 mx-auto mx:h-full">
+             class="transition-all motion-reduce:transition-none duration-300 ease w-full p-4 mx-auto mx:h-full">
             <div class="p-4 bg-indigo-700 bg-opacity-20 border-t-4 border-indigo-700 w-full font-semibold text-2 rounded-t display-none"
                  x-show="paused" x-cloak>
                 <i class="fas fa-info-circle"></i> This lecture is currently paused.
             </div>
             <video
                     id="my-video"
-                    class="video-js w-full theater-mode"
+                    class="video-js w-full vjs-theater-mode theater-mode"
                     controls
                     preload="auto"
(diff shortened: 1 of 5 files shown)


web/template/watch.gohtml
```diff
@@ -28,14 +28,14 @@
                 To view this video please enable JavaScript.
             </p></noscript>
         <div id="watchContent"
-             class="transition-all motion-reduce:transition-none duration-300 ease w-full pt-3 px-2 md:px-8 {{if $course.ChatEnabled}}md:w-4/6{{end}} lg:w-8/12 2xl:max-w-screen-xl mx-auto mx:h-full">
+             class="transition-all motion-reduce:transition-none duration-300 ease w-full pt-3 px-2 md:px-8 mx-auto mx:h-full">
```
MatthiasReumann: Since we increase player size we could use `p-2` instead of `pt-3 px-2 md:px-8`.

MatthiasReumann (commented):
Works as expected, good job! 💯

I added a short comment for a minor adjustment.

web/template/watch.gohtml
alexanderstephan: Yep, that looks more uniform. :+1: I think this is fine for now, but in the future we could make the alignment really seamless like it is on Twitch e.g.

alexanderstephan: What do you think of `p-4`? Then it looks a bit better when using the smaller player, otherwise the gap to the navbar seems to narrow.

alexanderstephan: Maybe you can do a follow-up after we merged https://github.com/joschahenningsen/TUM-Live/pull/320, so it looks like the player snaps to all edges when using the theater mode? I am not that good with design. This would also need some changes to the alignment of the description and icons.

MatthiasReumann: A follow-up sounds like a good plan. Will do.

MatthiasReumann (approved):

PR #323, opened 2022-03-14
Add specific type and size to unique columns

joschahenningsen:
I (painfully) found out that some database systems don't play well with unique constraints when the size of a column is not set. In our case GORM decided to convert some fields to `longtext` which can not have a unique constraint. This killed prod when I tried to restart it (: :tada:

### DIFF
--- a/model/shortlink.go
+++ b/model/shortlink.go
@@ -6,6 +6,6 @@ import "gorm.io/gorm"
 type ShortLink struct {
 	gorm.Model
 
-	Link     string `gorm:"not null unique"`
+	Link     string `gorm:"type:varchar(256); unique; not null"`
 	CourseId uint   `gorm:"not null"`
 }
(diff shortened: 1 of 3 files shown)


alexanderstephan:
Some not so neat gorm quirks lately ... 
I wondered, whether this also fixes the issue where new keys in the user table are created on startup? Unfortunately, I am still running into this issue. :sweat:

joschahenningsen:
>  I wondered, whether this also fixes the issue where new keys in the user table are created on startup? Unfortunately, I am still running into this issue. sweat


Yes this should also be fixed then. @MatthiasReumann  @mono424, if you run into DB migration issues when starting up TUM-Live, try to navigate to the users table in DataGrip, refresh it and delete the duplicated keys.

alexanderstephan (approved):
Otherwise, this looks good. :+1:  If it works on production it's fine. I'll try to figure out what's causing the key creation problem on my end.

PR #325, opened 2022-03-16
@-ing others and chat cleanup

### LINKED ISSUE #304: @-ing others
I have seen people @-ing each other, and this doesn't work at the moment.

From the little I know, I think this might be a suitable approach:
- Implement an in-memory hash map which stores all not-disconnected users
- when an user types `^@.$`, display a list of usernames starting with the given letter and insert _@username_ at the current position of the input field (Are colors enough to differentiate two users with the same name? I guess so.)
- On the receiver side of things check if the message is addressed to the receiver. (Maybe the same way replies work with an attribute `addressedTo` and if `user.ID == addressedTo` make **@username** bold or something.)

MatthiasReumann:
- Users can now @ others by selecting one of their peers in the list either with their mouse or up and down arrows.
- New Disconnect Message: Less aggressive. 
- Chat Cleanup: Move javascript from .gohtml files to typescript classes
- Chat preprocessors: In the `Chat` class I added the attribute `preprocessors` which is an array of functions that are applied to the ChatMessage before displaying them with Alpine. 
- Merged Polls 

https://user-images.githubusercontent.com/29633518/159641138-c21b5f5b-89e5-4dc3-a71d-c867fd4f21f2.mov

https://user-images.githubusercontent.com/29633518/159640766-3731d730-90e5-4670-ae66-b7fdcc4f45c7.mov

https://user-images.githubusercontent.com/29633518/160349967-3d23c3b8-7190-4980-bf39-dfd298dfd118.mov


joschahenningsen:
Looks like some darn nice refactoring on the frontend! :tada: 
I feel like you should take a look here @mono424 and coordinate with each other regarding the polls.

api/chat.go
```diff
@@ -260,15 +267,26 @@ func getMessages(c *gin.Context) {
 	c.JSON(http.StatusOK, chats)
 }
 
+func getUsers(c *gin.Context) {
+	_, exists := c.Get("TUMLiveContext")
+	if !exists {
+		c.AbortWithStatus(http.StatusNotFound)
+		return
+	}
+	users := userStorage.GetAll()
```
joschahenningsen: Just a thought without understanding your code in detail yet:

Why not just allow @-ing users who already sent a message to the chatroom, this way we can use the chats table to get them (`select u.* from chats join users u on chats.user_id = u.id where stream_id = ...`) and don't need the `userStorage`, right? (Feel free to make the `UserID` in the Chat table a proper foreign key if that would make things easier)

MatthiasReumann: That is a smart idea. I guess I sometimes like to make things more complicated than they need to be 😆

joschahenningsen: Me too 😀 a second pair of eyes always helps when I start over engineering

mono424:
Really like the create new poll design enhancement! :)

web/template/partial/stream/chat/chat.gohtml
```diff
@@ -1,60 +1,61 @@
...
-         x-on:chatlike.window="e => {messages.find(m=>m.ID===e.detail.likes).likes=e.detail.num; $dispatch('reorder')}"
-         x-on:chatdelete.window="e => {messages.find(m=>m.ID===e.detail.delete).deleted=true;}"
-         x-on:chatnewpoll.window="e => { if (!anonymous) { pollResult = null; activePoll = { ...e.detail, selected: null }; } }"
-         x-on:polloptionvotesupdate.window="e => { activePoll = { ...activePoll, pollOptions: activePoll.pollOptions.map(pollOption => pollOption.id === e.pollOptionId ? { ...pollOption, votes: e.votes } : pollOption) }; }"
-         x-on:polloptionresult.window="e => { activePoll = null; pollResult = e.detail; }"
-         x-on:chatresolve.window="e => {messages.find(m=>m.ID===e.detail.resolve).resolved=true;}"
-         x-on:chatapprove.window="messages = await watch.fetchMessages({{$stream.ID}}); $nextTick(() => { watch.scrollToBottom() })"
-         @reorder="watch.sortMessages(messages, orderByLikes)"
+         x-data="watch.initChat({{.IsAdminOfCourse}},{{$stream.ID}}, {{$userId}},'{{$userName}}');"
+         x-init="await c.loadMessages(); await c.poll.load(); $nextTick(() => { watch.scrollToBottom() })"
```
mono424: Everything is so clean now !! 💯 
I would say you could consider here: `await Promise.all([c.loadMessages(), c.poll.load()]);`.

web/ts/chat/Poll.ts
```diff
@@ -0,0 +1,64 @@
...
+        startPoll(
+            this.question,
+            // @ts-ignore
+            this.options.map(({ answer }) => answer),
+        );
+        this.reset();
+    }
+
+    async load() {
+        await fetch("/api/chat/" + this.streamId + "/active-poll")
```
mono424: Again just something really minor but probably its simpler to stick to await:
```
const res = await fetch("/api/chat/" + this.streamId + "/active-poll");
this.activePoll = await res.json();
```
Just wanted to mention it, but I am also cool with the current approach :)

mono424 (approved):
Looks very good! 👍 
Added some comments, but just minor stuff.

MatthiasReumann:
> Really like the create new poll design enhancement! :)

Thanks! I wouldn't really call it enhancement since I simply put into the same container as the @-ing userlist ^^

mono424:
> > Really like the create new poll design enhancement! :)
> 
> Thanks! I wouldn't really call it enhancement since I simply put into the same container as the @-ing userlist ^^

looks much better omo! ;)

PR #326, opened 2022-03-16
Update Go to 1.18

joschahenningsen:


alexanderstephan (changes requested):
Really hyped for the new features! :partying_face: We should probably also update https://github.com/joschahenningsen/TUM-Live/blob/dev/.idea/TUM-Live-Backend.iml
Otherwise lgtm. :+1:

alexanderstephan (approved):

PR #327, opened 2022-03-19
Chat Polls

### LINKED ISSUE #56: Tweedback like chat features
Chat will most probably not be adopted by lecturers in it's current state. Some things it needs:

- [x] Moderation: Lecturers see messages first and can decide to approve or dismiss them.
- [x] Voting: Students can vote for messages/questions that are useful. These can be shown prominently.
- [ ] Quiz messages: Lecturers send simple multiple choice quizzes to the chat.

mono424:
closes #56


mono424:
`handleSubmitPollOptionVote` could debounce the broadcast with some 2 seconds delay, to have less traffic maybe.

api/chat.go
```diff
@@ -68,6 +76,106 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+	if err := json.Unmarshal(msg, &req); err != nil {
+		log.WithError(err).Warn("could not unmarshal submit poll answer request")
+		return
+	}
+	if ctx.User == nil {
+		return
+	}
+
+	if err := dao.AddChatPollOptionVote(req.PollOptionId, ctx.User.ID); err != nil {
+		return
```
joschahenningsen, line 91: Error could be logged

api/chat.go
```diff
@@ -68,6 +76,106 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+	voteCount, _ := dao.GetPollOptionVoteCount(req.PollOptionId)
+
+	voteUpdateMap := gin.H{
+		"pollOptionId": req.PollOptionId,
+		"votes":        voteCount,
+	}
+
+	if voteUpdateJson, err := json.Marshal(voteUpdateMap); err == nil {
+		broadcastStream(ctx.Stream.ID, voteUpdateJson)
+	}
```
joschahenningsen, line 106: else: log err

api/chat.go
```diff
@@ -68,6 +76,106 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+
+	voteCount, _ := dao.GetPollOptionVoteCount(req.PollOptionId)
+
+	voteUpdateMap := gin.H{
+		"pollOptionId": req.PollOptionId,
+		"votes":        voteCount,
+	}
+
+	if voteUpdateJson, err := json.Marshal(voteUpdateMap); err == nil {
+		broadcastStream(ctx.Stream.ID, voteUpdateJson)
```
joschahenningsen: This sends the current votes to all chat participants. Shouldn't only admins get them here?

api/chat.go
```diff
@@ -68,6 +76,106 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+			Answer: answer,
+		})
+	}
+
+	poll := model.Poll{
+		StreamID:    ctx.Stream.ID,
+		Question:    req.Question,
+		Active:      true,
+		PollOptions: pollOptions,
+	}
```
joschahenningsen: We could validate the length of the question and the answers here (e.g. not empty)

dao/chat.go
```diff
@@ -88,3 +96,39 @@ func GetAllChats(userID uint, streamID uint) ([]model.Chat, error) {
...
+// GetActivePoll returns the active poll for the stream with the given ID.
+func GetActivePoll(streamID uint) (model.Poll, error) {
+	var activePoll model.Poll
+	err := DB.Preload("PollOptions").First(&activePoll, "stream_id = ? AND active = true", streamID).Error
+	return activePoll, err
+}
+
+// CloseActivePoll closes poll for the stream with the given ID.
+func CloseActivePoll(streamID uint) error {
+	return DB.Table("polls").Where("stream_id = ? AND active = true", streamID).Update("active", false).Error
```
joschahenningsen:
```suggestion
	return DB.Table("polls").Where("stream_id = ? AND active", streamID).Update("active", false).Error
```

dao/chat.go
```diff
@@ -88,3 +96,39 @@ func GetAllChats(userID uint, streamID uint) ([]model.Chat, error) {
...
+}
+
+// GetPollUserVote returns the id of the PollOption that the user has voted for. If not vote was found then 0.
```
joschahenningsen:
```suggestion
// GetPollUserVote returns the id of the PollOption that the user has voted for. If no vote was found then 0.
```

dao/chat.go
```diff
@@ -88,3 +96,39 @@ func GetAllChats(userID uint, streamID uint) ([]model.Chat, error) {
...
+func GetPollUserVote(pollId uint, userId uint) (uint, error) {
+	var pollOptionIds []uint
+	err := DB.Table("poll_option_user_votes").Select("poll_option_user_votes.poll_option_id").Joins("JOIN chat_poll_options ON chat_poll_options.poll_option_id=poll_option_user_votes.poll_option_id").Where("poll_id = ? AND user_id = ?", pollId, userId).Find(&pollOptionIds).Error
+	if err != nil {
```
joschahenningsen, line 116:
```suggestion
	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
```

web/template/partial/stream/chat/chat.gohtml
```diff
@@ -39,6 +46,86 @@
...
         </div>
+        <!-- Active Poll UI -->
+        <template x-if="activePoll && !admin">
+            <div class="z-50 absolute w-full">
+                <div class="text-white bg-sky-600 dark:bg-sky-700 text-center p-2">Active Poll</div>
+                <div class="shadow-lg p-4 bg-white dark:text-white dark:bg-secondary-lighter border-b dark:border-gray-800 rounded-b-lg">
+                    <p x-text="activePoll.question" class="text-sm font-semibold mb-3"></p>
+                    <template x-for="pollOption in activePoll.pollOptions" :key="pollOption.ID">
+                        <button class="flex items-center mb-1" @click="activePoll.selected = pollOption.ID" :disabled="activePoll.submitted != 0">
+                            <i class="" :class="(pollOption.ID == activePoll.selected || activePoll.submitted == pollOption.ID) ? 'fas fa-check-circle' : 'far fa-circle'"></i>
```
MatthiasReumann: Works great, good job 👍🏻 

A tiny thing I noticed: You used `==` and `!=` instead of `===` and `!==` a couple of times in chat.gothml. This might result in weird behaviour in some edge cases.

api/chat.go
```diff
@@ -68,6 +76,106 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+		"votes":        voteCount,
+	}
+
+	if voteUpdateJson, err := json.Marshal(voteUpdateMap); err == nil {
+		broadcastStream(ctx.Stream.ID, voteUpdateJson)
+	}
+}
+
+func handleStartPoll(ctx tools.TUMLiveContext, msg []byte) {
+	var req startPollReq
```
joschahenningsen, line 116: if you only use the type once you can declare it here in the function scope (nit)

api/chat.go
```diff
@@ -68,6 +76,106 @@ func configGinChatRouter(router *gin.RouterGroup) {
...
+	for _, option := range poll.PollOptions {
+		pollOptionsJson = append(pollOptionsJson, option.GetStatsMap(0))
+	}
+
+	pollMap := gin.H{
+		"active":      true,
+		"question":    poll.Question,
+		"pollOptions": pollOptionsJson,
+		"submitted":   0,
+	}
```
joschahenningsen, line 162: Using gin.H instead of declaring a type is an interesting choice, fine by me though :)

api/chat.go
```diff
@@ -260,6 +368,51 @@ func getMessages(c *gin.Context) {
...
+		return
+	}
+
+	isAdminOfCourse := tumLiveContext.User.IsAdminOfCourse(*tumLiveContext.Course)
+	var pollOptions []gin.H
+	for _, option := range poll.PollOptions {
+		voteCount := int64(0)
+
+		if isAdminOfCourse {
+			voteCount, _ = dao.GetPollOptionVoteCount(option.ID)
```
joschahenningsen: The Error should be logged

dao/chat.go
joschahenningsen: This should work(?)

web/template/partial/stream/chat/chat.gohtml
```diff
@@ -39,6 +46,86 @@
                 </div>
             </div>
         </div>
+        <!-- Active Poll UI -->
+        <template x-if="activePoll && !admin">
```
joschahenningsen: this should also be hidden when the chat is collapsed

joschahenningsen (changes requested):
Looks great and works as expected, good job! Just a few thoughts

Also, this happens on the admin interface when a student submits the poll (the poll options disappear, perhaps because the admin interface doesn't have some attribute that is accessed here?):
image

web/template/partial/stream/chat/chat.gohtml
mono424: Yes cleaner to compare types as well, I agree! :)

api/chat.go
mono424: Yes you are right! I currently throw them away in front end for normal users which is actually not secure (if the data is confidential). I thought as the data will be published anyways, that would be the easiest approach, but it would be much better to only send it to admins, I will take a look how to do that :)

dao/chat.go
mono424: oh, sweet!

dao/chat.go
mono424: Is gorm really throwing a `gorm.ErrRecordNotFound` on a find with 0 results?
OMO that cannot happen (https://gorm.io/docs/error_handling.html):
```
GORM returns ErrRecordNotFound when failed to find data with First, Last, Take, if there are several errors happened, you can check the ErrRecordNotFound error with errors.Is, for example:
```

api/chat.go
mono424: yes nicer, but a little inconsistent with the other req types. But probably It's better to continue this way :)

web/template/partial/stream/chat/chat.gohtml
mono424: yes i def. missed that!! Thanks! 😅

api/chat.go
mono424: Okay well, that was easy, just replaced by `broadcastStreamToAdmins` 👍

mono424:
> Looks great and works as expected, good job! Just a few thoughts
> 
> Also, this happens on the admin interface when a student submits the poll (the poll options disappear, perhaps because the admin interface doesn't have some attribute that is accessed here?): image

my fault i missed a `activePoll.` on `pollOptions` i guess, should be fixed now :)

joschahenningsen (approved):
Amazing, thanks!

PR #337, opened 2022-03-31
live-update title and description for viewers on update

joschahenningsen:
I noticed lecturers adding the tweedback link when the stream was already running. Users had to reload the page to see the link. This PR addresses this issue.

https://user-images.githubusercontent.com/44805696/161003720-bef59d62-ffd6-4e33-a99c-403670661f1e.mp4

I'm duplicating some logic in alpine and in gohtml here to support people with js turned off.

### DIFF
--- a/api/courses.go
+++ b/api/courses.go
@@ -420,13 +420,14 @@ type addUnitRequest struct {
 }
 
 func updateDescription(c *gin.Context) {
-	var req renameLectureRequest
-	jsonData, err := ioutil.ReadAll(c.Request.Body)
+	sIDInt, err := strconv.Atoi(c.Param("streamID"))
 	if err != nil {
 		c.AbortWithStatus(http.StatusBadRequest)
 		return
 	}
-	if err := json.Unmarshal(jsonData, &req); err != nil {
+	sID := uint(sIDInt)
+	var req renameLectureRequest
+	if err := c.Bind(&req); err != nil {
 		c.AbortWithStatus(http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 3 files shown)


api/courses.go
```diff
@@ -440,16 +441,25 @@ func updateDescription(c *gin.Context) {
 		c.AbortWithStatusJSON(http.StatusInternalServerError, "couldn't update lecture Description")
 		return
 	}
+	wsMsg := gin.H{
```
alexanderstephan, line 444: What is the meaning of `ws` in this context?

web/template/watch.gohtml
```diff
@@ -200,18 +200,18 @@
...
-                <div x-data="{less: {{$isMinLen}}}" class="text-3 p-4 border-t-2 my-3 border-secondary video-description">
-                    <span class="font-semibold">Video Description:</span>
-                    <div :class="less && 'truncate-children'">{{.Description}}</div>
-                    <template x-if="{{$isMinLen}}">
-                        <button class='text-xs text-gray-500 dark:text-gray-400 uppercase mt-2'
-                                @click="less = !less" x-text="less ? 'show more' : 'show less'">show less
-                        </button>
-                    </template>
+            <div x-data="{descriptionlen: `{{len .Description}}`, less:true}"
+                 class="text-3 p-4 border-t-2 my-3 border-secondary video-description">
```
alexanderstephan: Not sure if this was intended or not: Before, the `div` was completely hidden, now the `border-secondary` is always visible, also when there is no description. I would also remove border this since this makes the page scrollable on my 16:9 screen.

Reference:
image

web/template/watch.gohtml
```diff
@@ -200,18 +200,18 @@
...
+                <span class="font-semibold"
+                      x-show="description.length>0">{{if .Description}}Video Description:{{end}}</span>
+                <div :class="less && 'truncate-children'"
```
alexanderstephan, line 206: Nit: When there are new lines or paragraphs in the description, every paragraph gets truncated and still appears. We probably want to cut off the message independent of the formatting.

Reference:
image

alexanderstephan (changes requested):
Really nice feature, works as expected! :100:  Code also looks good, I just noticed some possibly unwanted behaviors.

web/template/watch.gohtml
joschahenningsen: I noticed that to, but as it's the current behavior as well and not caused by this pr I'd make a separate issue for this.

api/courses.go
joschahenningsen: WebSocket (it's intended to be sent through the websocket) But yeah, this doesn't really need to be in the variables name.

web/template/watch.gohtml
joschahenningsen: Thanks! Didn't notice that, should be fixed :+1:

api/courses.go
alexanderstephan: Yeah, that's fine anyways.

web/template/watch.gohtml
alexanderstephan: Looks perfect now, thanks!

alexanderstephan (approved):
Looks good! :beer:

PR #338, opened 2022-03-31
Don't send private or unnecessary data in chat user search

joschahenningsen:
This PR adds a dto that just contains the user id and name for the chat user search.
This prevents sending data like `createdAt`, `administeredCourses` or similar as they are not needed here.

### DIFF
--- a/api/chat.go
+++ b/api/chat.go
@@ -411,12 +411,12 @@ func getUsers(c *gin.Context) {
 		return
 	}
 	type chatUserSearchDto struct {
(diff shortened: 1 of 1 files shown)


MatthiasReumann (approved):
Great idea! I thought I dealt with that with `json: "-"`, but `createdAt` and similar isn't covered by that (Besides `administeredCourses`,  which I totally forgot 😅). 

Plus, this makes it easier to understand what is sent to the frontend. 👍

PR #340, opened 2022-03-31
Use JSON Web Tokens for authentication

joschahenningsen:
This PR removes the old cookie store for authentication and replaces it with :sparkles: shiny :sparkles:  JWTs for enhanced security. 

Since this will log out all users, this is a good time to also fetch the surnames of users from LDAP when they log in. We want to use them in the admin interface to add them to courses. (this was recently requested.)

This also fixes an issue where the watch page fails rendering completely when the chat is enabled but the user is not signed in.


api/users.go
```diff
@@ -117,9 +118,10 @@ func SearchUser(c *gin.Context) {
 }
 
 type userForLecturerDto struct {
-	ID    uint   `json:"id,omitempty"`
-	Name  string `json:"name,omitempty"`
-	Login string `json:"login,omitempty"`
+	ID       uint    `json:"id,omitempty"`
+	Name     string  `json:"name,omitempty"`
```
alexanderstephan, line 122: Maybe we can call this `FirstName` to clear things up? This would then match the `firstName` in the `LdapResp`

go.mod
```diff
@@ -38,6 +36,8 @@ require (
 	mvdan.cc/xurls/v2 v2.4.0
 )
 
+require github.com/golang-jwt/jwt v3.2.2+incompatible
```
alexanderstephan: It looks this is an pretty old version of JWT. v4.0+ would have proper go module support. Maybe we can use that?

tools/middlewares.go
```diff
@@ -5,8 +5,8 @@ import (
 	"TUM-Live/model"
 	"errors"
 	"github.com/getsentry/sentry-go"
-	"github.com/gin-contrib/sessions"
 	"github.com/gin-gonic/gin"
+	"github.com/golang-jwt/jwt"
```
alexanderstephan:
```suggestion
	"github.com/golang-jwt/jwt/v4"
```
As per [Migration Guide](https://github.com/golang-jwt/jwt/blob/main/MIGRATION_GUIDE.md). I saw that this deprecates `*jwt.StandardClaims` for example. This should get taken a look at, in case we want to actually update `jwt`.

web/ts/courseAdminManagement.ts
```diff
@@ -1,6 +1,7 @@
 class User {
     id: number;
     name: string;
```
alexanderstephan, line 3: For consistency this could possibly be `firstName` as well

config.yaml
```diff
@@ -1,9 +1,9 @@
 alerts:
   matrix:
-    username: "rbgbot"
-    password: "password"
-    homeserver: "matrix.org"
-    roomId: "!id:matrix.org" # Obtained in Element by "Room Settings" -> "Advanced" -> "Internal Room ID"
+    homeserver: matrix.org
+    password: password
+    roomid: '!id:matrix.org'
+    username: rbgbot
```
alexanderstephan, line 6: Maybe we want to keep the comment. There are different way to get something called `roomId`, but only copying like described in the comment works.

alexanderstephan (changes requested):
That looks a really nice security update. :+1: I couldn't test this quite so thoroughly as I think I am missing some secrets. but I assume that it works. I noticed that we use an older/deprecated version of JWT, we might want to use the current version. Otherwise just some nits. Code quality looks very good otherwise. :100:

api/users.go
joschahenningsen: That would probably clear things up :+1: but I avoid migrating the database wherever I can -.- :smile:

go.mod
joschahenningsen: Huh, no idea how this slipped through, thx!

config.yaml
joschahenningsen: Sure! This got lost when the config was written after generating the key.

api/users.go
alexanderstephan: Ah, if this can cause issues, we better leave it as it is.

joschahenningsen: It doesn't really cause issues, but it sure is annoying :D 
That's on me though, for not thinking forward to this

joschahenningsen:
Decided to **not** refactor name -> firstName as migrating the database is sort of annoying and also, `Name` is not too bad of a name for the variable, since it's the primary name, we use to refer to users and display them (e.g. chat, greeting message etc.)

PR #343, opened 2022-03-31
Mark videos as watched

### LINKED ISSUE #302: Checkbox to mark watched videos on the overview page for a course

alexanderstephan:
Fixes #302.

**Design**:

Peek 2022-04-01 23-02

### DIFF
--- a/api/progress.go
+++ b/api/progress.go
@@ -59,7 +59,8 @@ func (b *progressBuffer) run() {
 	}
 }
 
-// configProgressBuffer configures the progress buffer
+// configProgressRouter sets up the router and initializes a progress buffer
+// that is used to minimize writes to the database.
 func configProgressRouter(router *gin.Engine) {
 	progressBuff = newProgressBuffer()
 	go progressBuff.run()
@@ -67,8 +68,8 @@ func configProgressRouter(router *gin.Engine) {
 	router.POST("/api/watched", markWatched)
 }
 
-// ProgressRequest corresponds the request that is sent by the video player when it reports its progress for VODs
-type ProgressRequest struct {
+// progressRequest corresponds the request that is sent by the video player when it reports its progress for VODs
+type progressRequest struct {
 	StreamID uint    `json:"streamID"`
 	Progress float64 `json:"progress"` // A fraction that represents currentTime / totalTime for a given video
 	// Note: To be able to save the progress, we also need the userID, but it`s already contained in the Gin context
@@ -76,7 +77,7 @@ type ProgressRequest struct {
 
 // saveProgress saves progress to a buffer that is flushed at a fixed interval.
 func saveProgress(c *gin.Context) {
-	var request ProgressRequest
+	var request progressRequest
 	err := c.BindJSON(&request)
 
 	if err != nil {
@@ -100,18 +101,18 @@ func saveProgress(c *gin.Context) {
 	})
 }
 
-// WatchedRequest corresponds the request that is sent when a user marked the video as watched on the watch page.
-type WatchedRequest struct {
+// watchedRequest corresponds the request that is sent when a user marked the video as watched on the watch page.
+type watchedRequest struct {
 	StreamID uint `json:"streamID"`
 	Watched  bool `json:"watched"`
 }
 
 // markWatched marks a VoD as watched in the database.
 func markWatched(c *gin.Context) {
-	var request WatchedRequest
+	var request watchedRequest
 	err := c.BindJSON(&request)
 	if err != nil {
-		log.WithError(err).Warn("Could not bind JSON.")
+		log.WithError(err).Error("Could not bind JSON.")
 		c.AbortWithStatus(http.StatusBadRequest)
 		return
 	}
@@ -124,7 +125,6 @@ func markWatched(c *gin.Context) {
 		c.AbortWithStatus(http.StatusForbidden)
 		return
 	}
-
 	progress := model.StreamProgress{
 		UserID:   tumLiveContext.User.ID,
 		StreamID: request.StreamID,
@@ -132,7 +132,7 @@ func markWatched(c *gin.Context) {
 	}
 	err = dao.SaveProgresses([]model.StreamProgress{progress})
 	if err != nil {
-		log.WithError(err).Warn("Could not mark VoD as watched.")
+		log.WithError(err).Error("Could not mark VoD as watched.")
 		c.AbortWithStatus(http.StatusInternalServerError)
 		return
 	}
--- a/api/progress.go
+++ b/api/progress.go
@@ -93,19 +93,10 @@ func saveProgress(c *gin.Context) {
 		c.AbortWithStatus(http.StatusForbidden)
 		return
 	}
-	streams, err := dao.GetStreamsByIds([]uint{request.StreamID})
-
-	if err != nil {
-		log.WithError(err).Warn("Could not get stream.")
-		c.AbortWithStatus(http.StatusInternalServerError)
-		return
-	}
-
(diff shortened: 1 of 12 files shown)


model/progress.go
```diff
@@ -1,9 +1,11 @@
...
 type StreamProgress struct {
-	Progress float64 `gorm:"not null"`
-	// We need to use a primary key in order to use ON CONFLICT in dao/progress.go, same as e.g. https://www.sqlite.org/lang_conflict.html.
-	StreamID uint `gorm:"primaryKey"`
-	UserID   uint `gorm:"primaryKey"`
+	Progress float64 `gorm:"not null"`               // The progress of the stream as represented as a floating point value between 0 and 1.
+	Watched  bool    `gorm:"not null;default:false"` // Whether the user has marked the stream as watched.
+	StreamID uint    `gorm:"primaryKey"`
+	UserID   uint    `gorm:"primaryKey"`
+	CourseID uint    `gorm:"primaryKey"`
```
joschahenningsen: Why add the course id? We can join this in sql using the streams table 

Sort of like this: `select stream_progresses.*, s.course_id from stream_progresses join streams s on s.id = stream_id`

web/course.go
```diff
@@ -94,15 +96,81 @@ func CoursePage(c *gin.Context) {
 		c.AbortWithStatus(http.StatusInternalServerError)
 		return
 	}
+
+	progs, err := dao.LoadProgressesForCourseAndUser((*tumLiveContext.Course).ID, (*tumLiveContext.User).ID)
+	if err != nil {
+		sentry.CaptureException(err)
+		log.Printf("%v", err)
```
joschahenningsen:
```suggestion
		log.WithError(err).Error("loading progresses for course and user failed")
```

web/course.go
```diff
@@ -94,15 +96,81 @@ func CoursePage(c *gin.Context) {
...
+	}
+
+	progressResponses, progressStreams := prepareCourseProgressData(tumLiveContext, progs)
+
+	encoded, err := json.Marshal(progressResponses)
+	if err != nil {
+		sentry.CaptureException(err)
+		log.Printf("%v", err)
```
joschahenningsen: log.WithError ...

api/progress.go
```diff
@@ -71,33 +74,86 @@ type ProgressRequest struct {
...
 		UserID:   tumLiveContext.User.ID,
+		CourseID: streams[0].CourseID,
 	})
 }
+
+// WatchedRequest corresponds the request that is sent when a user marked the video as watched on the watch page.
+type WatchedRequest struct {
+	StreamID uint `json:"streamID"`
+	Watched  bool `json:"watched"`
+}
```
joschahenningsen: This could be in the markWatched function or unexported (nit)

api/progress.go
```diff
@@ -71,33 +74,86 @@ type ProgressRequest struct {
...
+	foundContext, exists := c.Get("TUMLiveContext")
+	if !exists {
+		return
+	}
+	tumLiveContext := foundContext.(tools.TUMLiveContext)
+	if tumLiveContext.User == nil {
+		c.AbortWithStatus(http.StatusForbidden)
+		return
+	}
+	streams, err := dao.GetStreamsByIds([]uint{request.StreamID})
```
joschahenningsen: Why not `getStreamByID`?

web/course.go
```diff
@@ -94,15 +96,81 @@ func CoursePage(c *gin.Context) {
...
 	}
 }
 
+func prepareCourseProgressData(tumLiveContext tools.TUMLiveContext, streamProgresses []model.StreamProgress) ([]ProgressInfo, []ProgressStream) {
+	var progressStreams []ProgressStream
+	var progressResponses []ProgressInfo
+
+	courseStreams := (*tumLiveContext.Course).Streams
+	// Combine streams with existing progresses.
+	for _, stream := range courseStreams {
```
joschahenningsen: It seems a little like this function could be a (granted, rather large) sql statement, did you try that or do am I thinking in the wrong direction?

web/course.go
```diff
@@ -94,15 +96,81 @@ func CoursePage(c *gin.Context) {
...
+	err = templ.ExecuteTemplate(c.Writer, "course-overview.gohtml",
+		CoursePageData{IndexData: indexData, Course: *tumLiveContext.Course, StreamsWithProgress: progressStreams, ProgressResponses: string(encoded)})
 	if err != nil {
 		sentrygin.GetHubFromContext(c).CaptureException(err)
 	}
 }
 
+func prepareCourseProgressData(tumLiveContext tools.TUMLiveContext, streamProgresses []model.StreamProgress) ([]ProgressInfo, []ProgressStream) {
+	var progressStreams []ProgressStream
+	var progressResponses []ProgressInfo
```
joschahenningsen: Can we use `make(...)` to allocate the required memory beforehand?

web/course.go
```diff
@@ -94,15 +96,81 @@ func CoursePage(c *gin.Context) {
...
 	// in any other case assume either validated before or public course
-	err := templ.ExecuteTemplate(c.Writer, "course.gohtml", CoursePageData{IndexData: indexData, Course: *tumLiveContext.Course})
+	err = templ.ExecuteTemplate(c.Writer, "course-overview.gohtml",
+		CoursePageData{IndexData: indexData, Course: *tumLiveContext.Course, StreamsWithProgress: progressStreams, ProgressResponses: string(encoded)})
 	if err != nil {
 		sentrygin.GetHubFromContext(c).CaptureException(err)
 	}
 }
 
+func prepareCourseProgressData(tumLiveContext tools.TUMLiveContext, streamProgresses []model.StreamProgress) ([]ProgressInfo, []ProgressStream) {
```
joschahenningsen: IMO this method is a little hard to wrap my head around (especially because of its complex signature), could you add a comment explaining what it does?

joschahenningsen (commented):
Looks good and works as expected! I just think some of the backend logic could be a little simplified (e.g. by preparing some of the data in sql)

web/course.go
alexanderstephan: No, this could be (probably) done in SQL. I just was not sure if this is worth the pain :joy: But I'll give it a try. :+1:

web/course.go
alexanderstephan: We would have calculate the number of recordings then. But that would work. :+1:

model/progress.go
alexanderstephan: At first I thought it would be better if everything needed for the progress calculations is in one struct. But I see your point. :+1:

api/progress.go
alexanderstephan: Good point, I forgot to change it after refactoring.

model/progress.go
joschahenningsen: You can also scan your query result into a custom struct

web/course.go
joschahenningsen: Nah no worries, feel free to leave this as is 😀

alexanderstephan:
I moved a lot of stuff to the dao. I think it is more readable now. I would like to avoid the code duplication, but `gorm` seems to be struggling with multiple models in one `struct`. I would be glad about any clues on how I can make this prettier.

web/course.go
alexanderstephan: I didn't use make as most lectures only have like 20 streams maximum. I don't think really think it's needed here and has the potential to cause problems with my implementation.

web/course.go
alexanderstephan: It's gone now for good. :yum:

joschahenningsen (approved):
thx :)

PR #353, opened 2022-04-04
Mini UI Fixes for @-ing others

MatthiasReumann:
- Remove Safari `focus:outline`
- Fix the buggy scrollbar 
- Fix that sometimes the user-list didn't disappear when the current message was empty.
- Removed some code
- Moved emojis to input-box (just as polls)
- Fix combined use of emojis and at-ing (didn't really work smoothly before)


joschahenningsen (approved):
Looks good, ty!

PR #356, opened 2022-04-06
add creator of course to course admins and filter duplicate courses

### LINKED ISSUE #354: Owner can also be admin and sees course twice
I created a new course (id 283) and then noticed there were no admins, so I added myself as admin. Now I see the the course twice on the admin management panel to the right.

image

Removing myself from the admins again still lets me have admin privileges (I assume because I'm the creator/owner):

image

**Expected behavior**

The owner of a course should be a super-role of the admin and displayed as "owner" or regular "admin" in the list of "Course Admins" in the admin panel.

joschahenningsen:
fixes #354 

When a lecturer creates a course, `course.user_id` is set to their ID, when they then add themselves to the courses admins, they see the course twice under `my courses` and in the admin panel.  This PR makes sure that
- Creators are automatically added to the admin list as expected
- Duplicates caused by `course.user_id` are filtered

This also fixes the staticcheck action to work with go 1.18


joschahenningsen:
It seems like the staticcheck action doesn't support go 1.18 yet, I opened an issue upstream: https://github.com/dominikh/staticcheck-action/issues/5

PR #365, opened 2022-04-15
Refactor mark watched

alexanderstephan:
This is a follow-up to #343.

- I adjusted the design of the button a bit: 
grafik.
So an open eye should mean show in this case and so on. This might be a bit more intuitive than just calling it filter.

- Rewrote the database query. From my testing it takes <1ms now which should be good enough.
- Refactored everything in a way that a stream can hold all the necessary information, which cleans up the implementation by a lot.
- A bit more dynamic course page layout e.g. when there was no next lecture it doesn't show "Next lecture" without a date anymore.

### DIFF
--- a/web/template/course-overview.gohtml
+++ b/web/template/course-overview.gohtml
@@ -24,16 +24,18 @@
                 <div class="flex justify-between h-16 px-3 border-b dark:border-gray-800">
                     <span class="bg-green-500 rounded-full text-lg font-bold text-white px-5 my-auto">VoDs</span>
                     <div class="my-auto">
+                        {{if .Course.HasRecordings}}
                         <button class="hover:bg-gray-200 dark:hover:bg-gray-600 rounded px-2"
                                 @click="asc = !asc; global.reorderVodList()">
                             <span class="text-sm font-semibold uppercase dark:text-white"
                           x-text="asc ? '&#x25B2; asc' : '&#x25BC; desc'">desc &#x25BC;</span>
                         </button>
-                        {{if .IndexData.TUMLiveContext.User}}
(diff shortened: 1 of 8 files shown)


web/template/course-overview.gohtml
```diff
@@ -45,6 +47,7 @@
             </div>
         </div>
         <!-- Next Lecture -->
+        {{if $course.HasNextLecture}}
```
MatthiasReumann, line 50: Since it's hidden (Great idea btw!) we could get rid of one if-branch in stream.FriendlyNextDate:
```
func (s Stream) FriendlyNextDate() string {
 // start remove
	if now.With(s.Start).Before(time.Now()) {
		return "No upcoming stream"
	}
// end
	if now.With(s.Start).EndOfDay() == now.EndOfDay() {
		return fmt.Sprintf("Today, %02d:%02d", s.Start.Hour(), s.Start.Minute())
	}
	if now.With(s.Start).EndOfDay() == now.With(time.Now().Add(time.Hour*24)).EndOfDay() {
		return fmt.Sprintf("Tomorrow, %02d:%02d", s.Start.Hour(), s.Start.Minute())
	}
	return s.Start.Format("Mon, January 02. 15:04")
}
```

MatthiasReumann (commented):
Works as expected, 💯. The button with the eye-thingy looks great!

I added a tiny suggestion.

web/template/course-overview.gohtml
alexanderstephan: Good catch, I removed it. :+1:

joschahenningsen (approved):
Looks great, thanks!

PR #368, opened 2022-04-18
Don't show download button when there are no files present

joschahenningsen:
This PR also fixes the `video_only` mode


alexanderstephan (approved):
Looks good to me :ok_hand:

PR #372, opened 2022-04-19
Reorder planned courses

### LINKED ISSUE #370: Upcoming lectures are sorted in the wrong way
The lectures should be sorted such that the next lectures are on top of the page:

image

alexanderstephan:
Fixes https://github.com/joschahenningsen/TUM-Live/issues/370

This seems the most simple way to fix this.  Reording once in the beginning does not have any impact on loading as far as I could tell.

Alternatively, we could implement a method for a course that sorts the the VoDs or do another data base query and attach the result to the CoursePageContext.

### DIFF
--- a/dao/courses.go
+++ b/dao/courses.go
@@ -6,7 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"time"
-  
+
 	"github.com/RBG-TUM/commons"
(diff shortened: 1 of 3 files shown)


joschahenningsen (approved):
Straightforward changes and works as expected, thanks for fixing this so quickly!

PR #385, opened 2022-04-23
User Management Pagination

### LINKED ISSUE #362: Pagination on user management page
The user management page contains quite a few users already. They should be split in chunks of about 10-20 users with pagination below the table to navigate them.

MatthiasReumann:
- Closes #362 
- Added User pagination
- Moved add-user-form out of table


joschahenningsen (approved):
At some point we should probably dynamically load only the users displayed to prevent the DOM from becoming excessive and unresponsive but for our current use case this is perfectly fine! :)

PR #386, opened 2022-04-23
Remove curl

### LINKED ISSUE #382: tools/camera shells out to curl
https://github.com/joschahenningsen/TUM-Live/blob/dev/tools/camera/camera.go#L92-L100

There is no need to require `curl`, the `net/http` package should be sufficient to make these requests.

MatthiasReumann:
- Closes #382 
- Uses github.com/icholy/digest for digest authentication


tools/camera/camera.go
```diff
@@ -87,30 +90,40 @@ func (c Camera) GetPresets() ([]model.CameraPreset, error) {
...
+		req, err = http.NewRequest("POST", fmt.Sprintf("http://%s%s", c.Ip, url), bytes.NewReader([]byte(body)))
 	default:
 		return nil, fmt.Errorf("unsupported protocol: %v", method)
 	}
-	var bts []byte
-	buf := bytes.NewBuffer(bts)
-	camCurl.Stdout = buf
-	err := camCurl.Start()
 	if err != nil {
+		log.WithError(err).Error("could not create http request")
```
joschahenningsen: This module is intended to be somewhat importable, instead of logging errors we should return them with some extra context so they can be handled further up in the call chain:

```suggestion
		return nil, fmt.Errorf("create http request: %v", err)
```

tools/camera/camera.go
```diff
@@ -87,30 +90,40 @@ func (c Camera) GetPresets() ([]model.CameraPreset, error) {
...
+	}
+
+	res, err := client.Do(req)
+	if err != nil {
+		log.WithError(err).Error("1")
```
joschahenningsen: This should also return the error

tools/camera/camera.go
```diff
@@ -87,30 +90,40 @@ func (c Camera) GetPresets() ([]model.CameraPreset, error) {
...
 		return nil, err
 	}
-	err = camCurl.Wait()
+	defer res.Body.Close()
+
+	bts, err := io.ReadAll(res.Body)
 	if err != nil {
+		log.WithError(err).Error("2")
```
joschahenningsen: And here

joschahenningsen: (also what excacly does 2 mean here?)

joschahenningsen (changes requested):
Looks great except for one small remark :)

tools/camera/camera.go
MatthiasReumann: Ah god, forgot to remove that. Means nothing.

joschahenningsen (approved):
Looks good, thanks!

PR #394, opened 2022-04-26
DAO mocks and tests

### LINKED ISSUE #388: Make dao mockable & modular
Testing is not easy when a database is involved. To mitigate this, we should move forward like this when writing dao code:

1. Create a file for a set of data we want to work with, e.g. `dao/pizza.go`
2. Create an interface with the methods we want to use. (this interface can be mocked later):

```go
type PizzaDao interface {
	Create(pizza *model.Pizza) error
	Get(id uint) (model.Pizza, error)
	Delete(id uint) error
	// whatever you need, e.g. All() or something specific like TastyPizzas(vegan bool)
}
```
3. Create an implementation of this interface, along with a `new` function
```go
type pizzaDao struct { //private struct with concrete implementation
	db *gorm.DB
}

func NewPizzaDao() PizzaDao { //exported(public) function returning the concrete implementation of the exported interface
	return pizzaDao{db: DB}
}
```
4. Write the actual dao functions for the struct so it implements the interface:
```go
func (d pizzaDao) Create(pizza *model.Pizza) error {
	return d.db.Create(pizza).Error
}

func (d pizzaDao) Get(id uint) (pizza model.Pizza, err error) {
	return pizza, d.db.First(&pizza).Error
}

func (d pizzaDao) Delete(id uint) error {
	return d.db.Delete(&model.Pizza, id).Error
}
```
5. You can now use the implementation like this:
```go
pizza := model.Pizza{type: "Funghi"}
err := dao.NewPizzaDao().Create(&pizza)
```
6. We can mock this interface using https://github.com/golang/mock like this: 
```bash
mockgen -source=dao/pizza.go -destination=mock_dao/pizza.go
```

These mocks can be used in place of the real dao in tests. The first PR implementing this is #387. We should also migrate other dao files if we have some spare time for it.

MatthiasReumann:
- Closes #388 
- Generate mocks with `go generate ./...` in `src/`
- Some tests


joschahenningsen:
> Do we want global DAO variables such as chatDao? Alternatively, we need to call New[Name]Dao every time we need those functions. That's especially interesting for writing tests to inject the mock.

Not sure which is cleaner but exported global vars in the dao package that are also initialized there seem fine. this way we could call `dao.Chat.Get(42)`

joschahenningsen:
Or maybe a method `dao.Chat()` that creates a singleton sort of like this

```go
var chat *chatDao

func Chat() ... {
    if chat == nil {
        chat = New...
    }
    return chat
}
```

MatthiasReumann:
> > Do we want global DAO variables such as chatDao? Alternatively, we need to call New[Name]Dao every time we need those functions. That's especially interesting for writing tests to inject the mock.
> 
> Not sure which is cleaner but exported global vars in the dao package that are also initialized there seem fine. this way we could call `dao.Chat.Get(42)`

Yup, that was my initial idea also (Is also the way I currently implemented it). Singleton might also be a good idea. Will look into it, thanks!

joschahenningsen:
I hope you automated this somehow, looks like a hell lot of work :100:

MatthiasReumann:
> I hope you automated this somehow, looks like a hell lot of work 💯

:-)

MatthiasReumann:
I've added a single test for the /api/worker route. Maybe take a quick look at the implementation. But basically what I do is to inject the DAO into the configRouter function. With that we can test all API functions with ease. (i guess)

api/worker_test.go
```diff
@@ -0,0 +1,50 @@
+package api
+
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/golang/mock/gomock"
+	"github.com/joschahenningsen/TUM-Live/mock_dao"
+	"github.com/joschahenningsen/TUM-Live/model"
+	"github.com/joschahenningsen/TUM-Live/tools"
+	"github.com/magiconair/properties/assert"
```
joschahenningsen, line 10: Looks like a great opportunity to use https://github.com/stretchr/testify

api/router.go
```diff
@@ -20,6 +21,6 @@ func ConfigGinRouter(router *gin.Engine) {
 	configProgressRouter(router)
 	configServerNotificationsRoutes(router)
 	configTokenRouter(router)
-	configWorkerRouter(router)
+	configWorkerRouter(router, dao.NewWorkerDao())
```
joschahenningsen: what do we do if a router needs multiple daos? pass them all here? Create a struct in the dao package that contains all sub-daos for the routes? Any idea which might be the best option?

joschahenningsen (commented):
I'm a huge fan of the approach, there still are some questions to consider

api/router.go
MatthiasReumann: I feel like a struct containing all DAOs would be the best option. Each router receives a pointer to that struct object and uses the DAOs as needed. Another benefit of that approach is that if some endpoint ever needs other DAO objects we don't have to add additional parameters.

api/worker_test.go
MatthiasReumann: Right now, I am only using the` assert.Equals` function, do you have any particular features in mind?

MatthiasReumann:
For documentation purposes I think it would be quite nice to explain what this PR changes:

1. Each `config<Name>Router` now receives a `dao.DaoWrapper` object which contains all Dao objects (e.g. `LectureHallsDao`, `UsersDao`, ...)
2. All API functions now dangle from a `<API-name>Routes` struct which contains a `dao.DaoWrapper` so that all API functions can use the injected DAO.
3. Middleware functions which need DAOs such as `InitStream` or `InitCourse` now are in the form of:

```
func InitCourse(wrapper dao.DaoWrapper) gin.HandlerFunc {
 	return func(c *gin.Context) { ... }
}
```
4. There are no global DAOs anymore.
5. `go generate ./...` in `/src` generates all DAO mock files in `/mock_dao`.

joschahenningsen (approved):
:beers:

PR #395, opened 2022-04-27
Slide upload for presentations etc

MatthiasReumann:
- Add upload input field on admin page
- Download button on stream page

[OpenSource Lab]


joschahenningsen:
Nice Idea, before you start implementing this: Can you write a short proposal about what changes you'll need to make considering api endpoints and especially database models? We could probably utilize the `File` model with some adaptions?

MatthiasReumann:
**Backend:**
API Endpoint: `/api/streams/:id/downloads`
- GET: Retrieve the paths to all downloadable files associated with streamId `:id`
- POST: Add file to stream with `:id`

API Endpoint: `/api/streams/:id/downloads/:did`
- DELETE: Delete file with id `:did`

Reusing the file model sounds like a good idea. I think there are two ways we can handle this (Which one do you prefer?): 
1. Add a column file-type (`select * from files where filetype = ....`). 
2. Use GORM's [`embedded`](https://gorm.io/docs/models.html#Embedded-Struct) feature and create a "wrapper" model for slide uploads. (`select * from streamdownloads`). This way we can easily add more columns for download-related stuff. 

**Frontend**
_Player:_ 
Add download button to the arrays of buttons beneath the video player. If more than one download is available a dropdown will open with a list of all downloadable files. Preferably rendered with go templates.

_Admin:_ 
The Admin upload shall be a simple button which opens the system's native filesystem dialog and uploads the chosen files. Alternatively, a url which points to a file (check for the ending `.pdf`?) can be used to add a file. This fits quite nicely with the `File` implementation which has an attribute `Path`. Here we need to make sure that there is some information which describes if the path is fully-qualified or relative, e.g. `uploads.pizza/files/1234` vs `/files/1234 `. This distinction should be achievable with javascript and regex only. 

**Further Considerations**
_Lecture Series Action_
Some _"prefer"_ to only provide slide decks with 1200 slides+, which then would stay the same throughout the semester, so we need to make sure to add a lecture series action too. 

_File types_ 
PDF Upload only? 

_Size limit_ 
As described above some slide decks have an enormous amount of slides. For example, the current slide deck of Introduction to the Theory of Computation has a size of 9.8MB.

joschahenningsen:
I never used [embedded](https://gorm.io/docs/models.html#Embedded-Struct). Feel free to give it a quick try and post what it does to the database internally. Does it really create a `streamdownloads` table?

> Add download button to the arrays of buttons beneath the video player. If more than one download is available a dropdown will open with a list of all downloadable files. Preferably rendered with go templates.

The design could be merged with the video downloads (if enabled and available)

> Some "prefer" to only provide slide decks with 1200 slides+

Ofc. it would be nice if the file would then only be stored once and referenced by all lectures. 

> File types

I'm fine with allowing more, I can see reasons to upload pdf, docx, xlsx, csv and even zip files. The question would be whether or not course admins are trusted parties. 

> Size limit

Not a concern for me. As long as the files are saved somewhere in `tools.Cfg.Paths.Mass`. How about 50 MB for PDF files?

joschahenningsen:
I had another thought regarding the design and Imo this should be as minimal as possible. How about allowing to drop files into the lecture description input which generates a url for the file that is inserted into the markdown as a link like on github?

https://user-images.githubusercontent.com/44805696/166140380-2000f74c-7892-4d55-947f-6c812b23bad3.mp4

MatthiasReumann:
> I had another thought regarding the design and Imo this should be as minimal as possible. How about allowing to drop files into the lecture description input which generates a url for the file that is inserted into the markdown as a link like on github?
> 
>  out.mp4

This sounds like a great idea. Will look into it. 💯

MatthiasReumann:
> I never used [embedded](https://gorm.io/docs/models.html#Embedded-Struct). Feel free to give it a quick try and post what it does to the database internally. Does it really create a `streamdownloads` table?
> 
> > Add download button to the arrays of buttons beneath the video player. If more than one download is available a dropdown will open with a list of all downloadable files. Preferably rendered with go templates.
> 
> The design could be merged with the video downloads (if enabled and available)
> 
> > Some "prefer" to only provide slide decks with 1200 slides+
> 
> Ofc. it would be nice if the file would then only be stored once and referenced by all lectures.
> 
> > File types
> 
> I'm fine with allowing more, I can see reasons to upload pdf, docx, xlsx, csv and even zip files. The question would be whether or not course admins are trusted parties.
> 
> > Size limit
> 
> Not a concern for me. As long as the files are saved somewhere in `tools.Cfg.Paths.Mass`. How about 50 MB for PDF files?

**Short update:** I ended up reusing the `Files` model, hence I didn't need `embedded`. Which is extra nice because then we can utilize the `/api/download/:fileid` route. However, I added a `Filename` attribute, so that I can store the files as `<uuid>.<file-extension>`, since if we store all files at the same place, same names would simply overwrite the other file and that way the download name equals the original file's name.

joschahenningsen:
> since if we store all files at the same place, same names would simply overwrite the other file

We store the stream recordings in `tools.Cfg.Mass/course.Year/course.Semester/<date>`, how about storing the files somewhere like `tools.Cfg.Mass/course.Year/course.Semester/files/<uuid>.<file-extension>`

MatthiasReumann:
Additionally, to what we discussed earlier, I added a `fileType` attribute to differentiate between `attachments` and `stream-download-files`.

api/stream.go
```diff
@@ -257,3 +267,83 @@ func deleteVideoSection(c *gin.Context) {
...
+	course := *tumLiveContext.Course
+
+	var path string
+	var filename string
+
+	switch c.Query("type") {
+	case "file":
+		file, err := c.FormFile("file")
+		if err != nil {
+			c.AbortWithStatusJSON(http.StatusBadRequest, "missing form parameter 'file'")
```
joschahenningsen, line 318:
```suggestion
			c.AbortWithStatusJSON(http.StatusBadRequest, "missing form parameter 'file'")
			return
```

joschahenningsen: There's a few aborts without a return, this might lead to unwanted behavior.

model/file.go
```diff
@@ -5,11 +5,18 @@ import (
 	"strings"
 )
 
+const (
+	FILETYPE_DOWNLOAD   = 0
+	FILETYPE_ATTACHMENT = 1
```
joschahenningsen: This is how we emulated enums in go before:

```suggestion
	FILETYPE_DOWNLOAD   = iota + 1 // <- this resolves to 1
	FILETYPE_ATTACHMENT // <- and this becomes 2
```

model/file.go
```diff
@@ -5,11 +5,18 @@ import (
...
+)
+
 type File struct {
 	gorm.Model
 
 	StreamID uint   `gorm:"not null"`
 	Path     string `gorm:"not null"`
+	Filename string `gorm:"null;"`
+	Type     uint   `gorm:"default: 0"`
```
joschahenningsen:
```suggestion
	Type     uint   `gorm:"not null;default: 1"`
```

model/file.go
```diff
@@ -5,11 +5,18 @@ import (
...
+	FILETYPE_DOWNLOAD   = 0
+	FILETYPE_ATTACHMENT = 1
+)
+
 type File struct {
 	gorm.Model
 
 	StreamID uint   `gorm:"not null"`
 	Path     string `gorm:"not null"`
+	Filename string `gorm:"null;"`
```
joschahenningsen: What does `gorm:"null"` do?
If it should be nullable, this should be a pointer.

model/file.go
```diff
@@ -28,5 +35,13 @@ func (f File) GetFriendlyFileName() string {
...
 	}
+
+	if f.Filename != "" {
+		return f.Filename
+	}
 	return "Default view"
 }
+
+func (f File) IsAbsolutePath() bool {
+	return strings.HasPrefix(f.Path, "https://") || strings.HasPrefix(f.Path, "http://")
```
joschahenningsen: Shouldn't `IsAbsolutePath` check if f.Path has the prefix `/`?

go.mod
```diff
@@ -41,6 +41,7 @@ require (
 	github.com/RBG-TUM/commons v0.0.0-20220406105618-030c095f6a1b
 	github.com/crewjam/saml v0.4.6
 	github.com/golang/mock v1.6.0
+	github.com/google/uuid v1.1.2
```
joschahenningsen, line 44: we already have `github.com/satori/go.uuid` in place, can you use

```go
import (
    uuid "github.com/satori/go.uuid"
)
```
instead to avoid including redundant dependencies?

joschahenningsen (changes requested):
Looks good, the only thing I noticed is, that attachments are also displayed in the stream download dropdown.

MatthiasReumann:
> Looks good, the only thing I noticed is, that attachments are also displayed in the stream download dropdown.

Oh, forgot to change that. :( Thanks for pointing out!

model/file.go
MatthiasReumann: _Absolutely_ bad naming from my side 😅 How about `IsURL`?

joschahenningsen: IsURL sounds great :+1: but if you want to verify that it's a valid url, I'd suggest using the `net/url` package, parse the string into a url, check parsing errors and the urls protocol.

MatthiasReumann: Sounds like a great idea. Will do!

joschahenningsen (approved):
Forgot to merge, sorry :D Looks good!

PR #396, opened 2022-04-27
Search through lectures

MatthiasReumann:
- Add search button on course page
- Add route /search?q=<query string>
- Initally search for titles, sections, descriptions 

[OpenSource Lab]


joschahenningsen:
https://mariadb.com/kb/en/full-text-index-overview/ might me helpful here. https://nav.tum.sexy uses Meilisearch as search engine. Worth a look but imo that's overkill here, even when searching through chats or auto-generated transcripts.

MatthiasReumann:
**Update/Documentation:** 
I added a new API Endpoint `/api/search/streams/<searchDomain>` with `<searchDomain>` := `sections` | `streams` | `chats` so that in the future one could easily add another search endpoint like `/api/search/courses` and don't have to worry about bad naming. 

Each streams-search Endpoint returns a JSON encoded object of 
```
type searchResponse struct {
	StreamIds []uint
}
```

I've utilised the full-text-search capabilities of MariaDB, which meant indices to columns were added. 

Frontend-wise, the client makes calls the respective endpoints, merges the results and makes a call to `/api/streams/:streamID` and displays that information. I've updated the `/api/streams/:streamID`-Endpoint insofar that non-admins can also call `GET` to retrieve a reduced version of the stream JSON object:

```
response =
  gin.H{"name": stream.Name,
  "description":  stream.Description,
  "friendlyTime": stream.FriendlyTime()}
```

**UI:**
Screenshot 2022-05-15 at 10 51 37
Screenshot 2022-05-15 at 10 53 00


**Follow-up:**
Once I've done enough research for voice-to-text, I will start a another PR which addresses the search in that regard.

model/stream.go
```diff
@@ -24,8 +24,8 @@ const (
 type Stream struct {
 	gorm.Model
 
-	Name             string
-	Description      string
+	Name             string `gorm:"index:,class:FULLTEXT"`
+	Description      string `gorm:"type:text;index:,class:FULLTEXT"`
```
joschahenningsen, line 28: Just wondering: Why does this need `type:text`?

web/ts/stream-search.ts
```diff
@@ -0,0 +1,77 @@
...
+            this.timeout = setTimeout(() => {
+                Promise.all([
+                    this.search(q, SearchDomain.VideoSections),
+                    this.search(q, SearchDomain.Streams),
+                    this.search(q, SearchDomain.Chats),
+                ]).then((searchResponses) => {
+                    this.loading = false;
+                    this.results = this.getResults(searchResponses.map((sr) => sr.StreamIds));
+                });
+            }, 250);
```
CommanderStorm: Is the Timeout here really intended?
> The `setTimeout()` method calls a function after a number of milliseconds

so as far as I read this, there is an delay of 250ms between the user inputting something and the search being executed?

web/ts/stream-search.ts
```diff
@@ -0,0 +1,77 @@
...
+        } else {
+            this.results = [];
+            this.loading = false;
+        }
+    }
+
+    private async search(q: string, searchDomain: SearchDomain): Promise<searchResponse> {
+        return await fetch(`/api/search/streams/${searchDomain}?q=${q}&courseId=${this.courseId}`)
```
CommanderStorm: you might want to use something like a cached fetch (see https://github.com/TUM-Dev/navigatum/blob/db0a6e7c0662c1e6211052eb18ebdefefbf3d0c1/webclient/src/core.js#L8 for inspiration) to improve the search experience. 
In a context where responsiveness matters, doing unnessesary work seems wrong.

web/template/course-overview.gohtml
```diff
@@ -6,10 +6,50 @@
...
+    <div class="w-full mt-5 md:mt-24 mx-auto md:w-2/4 bg-white border shadow rounded-lg h-fit dark:border-gray-600 dark:bg-secondary">
+        <div class="flex px-3 py-1 border-b"
+             @click.outside="showLectureSearchModal = false; streamSearch.reset()">
+            <div class="p-1 my-auto">
+                <i x-cloak x-show="!streamSearch.loading"
+                   class="fa-solid fa-search text-lg text-gray-500 dark:text-gray-600"></i>
+                <i x-cloak x-show="streamSearch.loading"
+                   class="fas fa-spinner animate-spin text-lg text-gray-500 dark:text-gray-600"></i>
+            </div>
+            <input @keyup="(e) => streamSearch.performSearch(e.target.value)"
```
CommanderStorm: You may want to include some extra keys on keydown
See https://github.com/TUM-Dev/navigatum/blob/db0a6e7c0662c1e6211052eb18ebdefefbf3d0c1/webclient/src/modules/autocomplete.js#L95

web/template/course-overview.gohtml
```diff
@@ -6,10 +6,50 @@
...
+                   name="lecture-search-input"
+                   placeholder="Search for lectures"
```
CommanderStorm: Kind of personal preference
```suggestion
                   placeholder="Search for a lecture"
```

web/ts/stream-search.ts
```diff
@@ -0,0 +1,77 @@
...
+    async performSearch(q: string) {
+        if (q !== "") {
+            clearTimeout(this.timeout);
+
+            this.loading = true;
+            this.timeout = setTimeout(() => {
+                Promise.all([
+                    this.search(q, SearchDomain.VideoSections),
+                    this.search(q, SearchDomain.Streams),
+                    this.search(q, SearchDomain.Chats),
```
CommanderStorm: I find it kind of strange to impose a fixed order of search results.
For NavigaTUM we did a lot of work on this, while for this code this seems to be hardcoded. This may mean that less relevant search results come first..

I can however see that this is a good shortcut, for implementing this, if this feature is not meant to be something core of the platform.
I would classify this feature as a "nice to have", thus this shortcut is imo okay.

This could also be because of the quality/speed tradeoff of the fulltext search never returns less relevant search results. Or the added complexity is would not really be worth the added tech dept.
In these cases a sourcecode comment documenting this would be appreciated xD

web/ts/stream-search.ts
```diff
@@ -0,0 +1,77 @@
...
+    results: object[];
+
+    constructor(courseId: number, slug: string) {
+        this.courseId = courseId;
+        this.results = [];
+        this.loading = false;
+    }
+
+    getResults(results: number[][]): object[] {
+        const all = []; // create key value pair from array of array
```
CommanderStorm: This should be a better documentation of this function
```suggestion 
        // flatten array of array, preserving only the first occurrence of each streamId
        const all = [];
```

CommanderStorm (commented):
Note that these are just some comments from a fellow radom internet stranger (read: I am not a maintainer of this project).
- I am not qualified to talk about Go, so I am not Go-ing to do that
- nor am I particularly qualified to talk about Typescript
- I have not tested your code (I just had a look), so this is not an attestation, that this works

I have left some inline comments.

About the engenering:
- I find it a bit strange to not directly fetch the data you would like to deliver in the search results in the search endpoint, but this may be a good idea, depending on how much you see this endpoint being used.
- collecting multiple requests in JS seems also kind of strange imo

web/ts/stream-search.ts
MatthiasReumann: The Timeout is there so that the function isn't called after every `keyup`, but rather the last. 
So for example if I type `h-e-l-l-o`:
After `h`,`e`,`l`,`l` `clearTimeout(timout) `resets the timeout, and after the letter `o` it executes the callback of `setTimeout`.

web/ts/stream-search.ts
MatthiasReumann: I generally think that the (limited) full-text-search of MariaDB is only an intermediate solution and something like ElasticSearch, Meilisearch, or similar is needed if, as you say, search becomes an integral part of TUMLive. 

However, I could add the `relevance` (a float value) to the `searchResponse` object, which would deal with the problem of your first paragraph.

joschahenningsen: relevance-wise I think lecture-name > description > chat. Weighing this with some factor and the relevance from mariadb could work out quite well. If we really want to create a real search engine for TUM-Live, we could have a look at https://blevesearch.com/ (But in a follow up :D).

Also I agree with Frank, we should have this be one request instead of three. Maybe by adding the searchDomains to the URL like this: `?searchDomains=a,b,c` or by adding a `all` search domain.

web/ts/stream-search.ts
joschahenningsen: How expensive is the search request? I don't think it would be a problem if this was fetched on every keystroke.  But please make sure, that the request is only sent if the input has changed, not on keys like `esc`, `shift`, `tab` etc.

api/stream.go
```diff
@@ -200,17 +200,33 @@ func (r streamRoutes) getStream(c *gin.Context) {
...
-			"ingest":      fmt.Sprintf("%sstream?secret=%s", tools.Cfg.IngestBase, stream.StreamKey),
-			"live":        stream.LiveNow,
-			"vod":         stream.Recording})
+
+	if stream.Name == "" {
+		stream.Name = stream.GetName()
+	}
+
+	var response gin.H
+	if tumLiveContext.User != nil && tumLiveContext.UserIsAdmin() {
```
joschahenningsen: The "larger" response should be sent to lecturers of the course as well, not only server admins. Also this needs a comment or two, which different purposes the if and else branches serve. 
But ideally, as stated below we don't need this endpoint.

dao/chat.go
```diff
@@ -193,3 +196,18 @@ func (d chatDao) ToggleLike(userID uint, chatID uint) error {
 func (d chatDao) CloseActivePoll(streamID uint) error {
 	return DB.Table("polls").Where("stream_id = ? AND active", streamID).Update("active", false).Error
 }
+
+func (d chatDao) Search(q string, courseId uint) ([]uint, error) {
```
joschahenningsen: Can you add a doc- comment and/or name the return params?

dao/streams.go
```diff
@@ -415,3 +416,17 @@ func (d streamsDao) DeleteStreamsWithTumID(ids []uint) {
 		return nil
 	})
 }
+
+func (d streamsDao) Search(q string, courseId uint) ([]uint, error) {
```
joschahenningsen: Can you add a doc- comment and/or name the return params?

dao/video-section.go
```diff
@@ -37,3 +40,18 @@ func (d videoSectionDao) GetByStreamId(streamID uint) ([]model.VideoSection, err
 	err := DB.Order("start_hours, start_minutes, start_seconds ASC").Find(&sections, "stream_id = ?", streamID).Error
 	return sections, err
 }
+
+func (d videoSectionDao) Search(q string, courseId uint) ([]uint, error) {
```
joschahenningsen: Here as well: Can you add a doc- comment and/or name the return params?

joschahenningsen (changes requested):
Looks quite good to me, just some comments. Also thanks @CommanderStorm for the input :blush:

joschahenningsen:
Three more things that concern me: 
- The server is a little slam hit with three requests for each query plus all requests for the streams that are found. IMO, the requests should be combined into one (as stated above) and already return the required information instead of just the stream ids. 
- The search bar needs some styling for the dark mode
- I have a video section named `a` but the search returns nothing when I search for `a`, I also have two lectures called `VL 1` and `VL 2` respectively, the search returns nothing for `VL`. Could you look into that?

MatthiasReumann:
> 

Is `VL1` and `VL2` a VoD? Since, currently,  it only displays VoD results.

joschahenningsen:
> Is VL1 and VL2 a VoD? Since, currently, it only displays VoD results.

yes, I'm testing on the database referenced in the readme and applied no changes to it except the automigration from this pr

model/stream.go
MatthiasReumann: `FULLTEXT` does only work on `CHAR`, `VARCHAR`, AND `TEXT`

MatthiasReumann:
**Update:** 

First all thanks to both of you for the great feedback ❤️ @CommanderStorm @joschahenningsen  

- Added SearchDAO: One query for everything (Endpoint: `/api/search/streams`), no domains anymore, sorted by relevance, no `GET` requests on frontend side.
- Added Cached Fetch
- Added dark mode styling 
- Added Keydown & Keyup support

joschahenningsen (approved):
Sorry for the long wait, this looks good! Can you just restore the run configuration?

joschahenningsen:
Thanks! :)

PR #397, opened 2022-04-27
improve look of embeded videos

joschahenningsen:
This PR adds a titlebar that allows navigating to the stream and course on TUM-Live:
image

And an overlay for streams that have not yet started:
image


alexanderstephan (approved):
Code and design looks really clean! :ok_hand:

Issue #358, opened 2022-04-13
User settings page


joschahenningsen:
TUM-Live should have a section for users to edit their preferences. We could replace the logout button with the user name and have the logout action and settings as a dropdown item.

Settings might include: 
- setting a preferred name for the chat
- ...? (open for ideas, feel free to edit)

PR #402, opened 2022-04-29
Changed "Moin" to "Servus" on the Start Page

EweLo:
Why? Because the TUM is located in Bavaria, not in Hamburg. Regarding the fact that most people here say "Servus" and not "Moin" for "Hello", I think this would be an appropriate change to make.

To be clear, this was not my idea:
Unbenannt

### DIFF
--- a/web/template/index.gohtml
+++ b/web/template/index.gohtml
@@ -21,7 +21,7 @@
     {{if not .TUMLiveContext.User}}
         <p class="text-2"><a href="/login" class="underline">Log in</a> to see your courses</p>
     {{else if .TUMLiveContext.User.Name}}
-        <p class="text-2">Moin {{.TUMLiveContext.User.Name}}, nice to see you!</p>
+        <p class="text-2">Servus {{.TUMLiveContext.User.Name}}, nice to see you!</p>
     {{end}}
     {{if .LiveStreams}}
         <h2 class="text-2xl text-1">Active Livestreams</h2>


Raupinger:
+1 for this

ro-i:
This is definitely not a good change. Never break the user, says Linus Torvalds.

heiko123abc:
I think we should let the user decide, maybe do it as a setting.

talemke:
We should give it an extra tweak and change it to:
> Servus, %s, nice to see you! 🥨

EweLo:
> We should give it an extra tweak and change it to:
> 
> > Servus, %s, nice to see you! 🥨

> Servus, %s, nice to see you! 🥨🍺 *

joschahenningsen:
Y'all think this is a coincidence? 😀

https://www.kartezumnamen.eu/index.php?sur=Henningsen&s=Suchen

EweLo:
> Y'all think this is a coincidence? 😀
> 
> https://www.kartezumnamen.eu/index.php?sur=Henningsen&s=Suchen

Yeah, this explains it 😁

joschahenningsen:
Probably something that could be addressed in #358

joschahenningsen:
:*

halbGefressen:
How about a custom greeting, where the user can enter a string like "Hello %s, nice to see you!". I think this would solve all regional issues. Edit: that was already suggested, my bad

ro-i:
Come on, @halbGefressen, you just want to enter `drop table students; --` and see what happens :D

joschahenningsen:
Feel free to 
`curl -X POST 'https://live.rbg.tum.de/api/users/settings/greeting' -H 'Content-Type: application/json' -H 'Cookie: jwt=this-is-secret-abc-123' -d '{"value":"Ahoi Matrose"}'`
(who needs server-side validation anyway? :smiley:)

PR #403, opened 2022-05-01
Add hot reload support

### LINKED ISSUE #366: Add Hot-Reload support
TUM-Live Server should serve the current GOHTML pages when they are changed. 
Currently a server restart is needed.

mono424:
- Hot reload of .gohtml and static files.
- Added tailwind jit to `npm run build-dev` to support updated tailwind CSS files on .gohtml change.

https://user-images.githubusercontent.com/17506411/166139202-858e3990-c8aa-4e17-8810-f850fecc2aa9.mov


tools/template_executor.go
```diff
@@ -0,0 +1,44 @@
+package tools
```
joschahenningsen, line 1: I think this could be in the web package or a sub package of web, or is it used somewhere else?

joschahenningsen (approved):
Code looks good and works flawless, amazing! Just one thought that is not important though :)

joschahenningsen:
If you merge dev, the staticcheck passes again :)

tools/template_executor.go
mono424: In `web` and in `tools.Middlewares` should I still move it, I would be okay with both actually as it is much more used in web than in tools 👍

mono424:
@joschahenningsen I merged dev. Dev & this branch not running locally because of `error="open fullchain.pem: no such file or directory"`. Should work tho, I think its my local env. :)

tools/template_executor.go
joschahenningsen: You're right :+1:

joschahenningsen:
@mono424 I merged again and forgot to comment out the saml config, it should run again now :)

PR #426, opened 2022-05-03
add starter db to docker command in README

joschahenningsen:
This command should be enough to set up a ready to go database to start working on TUM-Live

### DIFF
--- a/.idea/sqldialects.xml
+++ b/.idea/sqldialects.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="SqlDialectMappings">
+    <file url="file://$PROJECT_DIR$/docs/static/tum-live-starter.sql" dialect="MariaDB" />
+  </component>
+</project>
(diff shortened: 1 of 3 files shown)


MW360:
Looks good!
I was thinking about using `"$PWD"` to make the code executable right away.
But this may leads to problems in several shells.
An alternative would be to specify the shell with `#!/bin/bash`

so either:

```bash
docker run --detach \
  --name mariadb-tumlive \
  --env MARIADB_USER=root \
  --env MARIADB_ROOT_PASSWORD=example \
  --restart always \
  -p 3306:3306 \
  --mount type=bind,source="$PWD"docs/static/tum-live-starter.sql,target=/init.sql \
  mariadb:latest --init-file /init.sql
```

or

```bash
#!/bin/bash
docker run --detach \
  --name mariadb-tumlive \
  --env MARIADB_USER=root \
  --env MARIADB_ROOT_PASSWORD=example \
  --restart always \
  -p 3306:3306 \
  --mount type=bind,source="$(pwd)"docs/static/tum-live-starter.sql,target=/init.sql \
  mariadb:latest --init-file /init.sql
```

What do you think about that?

joschahenningsen:
Nice idea :ok_hand: I guess most (linux) shells should be fine with this

PR #429, opened 2022-05-04
Enhancement: follow system theme

### LINKED ISSUE #133: Toggle Light/Dark Mode Based on System Theme

just-max:
More or less like in #134 .

**All** of the code is still in the header because I wasn't sure how to get access to the Alpine.js object from the webpack.

Naming convention: "theme mode": whether the theme is set manually to light/dark or to automatic (that's what github calls it).

Visually I tried to follow the rest of the header's design language: hovering increases contrast (text-5 to text-1), the dropdown is basically the same as the one for notifications and the selected mode is marked with a darker background as on the semester selection widget.

Outstanding:
- is there a way to notify users? or should they just notice by themselves :)

Next steps?
- GitHub has a nice smooth transition between light and dark modes. Just setting `transition-colors duration-200` on the body doesn't really do it (not everything transitions smoothly, and it conflicts with the use of transitions for hovering).

resolves #134, resolves #133

### DIFF
--- a/web/template/headImports.gohtml
+++ b/web/template/headImports.gohtml
@@ -11,11 +11,6 @@
         <script defer src="/static/node_modules/alpinejs/dist/cdn.min.js"></script>
     {{end}}
     <script src="/static/assets/ts-dist/global.bundle.js?v={{if .}}{{.}}{{else}}development{{end}}"></script>
-    <script> // init in dom to prevent screen flickering
-        let darkTheme = localStorage.getItem("darkTheme") ? JSON.parse(localStorage.getItem("darkTheme")) : true;
-        if (!darkTheme) {
-            document.documentElement.classList.remove("dark");
-        }
-    </script>
+    {{template "theme-selector-head"}}
     <style>[x-cloak] { display: none !important; }</style>
 {{end}}
--- a/web/template/header.gohtml
+++ b/web/template/header.gohtml
@@ -87,25 +87,16 @@
                 <i class="fas fa-info-circle w-8"></i>
                 <span class="ml-2 text-sm">About</span>
             </a>
-            {{/* TODO: color scheme for mobile */}}
-            <div class="flex justify-end px-5 py-2 border-b dark:border-gray-600">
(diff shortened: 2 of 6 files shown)


joschahenningsen:
Amazing, thanks a ton! This works as expected. 

> All of the code is still in the header

Imo absolutely fine to do it like this.

> is there a way to notify users?

I can create a notification for the tiny bell.

image

It feels a little unintuitive to have a screen as a theme switcher button. Could we instead 
- show the current icon based on the system preference if system is selected (e.g. moon if my system prefers dark mode) **or**
- always show a generic icon like `fa-circle-half-stroke` if system is selected?

just-max:
:+1: to changing the switcher icon. I implemented suggestion 1:

auto_dark

joschahenningsen (approved):
Works and looks great, thanks a ton! :100:

just-max:
Thanks, no problem :)

PR #431, opened 2022-05-05
Replace name with lrz ID on welcome message

ghost:
Moin $deadname, nice to see you!

### DIFF
--- a/web/template/index.gohtml
+++ b/web/template/index.gohtml
@@ -20,8 +20,8 @@
     {{end}}
     {{if not .TUMLiveContext.User}}
         <p class="text-2"><a href="/login" class="underline">Log in</a> to see your courses</p>
-    {{else if .TUMLiveContext.User.Name}}
-        <p class="text-2">Moin {{.TUMLiveContext.User.Name}}, nice to see you!</p>
+    {{else if .TUMLiveContext.User.LrzID}}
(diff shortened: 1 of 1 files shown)


joschahenningsen:
Is the thought behind this that the name you're seeing on the start page is not the one you actually go by? Or is there a legitimate reason you need to see your lrz id?

The issue here is that not all users have an lrz id, if this was only about **not** getting called $deadname, we should work on #358 soon and let users decide how they want to be called.

just-max:
I know that the Informatics Zulip server has already made it possible to use a preferred name. Since #358 seems to be a ways off, maybe we could have something temporary?

Since the name is also displayed in the chat, this can't just be a client side change. Would changing it then require approval from an administrator (like on the Zulip server)? Of course it should be as easy as possible, but if left open people would probably start using nicknames, or worse >:(

Alternatively I could pick up #358 (but it would take a while longer, though)

joschahenningsen:
I hear Y'all, being called a name you don't identify with sucks. I'll move #358 up my priority list and start working on this!

joschahenningsen:
I thought of preventing misuse by limiting the name changing to something like once per month or semester.

just-max:
Ahh thanks :)
Sounds good, I think students will be reasonable with something like that.

joschahenningsen:
fyi, I removed the name from the start page until #462 is merged :)

PR #435, opened 2022-05-06
Extend issue reports

alexanderstephan:
This PR extends the interface for the matrix bot by allowing custom fields.

Additionally, a second matrix room can be configured. The goal is to use one room for important alerts, and the other room as logger (with notifications turned off).

If the person who reported the issue is not a lecturer, it will only send it to the log room. Only once there are two consecutive reports within 10 minutes, it will be also sent to the alert room. Once notified via the alert room, one can look into the log room, to find further information.

I also optimized the design of the admin buttons and modals for mobile use.

**Design**:

grafik

Students will find a small button in the stream icons (warning symbol) where they can send a warning. Lecturers still have the big button in the admin stream actions.


dao/statistics.go
```diff
@@ -100,6 +101,13 @@ func (d statisticsDao) GetCourseStatsHourly(courseID uint) ([]Stat, error) {
 	return res, err
 }
 
+// GetCurrentLiveViewers returns the number of viewers currently watching a live stream.
+func (d statisticsDao) GetCurrentLiveViewers(streamID uint) (int, error) {
+	var res int
+	err := DB.Raw(`SELECT viewers FROM stats WHERE stream_id = ? AND live = 1`, streamID).Scan(&res).Error
```
joschahenningsen: This should give you only the latest stat entry:

```suggestion
	err := DB.Raw(`SELECT viewers FROM stats WHERE stream_id = ? AND live = 1 ORDER BY id DESC LIMIT 1`, streamID).Scan(&res).Error
```

tools/bot/bot.go
```diff
@@ -0,0 +1,172 @@
...
+	if botInfo.PhoneNumber != "" {
+		infoText += "<tr><th>Phone</th><td>" + botInfo.PhoneNumber + "</td></tr>"
+	}
+	infoText += "</table>"
+
+	return infoText
+}
+
+// getFormattedMessageText generates a HTML styled message bot info
+func getFormattedMessageText(message string) string {
```
joschahenningsen, line 126: Do we need to create HTML in `GenerateInfoText` and then convert it to HTML again using the markdown renderer?

Maybe `GenerateInfoText` could return markdown instead?

tools/bot/bot.go
```diff
@@ -0,0 +1,172 @@
...
+// hasPrio returns true if 1% of the current viewers of a stream with streamID reported an issue.
+// When there threshold for sending an alert is greater than 1, it is also checked whether these reports are consecutive.
+func hasPrio(streamID uint, statsDao dao.StatisticsDao) bool {
+	liveViewers, err := statsDao.GetCurrentLiveViewers(streamID)
+	if err != nil {
+		sentry.CaptureException(err)
+		log.WithError(err).Error("Failed to get current live viewers")
+		return false
+	}
+	distinctReports := liveViewers
```
joschahenningsen: why is the number of reports initialized as the number of live viewers instead of `len(issuesPerStream[streamID])`? This might need a comment.

joschahenningsen (changes requested):
Looks great, just two or so questions :)

dao/statistics.go
alexanderstephan: Thanks, nice catch!

tools/bot/bot.go
alexanderstephan: I suppose, we don't even need the markdown, only the user input should get sanitized. Element does not support advances markdown formatting like tables e.g.

tools/bot/bot.go
alexanderstephan: I agree, this needs a bit of rework to be more logical :sweat_smile:

alexanderstephan: Yeah, I think that was just some last minute refactoring mistake, it should be `len(issuesPerStream[streamID])` of course

alexanderstephan:
I would like to do some testing on the test server, so this should not get merged yet

alexanderstephan:
Streaming works now, we could do one more run with alerts configured, but I assume this is fine anyways.

joschahenningsen (approved):
Looks great, thanks!

PR #439, opened 2022-05-09
330 ics file for course

### LINKED ISSUE #330: .ics file for course
It would be really nice if it were possible to download a .ics file for a course in a semester.

ata-keskin:
+ Added the file download_ics that is responsible for generating the ics file and making it available for downloading.
+ Added new go template ics.gotemplate for this purpose.
+ Added a new button on to the home page of courses for downloading the ics file.

Sidenotes:
Instead of using the already existing struct CalendarResult and the template ical.gotemplate, I opted to create new ones. The old struct CalendarResult and my struct CalendarEntry can be unified without a lot of problem, but I think the templates serve different purposes. The one I created is more human friendly IMO, but I'd also like to hear your opinions on this.


api/download_ics.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	if len(s.RoomCode) > 0 {
+		location += s.RoomCode
+	}
+	return CalendarEntry{
+		CreatedAt:   s.CreatedAt.Format(layout),
+		Start:       s.Start.Format(layout),
+		End:         s.End.Format(layout),
+		ID:          s.StreamKey,
+		Url:         "https://live.rbg.tum.de/course/" + strconv.Itoa(c.Year) + "/" + c.TeachingTerm + "/" + c.Slug,
```
joschahenningsen: Can you please use `tools.Cfg.WebUrl` as base?

api/download_ics.go
```diff
@@ -0,0 +1,104 @@
...
+		location += s.RoomName + " "
+	}
+	if len(s.RoomCode) > 0 {
+		location += s.RoomCode
+	}
+	return CalendarEntry{
+		CreatedAt:   s.CreatedAt.Format(layout),
+		Start:       s.Start.Format(layout),
+		End:         s.End.Format(layout),
+		ID:          s.StreamKey,
```
joschahenningsen: The streamKey is secret, it's used to authenticate with obs to ingest a stream, only lecturers and admins are allowed to see it, please use something like s.ID instead

api/download_ics.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/joschahenningsen/TUM-Live/dao"
+	"github.com/joschahenningsen/TUM-Live/model"
+	"github.com/joschahenningsen/TUM-Live/tools"
+	"html/template"
+	"net/http"
+	"strconv"
+)
+
+func configGinDownloadICSRouter(router *gin.Engine) {
+	router.GET("/api/download_ics/:year/:term/:slug", downloadICS)
```
joschahenningsen: `/api/course/:year/:term/:slug/events.ics` would be a little nicer semantically

joschahenningsen: You can group this and attatch the middleware `tools.InitCourse` like here

https://github.com/joschahenningsen/TUM-Live/blob/1f90e3841bfa43f7731b1b61e9103c8d28329738/api/lecture_halls.go#L35

This way you can get the course from the context in the route instead of manually using the params. (However, the middleware checks if the user requesting the file is allowed to see the course which makes it infeasible if we want calendar software to be able to fetch the file)

api/download_ics.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	course, err := dao.GetCourseBySlugYearAndTerm(c, slug, term, year)
+	if err != nil {
+		c.AbortWithStatus(http.StatusBadRequest)
+		return
+	}
+
+	if len(course.Streams) == 0 {
+		_, err := c.Writer.WriteString("No lectures found!")
```
joschahenningsen: Doesn't that break the ical? I suggest not showing the download button in this case.

web/template/course-overview.gohtml
```diff
@@ -101,11 +101,20 @@
...
-                        <span class="text-sm font-semibold uppercase dark:text-white"
-                              x-text="showPlanned ? '&#x25B2; hide' : '&#x25BC; show'">&#x25B2; hide</span>
-                    </button>
+                    <div class="my-auto">
+                        <button class="hover:bg-gray-200 dark:hover:bg-gray-600 rounded px-2 my-auto"
+                                @click="showPlanned = !showPlanned;">
+                            <span class="text-sm font-semibold uppercase dark:text-white"
+                                  x-text="showPlanned ? '&#x25B2; hide' : '&#x25BC; show'">&#x25B2; hide</span>
+                        </button>
+                        <button class="hover:bg-gray-200 dark:hover:bg-gray-600 rounded px-2"
```
joschahenningsen: An `<a>` tag might make more sense, or does that break the UI?

joschahenningsen (changes requested):
Looks good! I have a few small remarks but those should be a quick fix. 

> but I think the templates serve different purposes

Yes, I think so too. The other ical is made for hardware, not people :D 

In the future it would be nice, if people could add the link to the ical file to their calendar and have it automatically update it. This would require some option to copy the url.

api/download_ics.go
ata-keskin: I had no idea :) I am changing it accordingly!

api/download_ics.go
ata-keskin: Makes sense, I am changing it.

web/template/course-overview.gohtml
ata-keskin: I assumed it would, but it doesn't actually. I've changed it.

api/download_ics.go
ata-keskin: I am already checking if the User is logged in with
```
tumLiveContext := foundContext.(tools.TUMLiveContext)
  if tumLiveContext.User == nil {
    c.AbortWithStatus(http.StatusForbidden)
    return
}
```
Should I remove this to make sure calendar software can fetch the file? Or should I remove this and do as you suggested (attach `tools.InitCourse`)

ata-keskin: I don't know how much of a security risk this would be, but I think requesting the .ics file for a lecture can be made publicly available so that calendar software can fetch the file if needed as you suggested. However if this is a problem, I can do as you suggested and use `tools.InitCourse` and get the course directly from the context. @joschahenningsen

joschahenningsen: Lecture dates are public information in tumonline anyway, let's remove the user check as well and keep the course fetching as is :) If we need some kind of authentication here in the future we'd need to build a token based system

ata-keskin: Alright, I am updating the code!

ata-keskin: @joschahenningsen Done! We should be able to merge the pull request now.

joschahenningsen (approved):
Looks good, thanks! :beers:

PR #445, opened 2022-05-10
add shared run configurations

MW360:
Add shared run configurations for GoLand.
Includes combined run configurations for building and running by one click.


joschahenningsen (approved):
Nice addition :tada:

PR #447, opened 2022-05-11
Draft: Fix sort order of VoDs within a given month

### LINKED ISSUE #448: Descending Order of VODs are not daywise descending
**Describe the bug**
When visiting the VOD-section of a course, the "desc" order of the VODs is only monthwise descending instead of daywise.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to any course overview side with active VODs
2. See VOD-section

**Expected behaviour**
A day-wise descending VOD list. This improves the user/student experience, because most students are going to watch the most recent VOD. When there are a lot of lectures within a month, this issue gets even worse.

**Screenshots**
image

 - OS: Manjaro Linux 
 - Browser Firefox
 - Version 100 🎉 

 - Device: OnePlus 6
 - OS: Android 11 (stock/Oxygen OS)
 - Browser Google Chrome
 - Version 101

just-max:
Currently, changing between the ascending and descending sort order on the VoD overview only changes the order of the months, which I think is not the expected behaviour (and I've heard from at least one other person who found it confusing). Instead, the lectures should all be sorted by date (either ascending or descending) within a month AND the months should be sorted. This PR fixes that.

Before (order of lectures is 2, 3, 1):
image

After (order 3, 2, 1):
image

Resolves #448

### DIFF
--- a/web/template/course-overview.gohtml
+++ b/web/template/course-overview.gohtml
@@ -19,7 +19,7 @@
         <!-- VoD -->
         <div {{if .IndexData.TUMLiveContext}}x-init="watchedTracker.init({{.WatchedData}})"
              x-data="watchedTracker = new global.WatchedTracker" {{end}}class="w-full md:row-span-6 md:col-span-3">
-            <div x-data="{asc: false{{if .IndexData.TUMLiveContext.User}}, filterWatched: true, watchedAll: watchedTracker.userWatchedAll(){{end}}}"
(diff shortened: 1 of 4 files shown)


just-max:
I've also just noticed that logged-out users have the opposite sort order (for example, when DESC is selected, the newest lectures are at the bottom, and the "next lectures" card has the wrong order), presumably the sorting during page rendering is non-deterministic.

joschahenningsen (approved):
Thanks!

joschahenningsen:
I'm sorry, hope i didn't merged early, looked good already :D

just-max:
Nope, all good! I was going to fix the issue from my comment in the same branch but I'll do it separately instead :)

PR #451, opened 2022-05-12
Fix: Different speeds don't work on Firefox mobile

### LINKED ISSUE #430: Different speeds don't work on Firefox mobile
When changing the speed to over 1, it plays faster for a little while, but every few seconds it loads even though the internet speed is fast enough to potentially play at >10x (150mbit). Doesn't happen in Chrome, but a fix would still be nice for the probably ~1000 (2%) people at TUM using Firefox Mobile.
Oddly enough happens more often in 1.25x and 1.5x than 2x. There are also no websites I've ever had this problem with so I assume TUM Live has to be the problem.
Tested in 99.2.0 and 101.0.0-beta.1

just-max:
Fixes #430 

Also removed (declared) dependency on `videojs-contrib-hls`, since HLS support is already built into VideoJS 7. As far as I can tell nothing is using the package, but correct me if I missed something.


web/package.json
```diff
@@ -42,7 +42,6 @@
     "moment": "^2.29.3",
     "nouislider": "^15.6.0",
     "video.js": "^7.18.1",
-    "videojs-contrib-hls": "^5.15.0",
```
joschahenningsen, line 46: can you also sync the package-lock.json?

web/template/admin/lecture-cut.gohtml
```diff
@@ -20,7 +20,7 @@
     <div class="min-w-screen flex items-center justify-center">
         <div class="w-full lg:w-5/6">
             <p class="text-gray-300">Reference video:</p>
-            <video
+            <video-js
```
joschahenningsen, line 23: I was thinking way to complicated here :smile: where did you find this fix? We should definitely do some testing as to how browsers react to this, but looks good!

joschahenningsen (changes requested):
Amazing, thanks a lot for your engagement here, hit us up if you want to do this on a regular basis!

just-max:
> where did you find this fix

the stream itself worked fine in the videojs demo, so from there transplanting things into the demo until it broke again :) the videojs [docs](https://videojs.com/guides/setup/#creating-a-player) do mention it, but of course I only saw that afterwards...

> hit us up if you want to do this on a regular basis

sure, depending on how busy the semester gets :<

joschahenningsen (approved):
Looks good! Thanks :)

PR #452, opened 2022-05-12
Add tailwind and go template plugin for prettier

alexanderstephan:
Adds
- https://github.com/tailwindlabs/prettier-plugin-tailwindcss
- https://github.com/NiklasPor/prettier-plugin-go-template

Parser e.g. complains about closed p tags and so on, but doesn't seem to be harmful.


joschahenningsen:
The current conflicts seem to be hard to resolve. Let's close this for now and redo it if we need it, what do you think @alexanderstephan?

alexanderstephan:
@joschahenningsen It wasn't too bad of a merge conflict. Nonetheless, we should finally decide whether we want this. There are a couple nasty parser errors due to non standard html, but it doesn't do anything harmful as far as I can tell.

alexanderstephan:
Having the closing bracket on a dedicated line actually seems pretty common nowadays and I think one gets used to it (despite it looks very weird at first) 😂

alexanderstephan:
Closing for now as there is real benefit and the risk of breakage is very high.

PR #456, opened 2022-05-15
added Links to the rooms in the about section

CommanderStorm:
This PR adds links to the room listed in the about us section.
Possibly sourounding this with a ul/li elements would have been nicer for code-readability, but that would have been a layout change and I wanted to keep my change as minmal as possible to increase the change of this being merged.

### DIFF
--- a/web/template/about.gohtml
+++ b/web/template/about.gohtml
@@ -24,7 +24,18 @@
     </p>
     <p><b>1. Die Veranstaltung findet in einem von uns ausgestatteten Raum statt:</b><br>
         Ihre Veranstaltung läuft in einen dieser Räume:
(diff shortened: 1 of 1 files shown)


joschahenningsen (commented):
Neat :ok_hand: I'm just noticing that the list should include MW2001, MW0001 and 00.07.014. Could you include them as well?

joschahenningsen:
> I wanted to keep my change as minmal as possible to increase the change of this being merged

No worries there :D feel free! I actually wanted to convert this page into a markdown file that can be provided as a config instead of hardcoding it sometime.

web/template/about.gohtml
```diff
@@ -24,7 +24,18 @@
     </p>
     <p><b>1. Die Veranstaltung findet in einem von uns ausgestatteten Raum statt:</b><br>
         Ihre Veranstaltung läuft in einen dieser Räume:
-        <i>MI HS 1, MI HS 2, MI HS 3, 00.08.038 (Seminarraum), 00.13.009A (Seminarraum), Interims I (101), Interims I(102)</i>.<br>
+        <ul class="ml-3 pl-3 list-disc">
```
CommanderStorm, line 27: just margin does not give me the amount of indentation, I think is needed here, this is why I use margin+padding..

joschahenningsen (approved):
Thanks! :beers:

PR #459, opened 2022-05-16
API Tests

MatthiasReumann:
This is a first attempt to increase code coverage by adding tests for API Endpoints. 

**Some considerations:**
- We could also think about including something similar to https://github.com/marketplace/actions/codecov, and adding it to the `README.md`
- If we want to test `courseimport.go` thoroughly, we need to mock the functions we use of the `campusonline` library the same way we mocked DAOs.

**Missing Endpoints**
- `Stream`
- `Courses`
- A few of `LectureHall`


MatthiasReumann:
@joschahenningsen Do you have any idea why the tests fail (`*mock_dao.MockStatisticsDao does not implement dao.StatisticsDao (missing GetStreamNumLiveViews method)`)?

Locally everything runs unproblematic. (_"Runs on my machine"_), but as far as I know `GetStreamNumLiveViews` doesn't exist. I also re-ran `go generate ./...`. 

Any ideas are welcome, thanks.

joschahenningsen (approved):
This really looks like a ton of work, thank you so much!

PR #461, opened 2022-05-16
Use PUT to update description consistently

### LINKED ISSUE #460: HTTP Method Mismatch in Lecture Description Edit Leads to 404
Route is `POST`:
https://github.com/joschahenningsen/TUM-Live/blob/b260f68171736a595dbc3d4fec57c0abe1340e57/api/courses.go#L48

Client uses `PUT`:
https://github.com/joschahenningsen/TUM-Live/blob/b260f68171736a595dbc3d4fec57c0abe1340e57/web/ts/edit-course.ts#L187

This fails with a 404. Using the same request in cURL but with the method changed to `POST` works fine. This should probably be a `PUT` in both cases, since it is idempotent.

This was broken in #394 by [the look of it](https://github.com/joschahenningsen/TUM-Live/blame/b260f68171736a595dbc3d4fec57c0abe1340e57/api/courses.go#L48) @MatthiasReumann.

Maybe other uses of the API should be reviewed for consistency?

joschahenningsen:
fixes #460


just-max:
thanks :)

PR #462, opened 2022-05-16
User settings page

joschahenningsen:
closes #402 
closes #431 
closes #358 
closes #360 
closes #465

PR #464, opened 2022-05-17
Updated playbackRates

CommanderStorm:
I widedned the upper range for playbackRates and removed one in the lower rate

The reasoning for this Stepsize: 
  - between 1..2 0.25 is a sane stepsize, as I would assume that most users watch in this range
  - 0.5..1 and 2..3: This is a range, I would assume to be uncommon, so we dont need to offer such a granular stepsize.

### DIFF
--- a/web/ts/TUMLiveVjs.ts
+++ b/web/ts/TUMLiveVjs.ts
@@ -29,7 +29,7 @@ export const initPlayer = function (
         {
             liveui: true,
             fluid: fluid,
-            playbackRates: [0.5, 0.75, 1, 1.25, 1.5, 1.75, 2],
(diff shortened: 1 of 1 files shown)


CommanderStorm:
However, this may be a good setting for #462. What do you Think?

joschahenningsen:
0.75 is the perfect speed for lectures that are just a tiny bit too fast imo, but adding 2.5 and 3 for :sparkles: those classes :sparkles: seems like a good addition! 

Also I can include customizing these in #462

CommanderStorm:
Yea, costomising this in #462 is probably the best cause of actioin.
I'll close this PR. Should I create an Issue?

joschahenningsen:
> Should I create an Issue? 

That'd be great :blush:

PR #467, opened 2022-05-17
Improve filtering watched videos

alexanderstephan:
**Description**:

- Makes setting the VoD order and "filter watched" persistent
- Adds a counter behind "filter watched", so there is a visual indicator that there is something that is/can be filtered
- Only show the filter option, when watched lectures exist

If you have any other ideas for this feature, please tell me!

**Demo**:

Peek 2022-05-17 13-59


joschahenningsen (approved):
:heart:

PR #469, opened 2022-05-18
Presentation Only/Camera Only Streams

### LINKED ISSUE #226: Presentation only/camera only streams
Some profs only use the blackboard or don't need a camera image, they only need one of the three streams. 

On the stream administration page, select **camera only** or **presentation only**. The selected stream should be streamed as default (`playlist_url`)

ata-keskin:
Renamed "Camera Presets" section to "Source Settings".
Added a dropdown menu to set source mode (choose between: Presentation & Camera, Presentation Only, Camera Only). This affects all past and future streams and makes sure only the specified source(s) are being used
Made sure all lecture halls settings are visible by default if there is a stream using it.
Modified how camera presets are chosen. The UI is more straightforward now. The chosen preset is marked at the beginning. The user can select the desired preset and click on save changes to save it. The previous version looked buggy in my opinion.
Modified worker code to only stream the correct sources depending on the setting chosen per course per lecture hall.
Modified /watch api to always serve PRES url or CAM url if source mode is set to presentation only (resp. camera only)
Modified client-side code to show the correct buttons on the course overview page.

Known issues:
The source mode can't be changed mid-stream. The stream has to be restarted for changes to take effect, since the workers have to start serving the sources again. I don't think we need to work on making the source made changeable mid-stream, since the users are meant to set the source mode on a per course basis.

Right now source mode selection is per (course, lecture hall) as stated above. An extension of this would be to change it to per (stream, lecture hall) which would provide more freedom to the user. However I don't know how necessary this is.


ata-keskin:
@joschahenningsen I changed how the camera presets are chosen quite a bit. I hope I only enhanced the functionality and didn't actually break anything. The UI is much more intuitive now to say the least. I would appreciate it if you could take a look at how the Camera Presets section is looking now. I hope the idea is that the user can only choose one camera preset per lecture hall per course (I have pretty much excluded any option to do that by modfying the UI). I'm awaiting your feedback!

ata-keskin:
I still need to modify watch.gohtml and then hopefully it will be done :)

ata-keskin:
I think the implementation is complete, but I need to test it further on the test server. Feedback is appreciated!

worker/pb/api.pb.go
```diff
@@ -255,7 +255,7 @@ type StreamRequest struct {
 	unknownFields protoimpl.UnknownFields
 
 	WorkerId      string                 `protobuf:"bytes,1,opt,name=WorkerId,proto3" json:"WorkerId,omitempty"`
-	SourceType    string                 `protobuf:"bytes,2,opt,name=SourceType,proto3" json:"SourceType,omitempty"`
+	SourceType    string                 `protobuf:"bytes,2,opt,name=SourceMode,proto3" json:"SourceMode,omitempty"`
```
joschahenningsen: This is a generated file and should not be edited. If you want to change the api, please edit `worker/api/api.proto` and rebuild the grpc stub (see makefile)

Why does this need to be changed though?

model/course.go
```diff
@@ -68,6 +69,54 @@ func (c *Course) SetCameraPresetPreference(pref []CameraPresetPreference) {
 	c.CameraPresetPreferences = string(pBytes)
 }
 
+type SourcePreference struct {
+	LectureHallID uint `json:"lecture_hall_id"`
+	SourceMode    int  `json:"source_mode"`
```
joschahenningsen: Can you create a type for the source mode like here: https://github.com/joschahenningsen/TUM-Live/blob/15f794a088386c5fbf5f1246e14f05cb9edaa0f2/model/stream.go#L16

(this is usually how we simulate mocks in go)

model/course.go
```diff
@@ -68,6 +69,54 @@ func (c *Course) SetCameraPresetPreference(pref []CameraPresetPreference) {
...
+	mode := c.GetSourceModeForLectureHall(lectureHallID)
+	switch sourceType {
+	case "PRES":
+		return mode != 2
+	case "CAM":
+		return mode != 1
+	case "COMB":
+		return mode != 1 && mode != 2
+	}
+	return false
```
joschahenningsen: In case something unexpected happens, this should default to true

model/course.go
```diff
@@ -68,6 +69,54 @@ func (c *Course) SetCameraPresetPreference(pref []CameraPresetPreference) {
...
+}
+
+// SetSourcePreference updates the source preferences
+func (c *Course) SetSourcePreference(pref []SourcePreference) {
+	pBytes, err := json.Marshal(pref)
+	if err != nil {
+		log.Println(err)
```
joschahenningsen:
```suggestion
		log.WithError(err).Error("could not marshal source preference")
		return
```

api/worker_grpc.go
```diff
@@ -563,73 +563,83 @@ func NotifyWorkers(daoWrapper dao.DaoWrapper) func() {
...
-					continue
-				}
-				var slot model.StreamName
-				if sourceType == "COMB" { //try to find a transcoding slot for comb view:
-					slot, err = daoWrapper.IngestServerDao.GetTranscodedStreamSlot(server.ID)
-				}
-				if sourceType != "COMB" || err != nil {
-					slot, err = daoWrapper.IngestServerDao.GetStreamSlot(server.ID)
+
+			createStreamRequest := func(sourceType string, source string) {
```
joschahenningsen: This does not need to be an inline function. Can you extract it to package level?

web/template/admin/admin_tabs/edit-course.gohtml
```diff
@@ -25,9 +25,9 @@
                 </div>
 
                 <div class="bg-white dark:bg-secondary rounded shadow p-4 my-5">
-                    <h2 class="text-2 text-2xl font-semibold">Camera Presets</h2>
+                    <h2 class="text-2 text-2xl font-semibold">Source Settings</h2>
```
joschahenningsen: This could be more intuitive for lecturers:

```suggestion
                    <h2 class="text-2 text-2xl font-semibold">Lecture Hall Settings</h2>
```

web/template/partial/course/manage/source-settings.gohtml
```diff
@@ -0,0 +1,56 @@
...
+    <div x-cloak class="mt-2" x-data="{halls: [], changed: false}"
+         x-init="fetch(`/api/lecture-halls-by-id?id=${document.getElementById('courseID').value}`).then(r=>r.json()).then(d=>halls=d)">
+        <template x-if="halls !== null && halls.length > 0">
+            <template x-for="hall in halls">
+                <div x-data="{selected_preset: undefined, selected_preset_name: ''}" class="ml-2"
+                        x-init="selected_preset = hall.presets.find(e => e.PresetID==hall.selected_preset_id); selected_preset_name = selected_preset in window ? '' : selected_preset.Name"
+                        @reinit.window="selected_preset = hall.presets.find(e => e.PresetID==hall.selected_preset_id); selected_preset_name = selected_preset in window ? '' : selected_preset.Name">
+                    <div class="flex justify-between">
+                        <p class="font-semibold px-1 my-auto" x-text="hall.lecture_hall_name"></p>
+                        <select class="text-black dark:bg-secondary bg-gray-100 p-2"
```
joschahenningsen: This needs some styling for dark mode.

joschahenningsen (changes requested):
Looks good already, just a few remarks :)

worker/pb/api.pb.go
ata-keskin: I am not sure why I changed this, I don't exactly remember changing it. I will revert it and test to make sure everything still works.

model/course.go
ata-keskin: Done.

model/course.go
ata-keskin: Done

api/worker_grpc.go
ata-keskin: Done.

joschahenningsen (approved):
As discussed in the meeting, this looks great but should be merged in the semester break. Thanks for the effort! :)

web/watch.go
```diff
@@ -57,6 +58,23 @@ func (r mainRoutes) WatchPage(c *gin.Context) {
 			}
 		}
 	}
+
+	if tumLiveContext.Stream.LectureHallID != 0 {
+		switch data.IndexData.TUMLiveContext.Course.GetSourceModeForLectureHall(data.IndexData.TUMLiveContext.Stream.LectureHallID) {
+		// I'm not sure if this is the best solution, but it works.
```
joschahenningsen: looks good, why not :)
```suggestion
```

joschahenningsen (approved):

PR #470, opened 2022-05-19
Fix height of ICS export button

just-max:
Before:
image

After:
image

Peace at last.


alexanderstephan (approved):
Nice catch, thanks for the fix. :+1:

PR #471, opened 2022-05-19
Fix alert modal scope

alexanderstephan:
The code for the modal was not present for non-admins due to the modal being in the wrong scope. I also used the opportunity to move the modal to its own file as it has quite a bit code.


joschahenningsen (approved):
Thanks! :)

PR #473, opened 2022-05-19
Highlight current video section on watch page

MatthiasReumann:
**PR Changes:**
Add class `border-blue-500/50` to currently played section


web/ts/TUMLiveVjs.ts
```diff
@@ -343,10 +343,62 @@ export class StartInOverlay extends Component {
...
+                this.currentHighlightIndex = 0;
+            });
+    }
+}
+
+function attachCurrentTimeEvent(videoSection: VideoSections) {
+    player.ready(() => {
+        player.on("timeupdate", function () {
+            const currentTime = this.currentTime();
+            videoSection.currentHighlightIndex = videoSection.list.findIndex((section, i, list) => {
```
alexanderstephan: Right now the player needs some time to become ready and load the progress (~1s for me), which causes the video section highlighting to jump when the progress is beyond the first section.
It would be cool if we can directly set the `currentHighlightIndex` according to the users progress, but that might be tricky do since we would probably need the player for the duration.

Maybe we can init `currentHighlightIndex` with something like `-1` (or make it nullable) and don't show the highlighting in this case in the beginning. When the user presses play we could then set `currentHighlightIndex` to a valid index and then display the highlighting.

alexanderstephan (changes requested):
Nice enhancement! :) I just noticed an edge case that we might want to address. Otherwise looks really code and design wise. :+1:

web/ts/TUMLiveVjs.ts
MatthiasReumann: Initialising `currentHighlightIndex` with `-1` works quite well. The way it works now is that it highlights the correct section once the video is loaded (and not the first).

web/ts/TUMLiveVjs.ts
```diff
@@ -348,10 +348,66 @@ export class StartInOverlay extends Component {
...
+                console.log(err);
+                this.list = [];
+                this.currentHighlightIndex = 0;
+            });
+    }
+}
+
+function attachCurrentTimeEvent(videoSection: VideoSections) {
+    player.ready(() => {
+        player.on("timeupdate", function () {
```
alexanderstephan: Maybe we also want to also register on `seeked` here, see https://docs.videojs.com/player#event:seeked. We should also use a loop for this since this might heavy on resource (e.g. it trigger every full second). According to docs it can be fired up to every 15ms (https://docs.videojs.com/player#event:timeupdate). I have done something similar for the progress reports above, maybe we can reuse that.

web/ts/TUMLiveVjs.ts
alexanderstephan: Loading itself looks good now :+1: 

I noticed than in like 25% of the cases there is an error `VIDEOJS: ERROR: DOMException: Index or size is negative or greater than the allowed amount` now, which is pretty weird. Attaching to the events mentioned above seems to solve this issue.

Also, clicking on sections has a weird behavior such that the wrong section is highlighted when clicked. The reason seems to be that the time that the player jumps to does not trigger the new section yet. This is problematic when the video is paused. If the video is running it can look like highlighting is lagging behind.

Example for drag:
Peek 2022-05-25 22-27

Example for section not triggered:
Peek 2022-05-25 22-25

alexanderstephan (changes requested):

web/ts/TUMLiveVjs.ts
MatthiasReumann: `seeked` works like a charm thanks! ❤️ 

I've furthermore removed the `player.on("timeupdate"...)` and added, as you did in #251, a custom `500ms` timeout.

alexanderstephan (approved):
Thanks for the updates, switching sections feels really smooth now. :100:

PR #475, opened 2022-05-22
OCR Keyword extraction

MatthiasReumann:
# Changes
- Add (simple) keyword extractor (`ocr.NewKeywordExtractor(...)`)
- Add fromWorker grpc `NewKeywords` function 
- Add Table `Keywords (streamId, text, language, valid)`
- Extract keywords after streamEnd, vodUpload (after silence detection)

# Follow-Ups
- Integrate Keywords in Search Results
- Filter more keywords from Extractor results

# Privacy statement 
```
Um den Funktionsumfang von TUM-Live zu erweitern, werden Aufzeichnungen automatisiert und maschinell verarbeitet.
```


api/keyword.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/joschahenningsen/TUM-Live/model"
+	"github.com/joschahenningsen/TUM-Live/tools"
+	log "github.com/sirupsen/logrus"
+	"net/http"
+	"strconv"
+)
+
+func configGinKeywordRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
+	routes := keywordRoutes{daoWrapper}
+	router.Use(tools.Admin) // Todo: Can workers post as admin?
```
joschahenningsen: Workers don't use the http api, they communicate using grpc. See:

https://github.com/joschahenningsen/TUM-Live/blob/dev/worker/api/api.proto
https://github.com/joschahenningsen/TUM-Live/blob/fcc1301ac22ad75feaf78306babd4bfc924106a8/worker/Makefile
https://github.com/joschahenningsen/TUM-Live/blob/dev/api/worker_grpc.go

MatthiasReumann:
## New Problem

The current approach (store every keyword in an entry) is super slow. The full-text-search-query takes a lot of time. Will need to find another way to store keywords and potentially store less of them.

joschahenningsen:
What does super slow mean in terms of milliseconds?

MatthiasReumann:
> What does super slow mean in terms of milliseconds?

```sql
SELECT streams.*,
       (match(streams.name) against('correctness*' in boolean mode) +
        match(streams.description) against('correctness*' in boolean mode) +
        match(c.message) against('correctness*' in boolean mode) +
        match(vs.description) against('correctness*' in boolean mode) +
        match(kw.text) against('correctness')) as 'relevance'
FROM `streams`
         left join video_sections vs on vs.stream_id = streams.id
         left join chats c on c.stream_id = streams.id
         left join tumlive.keywords kw
                   on kw.stream_id = streams.id
WHERE ((match(streams.name) against('correctness*' in boolean mode) OR
        match(streams.description) against('correctness*' in boolean mode) OR
        match(c.message) against('correctness*' in boolean mode) OR
        match(vs.description) against(53 in boolean mode) OR
        match(kw.text) against('correctness')) AND streams.course_id = 53 AND streams.recording = 1)
  AND `streams`.`deleted_at` IS NULL
GROUP BY `streams`.`id`
ORDER BY relevance
```
This query takes 29s(!!!) on my machine. However, I might be able to write a more sophisticated, better performing query. 

## Update 
```sql
select t.id, t.relevance
from (SELECT distinct streams.*, match(keywords.text) against('correctness*') 'relevance'
      FROM streams
               JOIN tumlive.keywords on streams.id = keywords.stream_id
      WHERE match(keywords.text) against('correctness*')
      UNION
      SELECT distinct streams.*, match(streams.name) against('correctness*') 'relevance'
      FROM streams
      WHERE match(streams.name) against('correctness*')
      UNION
      SELECT distinct streams.*, match(streams.description) against('correctness*') 'relevance'
      FROM streams
      WHERE match(streams.description) against('correctness*')
      UNION
      SELECT distinct streams.*, match(c.message) against('correctness*') 'relevance'
      FROM streams
               JOIN chats c on streams.id = c.stream_id
      WHERE match(c.message) against('correctness*')
      UNION
      SELECT distinct streams.*, match(vs.description) against('correctness*') 'relevance'
      FROM streams
               JOIN video_sections vs on streams.id = vs.stream_id
      WHERE match(vs.description) against('correctness*')) t
where t.course_id = 53
  and t.recording = 1
  and t.deleted_at IS NULL
group by t.id
order by relevance;
```

This one takes about 38ms. I think the db can parallelize each individual query.

worker/worker/request_handlers.go
```diff
@@ -76,20 +81,20 @@ func HandleSelfStream(request *pb.SelfStreamResponse, slug string) *StreamContex
 	return streamCtx
 }
 
-func HandleSelfStreamRecordEnd(ctx *StreamContext) {
-	S.startTranscoding(ctx.getStreamName())
-	err := transcode(ctx)
+func HandleSelfStreamRecordEnd(streamCtx *StreamContext) {
```
joschahenningsen: renaming these seems out of scope for this pr

worker/ocr/ocr.go
```diff
@@ -0,0 +1,91 @@
...
+		unfiltered = append(unfiltered, words...)
+	}
+	e.client.Close()
+	return unfiltered, nil
+}
+
+func removeStopwords(words []string) []string {
+	res := make([]string, 0)
+	for _, word := range words {
+		clean := stopwords.CleanString(word, "en", false)
```
joschahenningsen, line 62: This hardcodes english, would it make sense to just remove german and english stopwords just to be safe? Or would this mess up the output?

worker/ocr/ocr.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		unfiltered = append(unfiltered, words...)
+	}
+	e.client.Close()
+	return unfiltered, nil
+}
+
+func removeStopwords(words []string) []string {
+	res := make([]string, 0)
```
joschahenningsen: There is a lot of reallocation going on here, I'd initialize these slices with the capacity of the existing words.

```suggestion
	res := make([]string, 0, len(words))
```

worker/ocr/ocr.go
```diff
@@ -0,0 +1,91 @@
...
+
+		if len(clean) > 0 {
+			res = append(res, strings.TrimSpace(clean))
+		}
+	}
+	return res
+}
+
+func removeShortWords(words []string) []string {
+	res := make([]string, 0)
```
joschahenningsen:
```suggestion
	res := make([]string, 0, len(words))
```

worker/ocr/ocr.go
```diff
@@ -0,0 +1,91 @@
...
+		if len(word) > 2 {
+			res = append(res, word)
+		}
+	}
+	return res
+}
+
+func removeDuplicates(words []string) []string {
+	check := map[string]bool{}
+	res := make([]string, 0)
```
joschahenningsen, line 84:
```suggestion
	res := make([]string, 0, len(words))
```

worker/worker/request_handlers.go
```diff
@@ -252,7 +271,7 @@ func HandleUploadRestReq(uploadKey string, localFile string) {
 		log.WithError(err).Error("Error getting stream info for upload")
 		return
 	}
-	c := StreamContext{
+	streamCtx := StreamContext{
```
joschahenningsen: Here as well

worker/worker/request_handlers.go
```diff
@@ -489,3 +515,87 @@ func (s StreamContext) getStreamNameVoD() string {
...
+	dir, err := ioutil.TempDir("", "keyword-extraction")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(dir)
+	defer os.Remove("out.jpeg") // Remove thumbgen's out.jpeg
+
+	// generate images to process
+	g, err := thumbgen.New(ctx.getTranscodingFileName(), 768, 128,
+		"out.jpeg", thumbgen.WithJpegCompression(70), thumbgen.WithStoreSingleFrames(dir))
```
joschahenningsen: Using the same filename for every stream will likely cause trouble with parallel processing of multiple lectures

worker/worker/request_handlers.go
```diff
@@ -489,3 +515,87 @@ func (s StreamContext) getStreamNameVoD() string {
...
+	if err != nil {
+		return err
+	}
+	err = g.Generate()
```
joschahenningsen, line 541: This error is unhandled

worker/worker/request_handlers.go
```diff
@@ -489,3 +515,87 @@ func (s StreamContext) getStreamNameVoD() string {
...
+	for i, _ := range fileInfoArray {
+		files[i] = path.Join(dir, fileInfoArray[i].Name())
+	}
+
+	// extract keywords
+	engExtractor := ocr.NewOcrExtractor(files, []string{"eng"})
+	engKeywords, err := engExtractor.Extract()
+	if err != nil {
+		return err
+	}
```
joschahenningsen, line 562: You create keywords for german and english, how messy are the results of the wrong language? (e.g. the english results of a german lecture and vice versa?)

api/worker_grpc.go
```diff
@@ -244,6 +244,27 @@ func (s server) NotifyStreamFinished(ctx context.Context, request *pb.StreamFini
...
+func (s server) NewKeywords(ctx context.Context, request *pb.NewKeywordsRequest) (*pb.Status, error) {
+	if _, err := s.DaoWrapper.WorkerDao.GetWorkerByID(ctx, request.GetWorkerID()); err != nil {
+		return nil, errors.New("authentication failed: invalid worker id")
+	} else {
+		for _, keyword := range request.Keywords {
+			err := s.DaoWrapper.KeywordDao.NewKeyword(&model.Keyword{
+				StreamID: uint(request.StreamID),
+				Text:     keyword,
+				Language: request.Language,
+			})
```
joschahenningsen: Can you batch insert them instead of doing one query per keyword?

web/template/video_only.gohtml
```diff
@@ -75,7 +75,7 @@
 </video-js>
 </body>
 <script>
-    watch.initPlayer(false, true, true, {{.IndexData.TUMLiveContext.User.GetPlaybackSpeeds.GetEnabled}}, {{$stream.GetThumbIdForSource .Version}}, {{$stream.ThumbInterval}}, {{$stream.Model.ID}}, {{$course.Name}}, {{$stream.GetName}}, {{$course.GetStreamUrl $stream}}, {{$course.GetUrl}}, {{$stream.GetStartInSeconds}});
+    watch.initPlayer(false, true, true, {{.IndexData.TUMLiveContext.User.GetPlaybackSpeeds.GetEnabled}}, {{$stream.GetThumbnailIdForType .Version}}, {{$stream.ThumbInterval}}, {{$stream.Model.ID}}, {{$course.Name}}, {{$stream.GetName}}, {{$course.GetStreamUrl $stream}}, {{$course.GetUrl}}, {{$stream.GetStartInSeconds}});
```
joschahenningsen: I think a few unrelated changes slipped in this PR, can you merge `dev` please?

joschahenningsen (changes requested):
Looks great, just a few more nits :)

worker/worker/request_handlers.go
MatthiasReumann: Changed to `fmt.Sprintf("%s.jpeg", uuid.NewUUID())`

worker/worker/request_handlers.go
MatthiasReumann: That's actually quite interesting. The intersection of german and the english results are pretty good. 
For example the following query returns 297 of 1600 extracted keywords.

```sql
select text
from tumlive.keywords
where text in (select text
               from tumlive.keywords
               where language = 'deu')
  and language = 'eng';
```

So there are some differences, but theoretically we could only extract the keywords in one language. 

There are three ways to tackle that problem: 
1. Store all the keywords we can get and then later figure out a way to remove individual unproductive words. (That's why I added the `valid` attribute in the first place)
2. Only store words which have been extracted in both languages. However, potentially, missing productive words.
3. Only extract keywords in one language with the same problems as 2.

Which one would you prefer?

worker/ocr/ocr.go
MatthiasReumann: Works just fine. From 1600 keywords down to 1586.

joschahenningsen (approved):
Looks great! 
Feel free to merge when the conflicts are resolved :)

PR #477, opened 2022-05-23
Fix referrer bug

### LINKED ISSUE #454: Login page redisplayed after successful login following an unsuccessful one.
**Describe the bug**
If the user first tries to login using an internal account and fails, and then succeeds, we are redirected to the login page instead of the home page.

**To Reproduce**

0. Be logged out of the system
1. Go to 'Login'
2. Click on 'use an internal account'
3. Enter wrong login information.
4. Get the warning 'Couldn't log in. Please double check your credentials.'
5. Enter correct login information.
6. Get the login page again, which is not the expected behaviour.

**Expected behavior**
After step 5, the user should be redirected to the homepage. The user should not be able to visit "./login" if they are logged in. Maybe it would make sense to redirect them to the start page if they try this?

 - OS: [Ubuntu 20.04
 - Browser: Vivaldi

MatthiasReumann:
closes #454

### DIFF
--- a/web/ts/global.ts
+++ b/web/ts/global.ts
@@ -192,6 +192,11 @@ export function timer(expiry: string, leadingZero: boolean) {
     };
 }
 
+// getLoginReferrer returns "/" if document.referrer === "http://<hostname>:<port>/login" and document.referrer if not
+export function getLoginReferrer(): string {
+    return document.referrer === window.location.origin + "/login" ? window.location.origin + "/" : document.referrer;
+}
+
 window.onload = function () {
     initHiddenCourses();
 };
--- a/web/ts/global.ts
+++ b/web/ts/global.ts
@@ -192,9 +192,21 @@ export function timer(expiry: string, leadingZero: boolean) {
     };
 }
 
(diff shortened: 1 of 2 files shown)


web/ts/global.ts
```diff
@@ -192,6 +192,11 @@ export function timer(expiry: string, leadingZero: boolean) {
     };
 }
 
+// getLoginReferrer returns "/" if document.referrer === "http://<hostname>:<port>/login" and document.referrer if not
```
joschahenningsen: Nit :D 
```suggestion
// getLoginReferrer returns "/" if document.referrer === "http[s]://<hostname>:<port>/login" and document.referrer if not
```

joschahenningsen (changes requested):
While you're at it: I feel like this should definitely also check if the referrer is anywhere outside window.location.origin and redirect to `/` then. Currently a redirect from moodle to https://live.rbg.tum.de/login would redirect back to moodle on success.

joschahenningsen (approved):
Thanks! ❤️

PR #478, opened 2022-05-24
Thumbnails on hover

### LINKED ISSUE #472: Display thumbnails when hovering over the timeline

YouTube and other platforms show a small thumbnail of the video when hovering over the progress bar.

Peek 2022-05-19 10-53

This is useful when searching through lectures or when wanting to skip certain parts of lecture.

Sprites should be generated for every VoD in 5s intervals using https://github.com/flavioribeiro/video-thumbnail-generator.
Then we can fetch the sprites and integrate them in the timeline via https://github.com/phloxic/videojs-sprite-thumbnails

Generation of sprites uses a considerable amount of memory, therefore this feature is fairly expensive in term of resources. We could just leave it as it is, but I think it's worth it.

alexanderstephan:
Fixes #472

- Also removes usage of a deprecated file path for downloads
- Fixes `TestGetTranscodingFileName` by using the correct path

Deployed on the test server, streaming still works as expected. To fully test this it would be nice to allow VoD uploads after the stream ends.

Existing videos will be migrated in https://github.com/joschahenningsen/TUM-Live/issues/536


worker/worker/status.go
```diff
@@ -105,6 +114,19 @@ func (s *Status) endTranscoding(name string) {
 	statusLock.Unlock()
 }
 
+func (s *Status) endThumbnailGeneration(streamContext *StreamContext) {
+	defer s.SendHeartbeat()
+	statusLock.Lock()
+	s.workload -= costThumbnailGeneration
+	for i := range s.Jobs {
+		if s.Jobs[i] == fmt.Sprintf("transcoding %s", streamContext.getTranscodingFileName()) {
```
MatthiasReumann: I am not quite sure but shouldn't that be `fmt.Sprintf("generating thumbnail for %s", streamCtx.getTranscodingFileName())`?

MatthiasReumann, line 122: I am not quite sure but should that be `fmt.Sprintf("generating thumbnail for %s", streamCtx.getTranscodingFileName())`?

alexanderstephan: Thanks, seems like I forgot to adjust it

worker/worker/status.go
```diff
@@ -105,6 +114,19 @@ func (s *Status) endTranscoding(name string) {
 	statusLock.Unlock()
 }
 
+func (s *Status) endThumbnailGeneration(streamContext *StreamContext) {
+	defer s.SendHeartbeat()
+	statusLock.Lock()
+	s.workload -= costThumbnailGeneration
+	for i := range s.Jobs {
+		if s.Jobs[i] == fmt.Sprintf("generating thumbnail for  %s", streamContext.getTranscodingFileName()) {
```
joschahenningsen:
```suggestion
		if s.Jobs[i] == fmt.Sprintf("generating thumbnail for %s", streamContext.getTranscodingFileName()) {
```

joschahenningsen:
Code looks great, there still seems to be an issue with thumbgen I'm investigating.

joschahenningsen:
Also, I reconfigured some stuff on the testserver, vods should now work.

model/file.go
```diff
@@ -6,9 +6,14 @@ import (
...
 )
 
+type FileType uint
+
 const (
 	FILETYPE_DOWNLOAD = iota + 1
 	FILETYPE_ATTACHMENT
+	FILETYPE_THUMB_COMB
+	FILETYPE_THUMB_CAM
+	FILETYPE_THUMB_PRES
```
joschahenningsen, line 17: Can you make sure these are not shown in the stream download dropdown?

web/ts/TUMLiveVjs.ts
```diff
@@ -49,6 +52,20 @@ export const initPlayer = function (
...
+            .spriteThumbnails({
+                interval: spriteInterval,
+                url: "/api/downloads/" + spriteID,
+                width: 160,
+                height: 90,
+            })
+            .log.level("debug");
+    } else {
+        console.log("No valid thumbnail sprite ID found.");
+    }
```
joschahenningsen: We probably don't need the debugging in prod, right?

worker/Dockerfile
```diff
@@ -8,10 +8,10 @@ RUN GO111MODULE=on go mod download
 ARG version=dev
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -ldflags "-w -extldflags '-static' -X main.VersionTag=${version}" -o /worker cmd/worker/worker.go
 
-FROM alpine:3.16
+FROM alpine:3.15
 
 RUN apk add --no-cache \
-  ffmpeg=5.0.1-r1 \
+  ffmpeg=4.4.1-r2 \
```
joschahenningsen: This can be rolled back

joschahenningsen (changes requested):
I tried getting thumbnails working here: https://test.live.mm.rbg.tum.de/w/airplay/353/PRES 
The database looks fine to me, but the thumbnails don't show, can you verify this?

model/stream.go
```diff
@@ -24,46 +24,67 @@ const (
...
+	Chats             []Chat
+	Stats             []Stat
+	Units             []StreamUnit
+	VodViews          uint `gorm:"default:0"` // todo: remove me before next semester
+	StartOffset       uint `gorm:"default:null"`
+	EndOffset         uint `gorm:"default:null"`
+	LectureHallID     uint `gorm:"default:null"`
+	Silences          []Silence
+	Files             []File `gorm:"foreignKey:StreamID"`
+	ThumbnailInterval uint32 `gorm:"default:null"`
```
joschahenningsen: Would the plugin accept floats? Full seconds might cause some shifting

alexanderstephan: Hm, it is a `number`, see https://github.com/phloxic/videojs-sprite-thumbnails#configuration. I am not sure what happens in the calculation, but it can't hurt to have a float here. :+1:

worker/Dockerfile
alexanderstephan: Ah, nice. When I tried this yesterday the `ffmpeg` command crashed, so I thought this was the solution. It worked afterwards but might be a coincidence.

web/ts/TUMLiveVjs.ts
alexanderstephan: I'll remove it after testing on the test server is done :+1:

model/file.go
alexanderstephan: Ah, good point. I guess we only want VoD downloads anyways in the download modal as attachments are shown separately.

worker/Dockerfile
alexanderstephan: This seems to break it again, see

alexanderstephan:
I adjust everything to use intervals, as it seems more robust regarding potential rounding errors

joschahenningsen (approved):

PR #479, opened 2022-05-24
Fix video-js-hls-quality-selector deprecation warning

alexanderstephan:
Fixes the warning `player.tech().hls is deprecated` that is caused by the [video-js-hl-quality-selector plugin](https://github.com/chrisboustead/videojs-hls-quality-selector/issues/78) by using a PR that fixes is by using `vhs` instead of `hls`.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -4873,7 +4873,7 @@
     },
     "node_modules/videojs-hls-quality-selector": {
       "version": "1.1.4",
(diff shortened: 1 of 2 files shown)


alexanderstephan:
We should probably keep an eye on the project so see whether the maintainer becomes active again. :sweat_smile:

joschahenningsen (approved):
Thanks!

PR #486, opened 2022-05-29
Fix: Order of VoDs Is Reversed When Not Logged-In

### LINKED ISSUE #485: Order of VoDs Is Reversed When Not Logged-In
As a follow-up to #447 and in particular https://github.com/joschahenningsen/TUM-Live/pull/447#issuecomment-1124331674.

When no user is logged-in, the course overview has the VoD list in the reversed order (the "DESC" order puts the newest lectures at the bottom instead of the top).

This is because adding the watch state reverses the sort sort order w.r.t. the initial query:
https://github.com/joschahenningsen/TUM-Live/blob/c0869dd2c6d2fbfaae021699455232473464a0a5/dao/streams.go#L277
https://github.com/joschahenningsen/TUM-Live/blob/c0869dd2c6d2fbfaae021699455232473464a0a5/dao/courses.go#L222

just-max:
Fixes #485

### DIFF
--- a/web/template/vod_course_list.gohtml
+++ b/web/template/vod_course_list.gohtml
@@ -96,7 +96,7 @@
             </li>
         {{end}}
     {{end}}
-    {{if and .User $course.HasRecordings}}
+    {{if and $user $course.HasRecordings}}
         </div>
         <p x-show="watchedAll && filterWatched"
            class="font-semibold m-auto dark:text-white border-b-2 border-black dark:border-white">All VoDs watched</p>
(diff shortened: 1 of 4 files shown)


web/template/vod_course_list.gohtml
```diff
@@ -96,7 +96,7 @@
             </li>
         {{end}}
     {{end}}
-    {{if and .User $course.HasRecordings}}
+    {{if and $user $course.HasRecordings}}
```
alexanderstephan, line 99: Good catch, can you also remove the corresponding `User` field from `CoursePageData`? I don't think it's used anywhere else.

alexanderstephan (changes requested):
Nice fix, just some minor note that could be done before merge.

web/template/vod_course_list.gohtml
just-max: Removed.

alexanderstephan (approved):
Thanks! 👍

PR #487, opened 2022-05-30
Fix safari icons and theme selector dropdown

MatthiasReumann:
Before:
image
image

After:
image
image

**Tested with**
Version 15.5 (17613.2.7.1.8)
MacOS Monterey V12.4

### DIFF
--- a/web/template/partial/notifications.gohtml
+++ b/web/template/partial/notifications.gohtml
@@ -2,7 +2,7 @@
     {{- /*gotype: github.com/joschahenningsen/TUM-Live/model.User*/ -}}
     <div class="relative" x-data="{ n: new global.Notifications(), show: false}" x-init="n.fetchNotifications();"
          @keyup.escape="show=false; n.writeToStorage(true)">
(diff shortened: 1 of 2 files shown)


joschahenningsen:
Huh, weird. Does the dropdown on the username look like this too?

MatthiasReumann:
> Huh, weird. Does the dropdown on the username look like this too?

nope. I think the problem was the `role="..."` attribute on the `div`. Safari somehow misinterpreted the `div` and depicted it as default button.

just-max:
the role=button should just be an accessibility thing, along with the tabindex

maybe -webkit-appearance: none can fix this? https://stackoverflow.com/a/22878052 don't have any device with safari to test on

MatthiasReumann:
> the role=button should just be an accessibility thing, along with the tabindex
> 
> maybe -webkit-appearance: none can fix this? https://stackoverflow.com/a/22878052 don't have any device with safari to test on

The `role=button` wasn't the problem, but rather that it was an attribute on a `<div>` and not a `<button>`

joschahenningsen (approved):

just-max:
Ah, like I said a `button` is not allowed to contain a `div` though. Also, on a `button` both the role and the tabindex are redundant.

PR #476, opened 2022-05-23
Add Hotkeys for Video Player

just-max:
Based more or less on [YouTube's hotkeys](https://support.google.com/youtube/answer/7631406?hl=en).

Calling `handleClick` on `Component`s is based off of [this](https://github.com/videojs/video.js/blob/3ec2ac7f99704434d2fdc61642247ef0aa4e77b3/src/js/player.js#L3222) VideoJS code, but that approach doesn't work for all actions.

Draft PR, feedback is appreciated! I've marked some things with `TODO`s as well.

Related to #358, #462: being able to change the amount of time skipped by seek buttons would be nice; imo 15 seconds is too long.

Closes #43, closes #398.

PR #492, opened 2022-05-30
Hotkeys: Tweaks For Handling Live Streams

just-max:
Follow-up to #476, got caught up with other things and didn't get around to this :|
Properly handle seeking in (DVR) livestreams, since their duration is always reported as infinity.

Also possibly interesting, because I'm not sure if there are any public "live" HLS streams: I've played around with live streams by creating a stream with OBS following [these instructions](https://obsproject.com/forum/resources/how-to-do-hls-streaming-in-obs-open-broadcast-studio.945/) (but without a web server, the files can be served locally with e.g. [`npx http-server --cors -c-1`](https://www.npmjs.com/package/http-server) after changing the appropriate URLs in the database). By setting `hls_list_size=0` in the OBS "Muxer Settings" field the stream can also be used as a DVR stream (the whole thing remains seekable).

### DIFF
--- a/web/ts/hotkeys.ts
+++ b/web/ts/hotkeys.ts
@@ -4,13 +4,14 @@ import videojs from "video.js";
 // However, keycode uses deprecated event.keyCode https://github.com/timoxley/keycode/issues/52
 // The code used below avoids the mentioned IE issue (see link) by matching old key names too.
 
-// TODO: disable some actions for live streams (forward/backwards, seek to percentage, ...)
-
 // helpers
 const clamp = (value, min, max) => Math.min(Math.max(value, min), max);
+// optional type, e.g. optional(<value that may be undefined>).map((x) => <do something if not undefined>)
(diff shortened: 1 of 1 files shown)


joschahenningsen:
> Also possibly interesting, because I'm not sure if there are any public "live" HLS streams: I've played around with live streams by creating a stream with OBS following [these instructions](https://obsproject.com/forum/resources/how-to-do-hls-streaming-in-obs-open-broadcast-studio.945/) (but without a web server, the files can be served locally with e.g. [npx http-server --cors -c-1](https://www.npmjs.com/package/http-server) after changing the appropriate URLs in the database). By setting hls_list_size=0 in the OBS "Muxer Settings" field the stream can also be used as a DVR stream (the whole thing remains seekable).

Nice, didn't know obs could do that!

joschahenningsen (approved):

PR #494, opened 2022-05-31
load service worker from absolute location

joschahenningsen:
Currently the service worker can only be loaded from the main page, not from sub pages.

Because it (e.g.) tries to load the script on the admin page from `/admin/service-worker.js`  instead of `/service-worker.js`


mono424 (approved):
looks good, sorry that i missed that! :)

PR #516, opened 2022-06-04
Video Section Image Preview

MatthiasReumann:
- [x] Add thumbnail generation to worker 
- [X] Update UI

## UI

image

## Tests

Updated the download tests, so that it also uses `testutils.TestCases`. Since they succeed you can neglect them when reviewing.


joschahenningsen:
This calls a worker everytime a thumbnail is requested, the worker than loads the thumbnail and sends it back. This is a little wasteful for a number of reasons:
- communication between TUM-Live and Workers (granted, this is blazingly fast as they run in the same datacenter)
- ffmpeg probes the video
  - to know wether or not ffmpeg can extract a screenshot, it has to look up some details about the video. For example the length. We use h264 which has a moov atom at the end of the video, so ffmpeg has to read through the whole file to get a screenshot. This might take up to a second

Since many users might need the same file I suggest this workflow:

```
                         ┌──────────────┐
                         │ Ceph Storage │
                         │  (/mass/..)  │◄────────┐
                         └──────────────┘         │
                             ▲                    │
                             │                    │Put
                    Get/Has  │      No            │Thumbnail
                   Thumbnail?├────────────────┐   │
                             │                ▼   │
┌────┐ Get Thumb  ┌──────────┤              ┌─────┴──┐
│User├──────────► │ TUM-Live │              │ Worker │
└────┘            └──────────┘              └────────┘
```

Note that the ceph storage is available on both, TUM-Live and the worker. So the worker only has to save the thumbs to the storage and tum-live can get it anytime (we do the same with the converted recordings). This way the worker only has to compute the thumbs once. 

P.S. Ideally the thumbnails would get generated when a lecturer creates a section.

MatthiasReumann:
> This calls a worker everytime a thumbnail is requested, the worker than loads the thumbnail and sends it back. This is a little wasteful for a number of reasons:
> 
> * communication between TUM-Live and Workers (granted, this is blazingly fast as they run in the same datacenter)
> * ffmpeg probes the video
>   
>   * to know wether or not ffmpeg can extract a screenshot, it has to look up some details about the video. For example the length. We use h264 which has a moov atom at the end of the video, so ffmpeg has to read through the whole file to get a screenshot. This might take up to a second
> 
> Since many users might need the same file I suggest this workflow:
> 
> ```
>                          ┌──────────────┐
>                          │ Ceph Storage │
>                          │  (/mass/..)  │◄────────┐
>                          └──────────────┘         │
>                              ▲                    │
>                              │                    │Put
>                     Get/Has  │      No            │Thumbnail
>                    Thumbnail?├────────────────┐   │
>                              │                ▼   │
> ┌────┐ Get Thumb  ┌──────────┤              ┌─────┴──┐
> │User├──────────► │ TUM-Live │              │ Worker │
> └────┘            └──────────┘              └────────┘
> ```
> 
> Note that the ceph storage is available on both, TUM-Live and the worker. So the worker only has to save the thumbs to the storage and tum-live can get it anytime (we do the same with the converted recordings). This way the worker only has to compute the thumbs once.
> 
> P.S. Ideally the thumbnails would get generated when a lecturer creates a section.

I actually wanted to ask you if TUM-Live App is behind a nginx instance, so that we can circumvent the loading problem. However, I think your solution makes more sense. I would leave out the `HasThumbnail`-Step, generate thumbnails on admin-section-saving and simply static serve the images.

joschahenningsen:
Sound great! 
Tum live is behind a nginx instance but I'd not build any logic around that. We might switch to traefik soon anyways :)

api/download.go
```diff
@@ -50,25 +50,36 @@ func (r downloadRoutes) download(c *gin.Context) {
...
+			}
 		}
+		log.Info(fmt.Sprintf("Download request, user: %d, file: %d[%s]", tumLiveContext.User.ID, file.ID, file.Path))
+		sendDownloadFile(c, file)
 	}
-	c.AbortWithStatus(http.StatusForbidden)
 }
 
-func sendFile(c *gin.Context, file model.File) {
+func sendFileContent(c *gin.Context, file model.File) {
```
joschahenningsen: This adds a `image/jpeg` content type, I think a better name would therefore be 
```suggestion
func sendImageContent(c *gin.Context, file model.File) {
```

api/stream.go
```diff
@@ -315,6 +333,26 @@ func (r streamRoutes) deleteVideoSection(c *gin.Context) {
...
+		parameters := generateVideoSectionImagesParameters{
+			sections:           sections,
+			playlistUrl:        context.Stream.PlaylistUrl,
+			courseName:         context.Course.Name,
+			courseTeachingTerm: context.Course.TeachingTerm,
+			courseYear:         uint32(context.Course.Year),
+		}
+		err := GenerateVideoSectionImages(r.DaoWrapper, &parameters)
+		if err != nil {
+			log.WithError(err).Error("failed to generate video section images")
```
joschahenningsen: Why do we need to regenerate images when a section is deleted?

api/worker_grpc.go
```diff
@@ -686,6 +686,65 @@ func notifyWorkersPremieres(daoWrapper dao.DaoWrapper) {
...
+type generateVideoSectionImagesParameters struct {
+	sections                                    []model.VideoSection
+	playlistUrl, courseName, courseTeachingTerm string
+	courseYear                                  uint32
+}
+
+func GenerateVideoSectionImages(daoWrapper dao.DaoWrapper, parameters *generateVideoSectionImagesParameters) error {
+	workers := daoWrapper.WorkerDao.GetAliveWorkers()
+	workerIndex := getWorkerWithLeastWorkload(workers)
+	workers[workerIndex].Workload += 1 // Not sure if that's a reasonable amount.
```
joschahenningsen: No need to modify this variable, the worker reports its own workload that is set in its `status` package, also this is such a low workload, we don't need to increase it here imo.

api/worker_grpc.go
```diff
@@ -686,6 +686,65 @@ func notifyWorkersPremieres(daoWrapper dao.DaoWrapper) {
...
+	}()
+	if err != nil {
+		log.WithError(err).Error("Unable to dial server")
+		return err
+	}
+
+	client := pb.NewToWorkerClient(conn)
+
+	// remove old images
+	_, err = client.CleanSectionImageFolder(context.Background(), &pb.CleanSectionImageFolderRequest{
```
joschahenningsen: Wouldn't it make sense to instead of delete everything only regenerate the files the worker doesn't already have on disk?

joschahenningsen: Also I don't think this needs a extra endpoint, the worker can clean up the directory when it gets a `GenerateSectionImage` request

worker/api/api.proto
```diff
@@ -11,6 +11,28 @@ service ToWorker {
...
+
+message GenerateSectionImageRequest {
+  string PlaylistURL = 1;
+  string CourseName = 2;
+  uint32 CourseYear = 3;
+  string CourseTeachingTerm = 4;
+  uint32 Hours = 5;
+  uint32 Minutes = 6;
+  uint32 Seconds = 7;
 }
```
joschahenningsen: If we make this request send all sections that should be generated, we can omit `CleanSectionImageFolder` and do the cleanup in this request.

worker/pb/api.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.0
-// 	protoc        v3.20.1
+// 	protoc        v3.17.3
```
joschahenningsen: Not sure about the difference, probably nothing changed, but you might want to upgrade to 3.20.x

joschahenningsen (changes requested):
A few thoughts you can ignore as they are a matter of taste :D

api/download.go
MatthiasReumann: Yup, that's definitely better.

api/stream.go
MatthiasReumann: I probably should have implemented a `DeleteVideoSectionImage` request. For starters I simply regenerated everything so that we don't have unused image-files stored.

MatthiasReumann: Added a `DeleteSectionImage` request which solves that problem.

worker/api/api.proto
MatthiasReumann: Put all sections into one request. Works great, thanks!

worker/pb/api.pb.go
MatthiasReumann: Updated. Unfortunately, on MacOS the highest available version is v3.19.4.

joschahenningsen (approved):
:tada:

PR #518, opened 2022-06-05
Add null checks for user to for course overview and watch pages.

### LINKED ISSUE #517: Course Overview and Watch Page Broken for Logged-Out Users
Overview page is empty:

image

Watch page also doesn't work.

Problem: #484 (abaa27e54930a1424d85a3c68d51660bf52dfa10) added references to `.IndexData.TUMLiveContext.User` without a null-check.

just-max:
Fixes #517.

### DIFF
--- a/web/template/course-overview.gohtml
+++ b/web/template/course-overview.gohtml
@@ -9,6 +9,7 @@
 <body class="md:overflow-hidden" x-data="{searchContext: new global.SearchContext()}">
 {{- /*gotype: github.com/joschahenningsen/TUM-Live/web.CoursePageData*/ -}}
 {{$course := .IndexData.TUMLiveContext.Course}}
+{{$user := .IndexData.TUMLiveContext.User}}
 {{template "header" .IndexData.TUMLiveContext}}
 <div x-cloak x-show="searchContext.show"
      x-data="{streamSearch: new global.StreamSearch({{$course.ID}})}"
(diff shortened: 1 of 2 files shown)


web/template/watch.gohtml
```diff
@@ -180,7 +181,7 @@
                 <div class="flex my-auto">
                     <a href="/course/{{$course.Year}}/{{$course.TeachingTerm}}/{{$course.Slug}}">
                         <span class="text-5 hover:text-1 hover:underline text-lg">{{$course.Name}}</span></a>
-                    {{if or (.IndexData.TUMLiveContext.User.IsAdminOfCourse .IndexData.TUMLiveContext.Course) .IndexData.IsAdmin}}
+                    {{if and $user or ($user.IsAdminOfCourse .IndexData.TUMLiveContext.Course) .IndexData.IsAdmin}}
```
alexanderstephan, line 184:
```suggestion
                    {{if and $user (or ($user.IsAdminOfCourse .IndexData.TUMLiveContext.Course) .IndexData.IsAdmin)}}
```

web/template/course-overview.gohtml
```diff
@@ -64,7 +65,7 @@
 <div class="m-auto px-1 sm:px-3 lg:px-5 w-full xl:w-2/3" style="height: calc(100vh - (6.5rem));">
     <div class="flex my-8 pl-2 md:pl-0">
         <h1 class="font-bold text-3xl text-3 my-auto">{{$course.Name}}</h1>
-        {{if or (.IndexData.TUMLiveContext.User.IsAdminOfCourse .IndexData.TUMLiveContext.Course) .IndexData.IsAdmin}}
+        {{if and $user or ($user.IsAdminOfCourse .IndexData.TUMLiveContext.Course) .IndexData.IsAdmin}}
```
alexanderstephan, line 68:
```suggestion
        {{if and $user (or ($user.IsAdminOfCourse .IndexData.TUMLiveContext.Course) .IndexData.IsAdmin)}}
```

alexanderstephan (changes requested):
Thanks for fixing this! I think I saw this somewhere on Zulip today.
Not adding the parentheses causes this to fail for logged in users now. The result of the `or` function should be an argument that is passed to the `and` function so it needs parentheses. The syntax can be a bit confusing...

joschahenningsen:
Thanks for taking a look here! I'll close this as it's fixed in 65a633751e8c49fd97649428f152e641e52ce352 though :)

PR #523, opened 2022-06-12
Improve watch page UI

alexanderstephan:
- Fixes misalignment when in waiting room
- Fixes descriptions falsely showing show more (with new truncation logic)
- Fixes an issue where two instances of the action button are shown
- Unifies the layout of all watch page sections
- Prevents the description from changing the chat size
- Moves up the admin buttons for better reachability

Peek 2022-06-14 15-33


MatthiasReumann:
> A lot of stuff has been kinda broken. I tried to optimize everything to rescale properly. However, I am facing one issue before this is ready. When I expand a lengthy description the chat is pushed away no matter what attributes I set. I would be glad about some help from someone with more CSS experience.

What do you mean with pushed away? Space between the chat and the description?

alexanderstephan:
Sorry for being a bit vague, I was already pretty tired yesterday. What I mean can be seen here for example: https://test.live.mm.rbg.tum.de/w/airplay/353
If the text grows horizontally e.g. for large descriptions, the chat's width is way to narrow.

MatthiasReumann:
> Sorry for being a bit vague, I was already pretty tired yesterday. What I mean can be seen here for example: https://test.live.mm.rbg.tum.de/w/airplay/353
> If the text grows horizontally e.g. for large descriptions, the chat's width is way to narrow.

You might be able to fix this by adding the chat-width-counterpart classes to `<div id="watchwrapper"...>`. 

Chat currently has width classes: `w-full md:w-2/6 2xl:w-1/5`, means could you try adding `w-full md:w-4/6 2xl:w-4/5` to the `watchwrapper` container?

model/stream.go
```diff
@@ -236,3 +241,157 @@ func (s Stream) Attachments() []File {
...
+	// Copy ellipsis
+	output = append(output, []byte(ellipsis)...)
+
+	// Finally, create a closing tag for each tag in the stack.
+	for i := len(tagStack) - 1; i >= 0; i-- {
+		output = append(output, []byte(fmt.Sprintf("</%s>", tagStack[i]))...)
+	}
+
+	return output, nil
+}
```
joschahenningsen: Let's put all this in the `tools` package, having it in `model` feels off

joschahenningsen (approved):
Changes look great! :heart:

model/stream.go
alexanderstephan: Agree, I also had to take a bit of a different approach now, as using tools would have introduced a circular dependency, that would be very ugly to solve. It is kinda of a problem that `tools` depends on `dao` and `model`. Maybe we can decouple it somehow in the future.

PR #532, opened 2022-06-16
Stream Endpoint Testing

MatthiasReumann:
## Changes 
- Added Tests for Stream Endpoint

## Further Information
Currently, not everything is tested thoroughly, since that requires additional mocking: Coverage is at 71%. This might be a task for another person. 

One example of that would be the `NotifyWorkersToStopStream` function in the `worker_grpc` file. It would be nice to be able to mock things like that, however, I am not sure how much of an overhead that is.

### DIFF
--- a/api/stream.go
+++ b/api/stream.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


joschahenningsen (approved):
:heart:

PR #534, opened 2022-06-17
[Old] Resizable and Draggable Videos

### LINKED ISSUE #120: Combined presentation and camera view
In addition to the “comb-view” video there should be an option to view both camera and presentation videos side by side and modify their sizes.
Important gotchas: We need to make sure the videos are synchronized and only one of them should be un-muteable at the same time.

ata-keskin:
The goal of this pull request is to implement resizable and dragabble videos. So far this is already implemented for CAM and PRES. For COMB, I decided to use two canvases and copy over the frames from the main video-js playing somewhere on the screen. I need to add the player controls as well. I have decided to do this because I couldn't figure out how to sync two video players. I would also hugely appreciate it if someone has a solution for this and could assist me, otherwise we have to add a control bar to the bottom of the screen and go on with the canvas approach.

Apart from this I need to somehow include Jquery, JqueryUI and the css file for JqueryUI in the webpack but I haven't been able to do it succesfully. I also need to include some stylings and have written a lot of javascript. It would maybe make sense to separate that code from the html file being delivered to the client.

I also plan on adding a button to turn on this functionality instead of defaulting to this view. Feel free to test out the current status of the implementation.


web/template/watch.gohtml
```diff
@@ -12,6 +12,30 @@
     <link rel="stylesheet" href="/static/node_modules/video.js/dist/video-js.min.css">
     <link rel="stylesheet" href="/static/node_modules/videojs-seek-buttons/dist/videojs-seek-buttons.css">
     <link rel="stylesheet" href="/static/node_modules/@silvermine/videojs-airplay/dist/silvermine-videojs-airplay.css">
+
+    <link rel="stylesheet" href="//code.jquery.com/ui/1.13.1/themes/base/jquery-ui.css">
```
mono424, line 16: Wir JQueryUI wirklich gebraucht? Wir verwenden fast überall Tailwind CSS. Wenn es möglich ist würde ich versuchen statt JqueryUI Tailwind und im notfall inline-css zu verwenden :).

ata-keskin: I am using the "interactions" Draggable and Resizable that is present in JQueryUI. I am pretty sure this is something completely different than Tailwind CSS. Another option would be to implement this from scratch, but it would take me quite some time to do that since I am not that knowledgeable in javascript. Do you think this can be accomplished with Tailwind CSS and inline-css only?

ata-keskin: and this stylesheet needs to be included for JQueryUI to work. This pull request is still a draft as I have issues that I need to discuss first. Afterwards I think this needs to be included via webpack anyhow.

mono424: Yes, you would need some js too. Probably I am too biased as in my opinion it's never a good idea to use Jquery at all, but I think you can leave it like that if the others are okay with it 👍

mono424: Also just as an idea there are some more modern libs for draggable components like https://interactjs.io/ which would be lighter and has support for touch devices :)

ata-keskin: I didn't know about these! I am reading up now, thanks for the suggestions :smiley:

mono424: You are very welcome, sorry for commenting already on your draft by the way 😅

joschahenningsen:
superseded by #603

PR #546, opened 2022-06-23
show user popout on click instead of hover

### LINKED ISSUE #542: Behaviour of the Logout/Settings popover menu
**Describe the bug**
The popover menu that gets shown when you hover over your username in the top-left corner is a bit weird:
- the menu gets presented when you hover over your username
- it gets dismissed when you click your username while it is presented, but if you then click the username again, it won't reappear (which it feels like it should, given that clicking makes it disappear while presented, clicking while not presented should make it appear)
- it only gets dismissed if the cursor is moved out of the popover, but it does not get dismissed if the cursor is moved away from the username without first moving it into the popover (given that hovering above the username makes it appear, ending the hover (by moving the cursor away from the username w/out going through the popover) should also make it disappear again

**To Reproduce**
See above

**Expected behavior**
If some event (like e.g. moving the cursor into some specific area of the web page) makes the popover appear, then the opposite event (like e.g. moving the cursor away from that specific area of the web page) should trigger the opposite action, i.e. make the popover disappear.
(The same goes for dismissing/presenting via a click on the username)

**Screenshots**
n/a

 - OS: macOS
 - Browser Chrome
 - Version 102.0.5005.115

n/a, didn't try

**Additional context**
I realise that this is pedantic and probably not that important to anyone except me, so feel free to ignore.

One more thing: I feel like the order of the "Logout" and "Settings" links should probably be reversed, since that's how most web apps seem to order these actions (i.e., having the Logout link as the very last).

joschahenningsen:
fixes #542

### DIFF
--- a/web/template/header.gohtml
+++ b/web/template/header.gohtml
@@ -22,7 +22,7 @@
                 {{else}}
                     <div class="inline-block align-middle mt-0 mr-4" x-data="{ show: false }"
                          @click.outside="show=false">
-                        <a href="#" class="text-5 hover:text-1 font-medium" @click="show=!show">
+                        <a role="button" class="text-5 hover:text-1 font-medium" @click="show=!show">
                             {{.User.GetPreferredName}}
                             <i class="ml-1 fa-angle-down fas"></i>
                         </a>
--- a/web/template/header.gohtml
+++ b/web/template/header.gohtml
@@ -24,7 +24,7 @@
                          @click.outside="show=false">
                         <a href="#" class="text-5 hover:text-1 font-medium" @click="show=!show">
                             {{.User.GetPreferredName}}
-                            <i class="fa-chevron-down fas"></i>
+                            <i class="ml-1 fa-angle-down fas"></i>
                         </a>
                         <div class="relative">
                             <div x-cloak x-show="show"
--- a/web/template/header.gohtml
+++ b/web/template/header.gohtml
@@ -22,11 +22,11 @@
                 {{else}}
                     <div class="inline-block align-middle mt-0 mr-4" x-data="{ show: false }"
                          @click.outside="show=false">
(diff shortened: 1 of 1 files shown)


web/template/header.gohtml
```diff
@@ -22,11 +22,11 @@
                 {{else}}
                     <div class="inline-block align-middle mt-0 mr-4" x-data="{ show: false }"
                          @click.outside="show=false">
-                        <a href="#" class="text-5 hover:text-1 font-medium" @mouseenter="show=true">
+                        <a href="#" class="text-5 hover:text-1 font-medium" @click="show=!show">
```
alexanderstephan:
```suggestion
                        <a class="text-5 hover:text-1 font-medium" @click="show=!show">
```
Is there any use case for scrolling up when using the menu?

web/template/header.gohtml
```diff
@@ -22,11 +22,11 @@
...
                             {{.User.GetPreferredName}}
                             <i class="fa-chevron-down fas"></i>
```
alexanderstephan:
```suggestion
                            <i class="ml-1 fa-angle-down fas"></i>
```

alexanderstephan: This could be also an option as I think the chevron looks a bit too huge in our case.

alexanderstephan (changes requested):
Works as expected :+1:  Just noted some things I noticed

web/template/header.gohtml
joschahenningsen: The href makes the anchor behave like a link (mouse pointer on hover etc.) I replaced it with role="button" Can you check if this messes up the appearance in safari?

alexanderstephan: Works fine on Safari Version 15.6 (17613.3.3)

alexanderstephan (approved):
Looks good :heart:

PR #549, opened 2022-06-27
Course Endpoint Testing

MatthiasReumann:
## Changes 
- Add course endpoint tests
- Replace `ReadAll(...)` and `json.Unmarshal(...)` with `c.BindJSON(...)` in `submitCut`, `addUnit` 
    If there was any particular reason not to use `c.BindJSON(...)`, I can change that back to what it was.
- Sometimes `panic` is called if an error occurs. I replaced the `panic`'s with `c.AbortWithStatus(500)`
- Make `testCases.Run` more robust and flexible by making it an `interface{}` with type-checkin.
- Remove sexy endpoint

## Todos 

- [x] Check if Content-Type `application/json` is set for `submitCut` and `addUnit` requests in typescript.

## Follow up
Refactor the old tests to utilize the type-checking.


api/sexy.go
```diff
@@ -7,7 +7,7 @@ import (
 	"time"
 )
 
-func configGinSexyApiRouter(router gin.IRoutes, daoWrapper dao.DaoWrapper) {
+func configGinSexyApiRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
```
alexanderstephan: Can you remove the sexy ap (and also sadly the tests)? This was not used and the project that intended to use it is now archived.

alexanderstephan (dismissed):
Looks good, so many tests 💯 Also some good refactoring 👍 Just spoke with Joscha, and noticed that the sexy stuff can be removed 😁

joschahenningsen (approved):
Thanks a ton!

PR #556, opened 2022-06-29
Add x-cloak to info-dropdown

MatthiasReumann:
## Changes 

- Add `x-cloak` attribute to info-dropdown container

## Why 

On Reload or redirection to another page the dropdown was shown for a tiny amount of time until `hidden` was loaded. That's ugly.


joschahenningsen (approved):
Ahh yes noticed that too, thanks!

PR #575, opened 2022-07-04
Most replayed Heatmap (Logging)

### LINKED ISSUE #519: 'Most Replayed' Feature for Identifying Most Popular Parts of Videos
It's hard to find the most relevant/best parts of a VoD.

On other video platforms like youtube, there is a heatmap displayed over the timeline which are representing how often users have been jumped to that specific location to replay the part of the video. The most relevant locations are then visible in the heatmap displayed over the timeline to quickly find them.

Bildschirmfoto 2022-06-06 um 13 39 16

### Data Collection

My idea would be to have a video cutted in n sections (e.g. 100). For each seeking, increment counter on that section. To safe storage we could only store sections with at least 1 seek in the databse.

-> Then the heatmap can be created easily by interpolation.

mono424:
Splitted the heatmap implementation into two parts:
- Logging algorithm to track where people seek to
- Visuals to display the heatmap

This Pullrequest only contains the logging logic.


dao/video-seek.go
```diff
@@ -0,0 +1,52 @@
+package dao
+
+import (
+	"github.com/joschahenningsen/TUM-Live/model"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
```
MatthiasReumann, line 10: `//go:generate mockgen -source=video-seek.go -destination ../mock_dao/video-seek.go` missing

api/progress.go
```diff
@@ -67,6 +67,9 @@ func configProgressRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 	go progressBuff.run()
 	router.POST("/api/progressReport", routes.saveProgress)
 	router.POST("/api/watched", routes.markWatched)
+
+	router.POST("/api/seekReport/:streamID", routes.reportSeek)
```
MatthiasReumann: Could you add some tests for those two routes? The tests will be pretty similar to the ones in `search_test.go`, so you might want to have a look at them.

web/ts/global.ts
```diff
@@ -12,6 +12,7 @@ export async function putData(url = "", data = {}) {
 }
 
 export async function postData(url = "", data = {}) {
+    console.log(data);
```
MatthiasReumann: `- console.log(data)`

dao/video-seek.go
```diff
@@ -0,0 +1,52 @@
...
+	return videoSeekDao{db: DB}
+}
+
+func (d videoSeekDao) Add(streamID string, pos float64) error {
+	var stream *model.Stream
+	if err := DB.First(&stream, "id = ?", streamID).Error; err != nil {
+		return err
+	}
+
+	hitPos := float64(stream.Duration) * pos
```
MatthiasReumann: In typescript `pos` is calculated by `player.currentTime() / player.duration()` and this value is then multiplied with `stream.Duration` in `.Add(...)`. 

Shouldn't this be the same as `player.currentTime()`? Or is the problem that `stream.Duration != player.duration()`?

MatthiasReumann (commented):
I am not sure if this PR is reviewable yet. However, I was super interested in how you implemented it, so, if it isn't __opened__ officially let's call this _early feedback_. 😅

Nevertheless, looks pretty cool 💯

mono424:
@MatthiasReumann Thank you for your feedback!

@MatthiasReumann @joschahenningsen Lets discuss if its probably better to use progressReport to increment section counters for the video :).

dao/video-seek.go
mono424: Yes that would be much more straight forward! I think the duration should be the same 👍

mono424:
> I am not sure if this PR is reviewable yet. However, I was super interested in how you implemented it, so, if it isn't **opened** officially let's call this _early feedback_. 😅
> 
> Nevertheless, looks pretty cool 💯

Thanks for your feedback! The testutils you have written are pretty sick!! 🚀

api/seek_stats.go
```diff
@@ -0,0 +1,64 @@
...
+type seekStatsRoutes struct {
+	dao.DaoWrapper
+}
+
+// configSeekStatsRouter sets up the router and initializes a progress buffer
+// that is used to minimize writes to the database.
+func configSeekStatsRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
+	routes := seekStatsRoutes{daoWrapper}
+
+	router.POST("/api/seekReport/:streamID", routes.reportSeek)
```
MatthiasReumann, line 20: Currently anyone can `post` reports. There is no "firewall" so to say. I don't know how problematic that is but malign folk might abuse that behavior.

A naive solution would be to only report logged-in seeks. That, however, potentially leads to fewer data points.

Otherwise we could install some sort of "spam-protection". This should be configurable in the nginx instance TUM-Live uses or maybe already is. There might even be some `gin-gonic` features that can deal with that problem.

MatthiasReumann (approved):
Looks great 💯 

If this isn't merged before #549 you can even use the _\*\*new and improved\*\*_ testutils. Then  `testutils.First(...)` and `json.Marshal(...)` are not necessary anymore and you can specify the body as 

```golang
{
    ...
    Body: gin.H{"position": testPosition}
    ...
}
```

I've left a comment regarding security concerns. But we might want to discuss that in a meeting.

api/seek_stats.go
mono424: Yes already thought about this in the beginning.

We wanted to store no userdata related to the seeks, we could store users who trigger the endpoint separately from the seek data, so there is no link between the user and the seek-data but still we would kind of capture the users who trigger the endpoint somehow... Idk :)

We could limit invocation by IP in a certain time frame. Probably better to set that on nginx level ? Not sure :)

PR #576, opened 2022-07-04
Video heatmap visuals

mono424:
Splitted the heatmap implementation into two parts:

- Logging algorithm to track where people seek to
- Visuals to display the heatmap
This Pullrequest only contain the visuals to display the heatmap

Many Data Points:
Bildschirmfoto 2022-07-10 um 16 20 51

Few Data Points:
Bildschirmfoto 2022-07-10 um 16 22 53


api/seek_stats.go
```diff
@@ -40,6 +40,9 @@ func (r seekStatsRoutes) reportSeek(c *gin.Context) {
 	}
 }
 
+const minTotalHits = 150
+const minNonZeroChunks = 50
```
MatthiasReumann: For consistency:

```golang 
const (
    minTotalHits     = 150
    minNonZeroChunks = 50
)
```

web/ts/watch.ts
```diff
@@ -257,6 +258,107 @@ export const videoStatListener = {
     },
 };
 
+/*const heatMapExamplePath =
```
MatthiasReumann: I am uncertain if we want to keep this or not. For testing purposes it is probably quite helpful, however, all of us could theoretically import a new database dump right?

web/ts/watch.ts
```diff
@@ -257,6 +258,107 @@ export const videoStatListener = {
...
+*/
+
+const repeatMapScale = 90;
```
MatthiasReumann: Could you please move the heat map code to a separate file, e.g. `heat-map.ts`? Thanks!

web/ts/watch.ts
```diff
@@ -257,6 +258,107 @@ export const videoStatListener = {
...
+        const slope = (lastVal + 100) / 2;
+
+        res += ` C 1001.0,${lastVal} 1000.0,${slope} 1000.0,100.0`;
+
+        return res;
+    },
+
+    data() {
+        return "";
+    },
```
MatthiasReumann: Couldn't find any usages of this function. 😅
```suggestion
```

web/template/watch.gohtml
```diff
@@ -116,10 +116,45 @@
...
+            <div
+                    x-data="{ size: { x: '0px', width: '0px' }, seekLogger: new watch.SeekLogger('{{$stream.ID}}'), heatMapPath: null }"
+                    x-on:updateheatmappath.window="e => { heatMapPath = e.detail.heatMapPath; }"
+                    x-on:updateheatmapsize.window="e => { size = e.detail; }"
+                    x-init="seekLogger.attach(); watch.repeatHeatMap.init('{{$stream.ID}}');"
+                    class="heatmap-wrap absolute bottom-[18px] rounded bg-gradient-to-t from-gray-800/[.6] pt-4 pointer-events-none"
+                    :style="{width: size.width, left: size.x}"
+                    x-show="heatMapPath != null"
+            >
+                <svg x-if="heatMapPath" height="100%" preserveAspectRatio="none" viewBox="0 0 1000 100"
```
MatthiasReumann: The `x-if` directive is only allowed on `<template>` tags. (See [this link](https://koenwoortman.com/alpinejs-cannot-read-property-clonenode-of-undefined/) for more information). 

Since currently Alpine throws this error:
image

MatthiasReumann (changes requested):
I think this is one of the coolest features implemented in the last couple of months. Great job. 🫡

Potential bug I found: When skipping forward (or backwards) the heat map wiggles a bit (Using Chrome):

https://user-images.githubusercontent.com/29633518/210318827-1faf3bdf-2aed-4763-8366-481828c3d89c.mov

Also make sure to merge `dev` properly into this branch, so that the bug fixed in [#852](https://github.com/joschahenningsen/TUM-Live/pull/852) isn't reintroduced.

mono424:
@MatthiasReumann the wiggeling is because of the time label that changes its length. It also changes the length of the seekbar, thats why the heatmap changes its size too. I didnt want to change so much inside video.js, but I see that this can be somehow anoying, what do you think, should we mess with videojs css and fix it, what do you think? :)

MatthiasReumann:
> @MatthiasReumann the wiggeling is because of the time label that changes its length. It also changes the length of the seekbar, thats why the heatmap changes its size too. I didnt want to change so much inside video.js, but I see that this can be somehow anoying, what do you think, should we mess with videojs css and fix it, what do you think? :)

Ouh. Definitely not worth the time.

mono424:
> > @MatthiasReumann the wiggeling is because of the time label that changes its length. It also changes the length of the seekbar, thats why the heatmap changes its size too. I didnt want to change so much inside video.js, but I see that this can be somehow anoying, what do you think, should we mess with videojs css and fix it, what do you think? :)
> 
> Ouh. Definitely not worth the time.

Yes lets go for a followup PR 👍

MatthiasReumann (approved):
Looks great. Thanks again 🚀

PR #577, opened 2022-07-05
Consistent Error Handling

### LINKED ISSUE #567: Error handling middleware
I think it would be pretty cool to have a middleware for consistent error handling. 

For example, we could attach errors with `c.Error()` and then collect them afterwards with `c.Errors`.

```golang 
func handleRequest(c *gin.Context){
   ...
   if err != nil {
      c.Error(err)
      return
   }
}
```

This improves the API-Design since we can return a lovely error-JSON such as 

```json
{
        "code": 500,
        "message":"can not perform search",
        "error": "<some error thrown by gorm>"
}
```

For some reference, this is [how Google's error response looks like](https://cloud.google.com/storage/docs/json_api/v1/status-codes).

MatthiasReumann:
## Changes 
- Add `RequestError` struct
- Add `Errorhandler` middleware 

## TODO

- [x] Update _every_ handler.

## Usage 

```golang
if err != nil {
    _ = c.Error(tools.RequestError{
	Status:        http.StatusInternalServerError,
	CustomMessage: "can not perform search",
	Err:           err,
    })
    return
}
```


tools/api-errors.go
```diff
@@ -0,0 +1,49 @@
...
+		res["error"] = r.Error()
+	}
+
+	return res
+}
+
+func ErrorHandler(c *gin.Context) {
+	c.Next()
+
+	if len(c.Errors) > 0 {
```
joschahenningsen, line 35: IMO for good measure this middleware should also abort the context if there are errors, so the request is not passed to other potential middlewares or handlers afterward.

joschahenningsen (changes requested):
Just one quick remark, looks great otherwise!

tools/api-errors.go
MatthiasReumann: Added `c.Abort()` if an error occurs 👍🏻

joschahenningsen (approved):
Thanks! ❤️

PR #535, opened 2022-06-18
Keep current section displayed & VS Mobile UI

MatthiasReumann:
## Changes 
- Update Mobile UI 
- New button `(Follow Sections)` to automatically switch pages.
- Move all video-section code to `video-sections.ts` and add TypeDoc-Style comments

### Mobile UI
image

**Minimized View**
image

PR #578, opened 2022-07-05
Fix elForSpot Alpine Error

MatthiasReumann:
## Changes 
- Fixes `elForSpot` Alpine Error after the creation of new sections. (See #535)
- Add invalid `fileID` check.


alexanderstephan (approved):
Works perfectly now, thanks! :100:

PR #579, opened 2022-07-05
Pin courses to the top

### LINKED ISSUE #373: Pin courses to the top
Courses could be pinnable to the top of the start page. This could be used for interesting courses that one is not registered for and that are publicly available. This could be also across different semesters e.g. when using an old recording. It could land in another section like "My courses". Courses could have a small menu for actions with a pin needle.

ata-keskin:
I added a context menu to each course-card and the ability to pin/unpin courses to the top by clicking on the pin button in the context menu. Currently I am storing the html data of the course-card in local storage and using it to render the "Pinned Courses" section. I think the functionality is complete, but I will be on the lookout for potential bugs.


alexanderstephan (changes requested):
I really like this UI wise, also works well 👍

Personally I constantly switch devices and want always want to have my settings there. For that it would be cool to actually save it in the database instead of local storage.

Maybe we could also have `PinnedCourses` in the `user` model like it is done for `Courses` and `AdministeredCourses`? The frontend logic should stay more or less the same.

ata-keskin:
> I really like this UI wise, also works well +1
> 
> Personally I constantly switch devices and want always want to have my settings there. For that it would be cool to actually save it in the database instead of local storage.
> 
> Maybe we could also have `PinnedCourses` in the `user` model like it is done for `Courses` and `AdministeredCourses`? The frontend logic should stay more or less the same.

Alright, I think that should be doable. If we do it like that I can also use the gotemplate "course-card" for generating the divs under Pinned Courses, since the page will be delivered from the server with the Pinned Courses section already constructed, as opposed to re-rendering it via javascript.

alexanderstephan:
I think two additional REST calls for pin and unpin should be enough. Maybe you can do the initialization like I did it for the marked watch feature, where I initalize the go template with the database entries as JSON.

api/users.go
```diff
@@ -248,6 +255,44 @@ func (r usersRoutes) addSingleUserToCourse(name string, email string, course mod
 	}
 }
 
+type pinRequest struct {
```
alexanderstephan: Can you inline the definition of `pinReques` into `pinCourse`? It is not used outside that method, so we want to hide it.

web/template/course-card.gohtml
```diff
@@ -31,17 +42,16 @@
...
                         <i class="fas fa-angle-right"></i>
                         <span class="underline">Join waiting room</span>
                     </a>
                 {{else}}
-                    {{$next := .GetNextLecture.Start}}Next lecture: {{printf "%v %02d. %02d:%02d" $next.Month $next.Day $next.Hour $next.Minute}}
+                    {{$next := $course.GetNextLecture.Start}}Next lecture: {{printf "%v %02d. %02d:%02d" $next.Month $next.Day $next.Hour $next.Minute}}
                 {{end}}
             {{else}}
                 No upcoming Lectures
             {{end}}
```
alexanderstephan:
```suggestion
            {{end}}
            </p>
```

alexanderstephan: Technically, this `</p>` should not be needed, but in general well-formed HTML is preferred.

alexanderstephan (changes requested):
Works as expected now 👍 Just some nits and then this is ready.

web/template/course-card.gohtml
ata-keskin: I think I deleted that without looking since GoLand suggested that the tag ```</p>``` matched nothing. I didn't see the starting tag on top.

joschahenningsen (approved):
Looks great! Feel free to merge after resolving the conflicts :)

alexanderstephan (approved):
Nice work! :100: Really looking forward to using this :tada:

PR #582, opened 2022-07-11
Regenerate preview images for lectures

alexanderstephan:
This PR enables:
- Generation of thumbnails on demand for a given stream via the admin interface. This includes video section and timeline thumbnails.
- An API endpoint that triggers regeneration of thumbnails for all VoDs. For each VoD path a request is sent to a worker. This has the advantage that we can split the regeneration tasks to different workers.
- A helper program that traverses all streams with VoDs and regenerates thumbnails for them.


joschahenningsen (approved):
@alexanderstephan  From my perspective this looks good now, feel free to have another look :)

alexanderstephan:
Yeah, looks great now! 🎉

alexanderstephan:
Ah, just one thing. Have you tested video sections previews?

joschahenningsen:
Let's see what happens :D 

> Ah, just one thing. Have you tested video sections previews?

I have not. How do they relate to this PR? Should they be generated as well?

alexanderstephan:
Yes, they should also get generated. At least there is still code left for that in `RegenerateThumb`.

PR #587, opened 2022-07-14
Fixes to hotkey handling

### LINKED ISSUE #530: Document keyboard shortcuts
Just lost my position in a lecture recording as a result of using the cmd+1 keyboard shortcut (i'm on a mac, so i guess the linux/windows equivalents would be different).
The shortcuts `cmd+n` (with $n \in \\{1, 2, ..., 8\\}$) in Chrome will usually switch to the n-th tab from the left, but it seems like the video player overrides these commands to jump to different sections of a video. However, this is not documented anywhere.

Document the keyboard commands.
Many web-based applications (such as e.g. Gmail or GitHub) are using the shift+? keyboard shortcut to present an overlay/menu listing all possible keyboard shortcuts, so this might be a good thing to do as well.
Documenting the commands would:
- make it clearer what they do exactly (i'm still not sure what exactly the overwritten cmd+n commands do. do they always jump in 14-minute intervals, or do they jump to 1/10-of-the-whole-video timestamps?)
- make the commands more discoverable. (after realising that the cmd+n commands exist i continued to press some random buttons, and it seems like `j`, `k`, `l`, `m`, `z`, and `x` also are mapped to certain actions, but then again it is impossible to easily know what they do exactly, without delving into the source code of the website)

- Just not do anything and leave it as-is
- Add a section listing all available keyboard shortcuts somewhere else (e.g. in the settings page)
- Have some kind of questionmark-button at the bottom right of the video player, next to the camera mode and feedback buttons

### LINKED ISSUE #545: Volume control keys shouldn't change stream player volume
**Describe the bug**
Using media keys to control the volume changes the stream player's volume. This doesn't make sense as I already modify my system volume.

**To Reproduce**
1. Open any livestream
2. Use your keyboard's media keys for volume up or volume down

**Expected behavior**
The media player's internal volume stays the same, the system's volume changes

 - OS: Windows 10
 - Browser: Google Chrome
 - Version: 102.0.5005.115

**Additional context**

### LINKED ISSUE #570: Video player captures Cmd+L keyboard shortcut
**Describe the bug**
The video player offers the `j`, `k`, and `l` shortcuts to seek back, pause/unpause, and seek forward in a video.
However, the forward-seek is also triggered when pressing `cmd+l` (on a mac, i guess it would be different on linux/windows), which ends up overriding the browser's built-in `cmd+l` "focus on the address bar" shortcut.

Interestingly this only seems to happen with `cmd+l`, pressing `option+l` does not skip forward in the video.
The same (`cmd+X` being captured in addition to just `X`) probably applies to some other shortcuts, but `l` was the only one i was able to find.

**To Reproduce**
1. Go to the home page (ie, https://live.rbg.tum.de)
2. Press `cmd+l` to make the address bar first responder (works fine)
3. Start a video 
4. Press `cmd+l` to make the address bar first responder (does not work as expected, instead the video player will seek forward)

**Expected behaviour**
The video player should only capture the `l` shortcut, and should ignore `cmd+l`.

**Screenshots**
n/a

 - OS: macOS 12.3.1 (21E258)
 - Browser: Chrome 103.0.5060.53 (Official Build) (arm64) 

n/a didn't try

### LINKED ISSUE #581: Media keys shouldn't change the stream/video directly
**Describe the bug**
The media keys, i.e., play/pause, volume up, volume down, etc., currently change the stream *and* the operating system. This means, that when pressing the volume up button, it'll change the volume of the operating system - which is the desired behavior - but also the stream's individual volume, which is the undesired behavior. You can also see this with the play/pause button. It works properly when the stream tab is not focused, but doesn't work when it is focused, i.e., pressing it will execute the command twice, meaning nothing happens and the stream still continues (or stays paused).

**To Reproduce**
* Focus stream tab
* Press the Play/Pause-button or a volume button.

**Expected behavior**
The buttons should not change the stream itself. Let the operating system do its job.

**Desktop:**
 - OS: Windows, Ubuntu
 - Browser: Edge (chromium)

just-max:
- Don't intercept media keys
  - resolves #545
  - resolves #581

  I'd be surprised if anyone was expecting the media keys to directly control the player and it was causing troubles.

- Don't handle hotkeys if alt, ctrl or meta modifier is pressed
  - resolves the hotkey stealing problem in #530
  - resolves #570
  
  Doing it like this still feels incorrect, because I'm not sure if someone has a keyboard where they still expect the shortcut to be sent to the player while pressing a modifier. Also not sure about composing keys/IMEs. However, this should at least make things work in more cases.

### DIFF
--- a/web/ts/hotkeys.ts
+++ b/web/ts/hotkeys.ts
@@ -117,14 +117,13 @@ export const defaultOptions = {
             handle: handleFullscreen,
         },
         mute: {
-            match: ["m", "M", "AudioVolumeMute", "VolumeMute"],
+            match: ["m", "M"],
             handle: handleWithClick("MuteToggle"),
             icon: volumeIcon(true),
         },
         // "Spacebar" is for IE+old Firefox
         playPause: {
-            // TODO: pause/play actions could depend on whether player is playing (e.g. "MediaPause" only pauses)
-            match: ["k", "K", " ", "Spacebar", "MediaPause", "Pause", "MediaPlay", "MediaPlayPause"],
+            match: ["k", "K", " ", "Spacebar"],
             handle: handleWithClick("PlayToggle"),
             icon: (player) => vjsIcon(player.paused() ? "pause" : "play"),
         },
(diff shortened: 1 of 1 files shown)


joschahenningsen (approved):
Looks great, thanks for the follow up!

PR #588, opened 2022-07-15
Fix 'follow sections' jumps to not existing page

MatthiasReumann:
## Changes 
- Add `validHighlightIndex` check

### Description 
This bug exists since `highlight(...)` is only called after the player is ready.
In particular this means that `currentHighlightIndex` is in an invalid state and still set to `-1`. Then, a click on the `Follow Sections` button was leading to the disappearing of the sections because of an invalid page index.


alexanderstephan (approved):
lgtm 👍

PR #593, opened 2022-07-18
x-cloak course menu

joschahenningsen:
This PR makes sure the course menu (for pinning or hiding) stays hidden while the page loads by utilizing alpines x-cloak directive https://alpinejs.dev/directives/cloak


alexanderstephan (approved):
Nice work! 🎉

PR #596, opened 2022-07-19
Restructure stream router definitions

MatthiasReumann:
## Changes 
- Refactor route definitions for `configGinStreamRestRouter(...)` for readability reasons

## Further considerations 
We could think about splitting up the big router configuration (such as streams and courses) functions into smaller ones.

### DIFF
--- a/api/stream.go
+++ b/api/stream.go
@@ -28,31 +28,34 @@ const (
 func configGinStreamRestRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 	routes := streamRoutes{daoWrapper}
 
-	// group for api users with token
-	tokenG := router.Group("/")
-	tokenG.Use(tools.AdminToken(daoWrapper))
-	tokenG.GET("/api/stream/live", routes.liveStreams)
+	stream := router.Group("/api/stream")
+	{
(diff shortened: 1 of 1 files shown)


api/stream.go
```diff
@@ -28,31 +28,34 @@ const (
...
+		{
+			// Admin-Only Endpoints
+			admins := streamById.Use(tools.AdminOfCourse)
+			admins.GET("", routes.getStream)
+			admins.GET("/pause", routes.pauseStream)
+			admins.GET("/end", routes.endStream)
+			admins.POST("/issue", routes.reportStreamIssue)
+			admins.PATCH("/visibility", routes.updateStreamVisibility)
+
+			admins.POST("/sections", routes.createVideoSectionBatch)
```
alexanderstephan: We could also group here

alexanderstephan (approved):
This is way more readable :+1: We should also group/format the rest of api at some point to have consistency.

api/stream.go
MatthiasReumann: Great idea - updated it. Could you take another look at it? Thanks! @alexanderstephan

alexanderstephan: Looks great now!

PR #602, opened 2022-07-21
Unify Websockets

### LINKED ISSUE #594: Unify Websockets with PubSub logic
Currently we have to add a new Websocket endpoint for every websocket feature we want to use

It would be better to have a single endpoint and subscribe to the channels you want to connect to.

mono424:
Allows communicating in different channels over one socket.
Supports params in channel names like `chat/:streamID`.

Bildschirmfoto 2022-08-04 um 16 00 50


mono424:
Added more tests to [PR#644](https://github.com/joschahenningsen/TUM-Live/pull/644) as there is a layer which makes testing easier (or even possible).

api/chat.go
```diff
@@ -6,81 +6,72 @@ import (
...
-		m = melody.New()
-	}
+const (
+	ChatRoomName = "chat/:streamID"
+)
 
-	m.HandleConnect(connHandler)
+var routes chatRoutes
+
+//var m *melody.Melody
```
MatthiasReumann: I guess we don't need that out-commented code anymore. Therefore, could you remove it please?

api/chat.go
```diff
@@ -92,6 +83,21 @@ func configGinChatRouter(router *gin.RouterGroup, daoWrapper dao.DaoWrapper) {
...
+	routes = chatRoutes{daoWrapper}
+
+	wsGroup := router.Group("/:streamID")
+	wsGroup.Use(tools.InitStream(daoWrapper))
+	wsGroup.GET("/messages", routes.getMessages)
+	wsGroup.GET("/active-poll", routes.getActivePoll)
+	wsGroup.GET("/users", routes.getUsers)
+	/*wsGroup.GET("/ws", routes.ChatStream)
+	if m == nil {
+		log.Printf("creating melody")
```
MatthiasReumann: Here too.

web/ts/socket.ts
```diff
@@ -0,0 +1,141 @@
...
+type MessageHandlerFn = (payload: object) => void;
+
+const RealtimeMessageTypes = {
+    RealtimeMessageTypeSubscribe: "subscribe",
+    RealtimeMessageTypeUnsubscribe: "unsubscribe",
+    RealtimeMessageTypeChannelMessage: "message",
+};
+
+export const realtime = {
+    _debugging: false,
```
MatthiasReumann: If you'd use TypeScript classes you could utilize TypeScripts `private` visibility keyword vor the `_<name>` attributes and methods.

MatthiasReumann (commented):
Works great 💯 Looks also pretty clean code-wise! 

I've left some minor comments. Furthermore, there are some `.DS_Store` files in a bunch of folders which need to be removed. Thanks!

api/chat.go
mono424: mac ... 👍

mono424:
> Works great 💯 Looks also pretty clean code-wise!
> 
> I've left some minor comments. Furthermore, there are some `.DS_Store` files in a bunch of folders which need to be removed. Thanks!

thanks! yeah mac... will remove them ! 👍

web/ts/socket.ts
mono424: good idea! Refactoring to a singleton ts-class! 👍

MatthiasReumann (approved):
Looks great 🌟 

One remark: If the changes in the `package-lock.json` weren't intended, revert to the old version please.

joschahenningsen (approved):
Thanks for all the work! Some remaining nits can be applied in #644

PR #603, opened 2022-07-22
Multiple Synced Players

ata-keskin:
### Motivation and Context
Adds multiple synced players as a viewing option. The current split-view implementation is very primitive and could
be enhanced in a later PR. An example off the top of my head: Show the chat below the CAM version for a bigger PRES stream. 

 ### Description
Adds View Version `STEP` with a ratio-picker with predefined ratios (3:1, 1:1, 1:3) and a splitter (split.js) for a fully customized view.

 ### Steps for Testing
 1. Navigate to a livestream which has a `PRES` and a `CAM` version
 2. Click on the split-view button in the actions-bar. You will be redirect to `.../SPLIT`
 3. Hovering over the video makes the ratio-picker visible
 4. Choose a ratio. Controls are only visible for "focused" player.
 6. Make sure to test `CAM`, `PRES`, and `COMB` on its own
 7. Make sure to test the chat replay

 ### Screenshots

#### View selector

<div style='display:flex'>
image

image
</div>

#### Split menu

https://user-images.githubusercontent.com/29633518/200484671-424ceacc-18c8-4c3d-88c1-208cc970acd4.mov


alexanderstephan:
Can you fix the CI issues? If it is not possible, you can also surpress the issues.

web/ts/watch.ts
```diff
@@ -227,20 +226,22 @@ export const videoStatListener = {
         this.update();
     },
     update() {
-        const player = getPlayer();
-        const vhs = player.tech({ IWillNotUseThisInPlugins: true }).vhs;
-        const notAvailable = vhs == null;
+        const players = getPlayers();
+        for (let j = 0; j < 1; j++) {
```
alexanderstephan: This doesn't look quite right. Right now, the loop isn't needed.

web/ts/watch.ts
```diff
@@ -227,20 +226,22 @@ export const videoStatListener = {
...
-            mediaRequests: notAvailable ? 0 : vhs.stats.mediaRequests,
-            mediaRequestsFailed: notAvailable ? 0 : vhs.stats.mediaRequestsErrored,
-        };
-        const event = new CustomEvent("newvideostats", { detail: data });
-        window.dispatchEvent(event);
+            const data = {
+                bufferSeconds: notAvailable ? 0 : players[j].bufferedEnd() - players[j].currentTime(),
+                videoHeight: notAvailable ? 0 : vhs.playlists.media().attributes.RESOLUTION.height,
+                videoWidth: notAvailable ? 0 : vhs.playlists.media().attributes.RESOLUTION.width,
+                bandwidth: notAvailable ? 0 : vhs.bandwidth, //player.tech().vhs.bandwidth(),
```
alexanderstephan:
```suggestion
                bandwidth: notAvailable ? 0 : vhs.bandwidth,
```

web/ts/TUMLiveVjs.ts
```diff
@@ -203,70 +210,169 @@ export const skipSilence = function (options) {
...
+                players[j].on("canplaythrough", () => {
+                    // Can be executed multiple times during playback
+                    if (!iOSReady) {
+                        players[j].currentTime(lastProgress * duration);
+                        iOSReady = true;
+                    }
+                });
+            }
+
+            const reportProgress = () => {
```
alexanderstephan: I think the progress report does not work anymore. 🤔

web/ts/splitview.ts
```diff
@@ -0,0 +1,63 @@
...
+    static Options = {
+        FullPresentation: 0,
+        FocusPresentation: 25,
+        SplitEvenly: 50,
+        FocusCamera: 75,
+        FullCamera: 100,
+    };
+
+    constructor() {
+        this.camPercentage = SplitView.Options.SplitEvenly;
```
alexanderstephan:
```suggestion
        this.camPercentage = SplitView.Options.FocusPresentation;
```

alexanderstephan (changes requested):
Looks really awesome. 🚀 Only some small suggestions, then I think this is good to go! 👍

web/ts/TUMLiveVjs.ts
MatthiasReumann: Locally the progress report still works. Why do you think it doesn't? 🧐

alexanderstephan: Hm, it actually works now for me, but there are two requests sent simultaneously. Can we disable the progress reports for the second player? As the progress reports are sent very often, I think sending double the amount is quite some overhead.

alexanderstephan (changes requested):
Looks good now. 👍 Just more one question regarding the progress report. Otherwise, this looks fantastic and mergeable imo. @joschahenningsen what do you think?

alexanderstephan (approved):
Thanks, looks all good to me now! 💯

joschahenningsen:
I'm in love with this! 

I noticed however that sometimes both videos are unmuted which sounds weird. Ideally the camera version should be unmuted and the presentation should be muted. I don't know how to handle unmuting and muting the video when the presentation has focus tho...

joschahenningsen:
I deployed this today for testing and it seems like it breaks the embed: https://github.com/joschahenningsen/TUM-Live/issues/844

MatthiasReumann:
> I deployed this today for testing and it seems like it breaks the embed: #844

I will have a look at the sound and embedded player problems tomorrow. Thanks for pointing out!

MatthiasReumann:
> I deployed this today for testing and it seems like it breaks the embed: #844

The issues should be fixed now. Thanks again! I guess that's what happens when you test everything on mute 🫠

PR #606, opened 2022-07-25
Retry dial-in after dynamic backoff

alexanderstephan:
**Motivation**: When deploying to the testserver, the worker has to be restarted as it starts faster than the backend and terminates early since it can't find the backend.

- Adds a backoff to the worker grpc client dial in using https://pkg.go.dev/google.golang.org/grpc@v1.48.0/backoff. Params are somewhat arbitrary: Start retrying every second, if failed multiply by 1.6 and cap it at 15s. I am very open to alternative values here.
- Dial is blocking now to prevent the status cronjobs from getting started when there is no backend to connect to.
- Moved the grpc setup into main as it does not belong to `cfg` semantically.
- I left the current behavior regarding joining the backend as is, since a fatal error is a good hint that the general setup seems to be faulty. Otherwise it might run and retry forever. Could be confusing...

### DIFF
--- a/worker/cfg/cfg.go
+++ b/worker/cfg/cfg.go
@@ -1,12 +1,9 @@
 package cfg
 
 import (
-	"context"
-	"fmt"
-	"github.com/joschahenningsen/TUM-Live/worker/pb"
+	"github.com/getsentry/sentry-go"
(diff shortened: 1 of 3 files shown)


joschahenningsen (approved):
:tada:

PR #611, opened 2022-08-01
Reduce Keyword-Extraction footprint

MatthiasReumann:
## Changes 
- Remove language attribute (No advantages - did introduce a lot of duplicates)
- Increase minimum word length
- Remove whitespaces from results


worker/worker/request_handlers.go
```diff
@@ -577,18 +577,12 @@ func extractKeywords(ctx *StreamContext) error {
 	}
 
 	// extract keywords
-	engExtractor := ocr.NewOcrExtractor(files, []string{"eng"})
+	engExtractor := ocr.NewOcrExtractor(files, []string{"eng", "deu"})
```
joschahenningsen: IMO extractor would be a better name now

MatthiasReumann: yup, definitely. 👍🏻

joschahenningsen (approved):
:tada:

PR #630, opened 2022-08-05
Don't preload stream stats

joschahenningsen:
I deployed #609 out of curiosity and found, that the stat preloading makes up for a significant amount of our request times:

image
image

Since the stats have their own rest endpoints, there is no need to preload them when fetching the course.

### DIFF
--- a/dao/streams.go
+++ b/dao/streams.go
@@ -79,7 +79,7 @@ func (d streamsDao) SaveTranscodingProgress(progress model.TranscodingProgress) 
 	return DB.Clauses(clause.OnConflict{UpdateAll: true}).Create(&progress).Error
 }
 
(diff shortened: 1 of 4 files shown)


dao/streams.go
```diff
@@ -206,16 +206,6 @@ func AddVodView(id string) error {
 	return err
 }
 
-func UpdateStream(stream model.Stream) error {
```
alexanderstephan, line 209: I think the method above `AddVodView` could also be deleted while we're at it.

alexanderstephan (approved):
Never would have thought loading stats is so heavy on resources. Nice catch! I noted down one more deprecation in the `dao`.

Code changes make sense. I can't find the per lecture stats for individual lectures anymore, so we should be fine doing it like this.

PR #633, opened 2022-08-11
Include Timestamp in URL

### LINKED ISSUE #632: Enhancement: Include Timestamp in URL
It would be nice to be able to share a URL to a stream that includes a timestamp. For example, instead of saying "go watch https://live.rbg.tum.de/w/gad/12474 starting at 14:58 :sheep:" you could just share the link https://live.rbg.tum.de/w/gad/12474?t=898. Then, playing the stream would automatically seek to the given timestamp. Adding a "share" button that (optionally) includes the timestamp would be an additional nice-to-have. (YouTube does this for example)

MatthiasReumann:
## Changes
- Closes #632
- Add `getQueryParam` function to `global.ts`
- Set the timestamp of the player for the given parameter in the `watchProgress` handler.
- Move logged-in check from .gohtml file to typescript

## Further Information
The current implementation ignores the user's video progress if query parameter `t` is set. (`if ... else`).

### DIFF
--- a/web/template/watch.gohtml
+++ b/web/template/watch.gohtml
@@ -303,8 +303,8 @@
     {{else}}
     watch.initPlayer(true, false, false, {{.IndexData.TUMLiveContext.User.GetPlaybackSpeeds.GetEnabled}}, {{$stream.LiveNow}});
     {{end}}
-    {{if and $stream.Recording .IndexData.TUMLiveContext.User}}
-    watch.watchProgress({{$stream.Model.ID}}, {{ .Progress.Progress }});
+    {{if and $stream.Recording}}
+    watch.watchProgress({{$stream.Model.ID}}, {{ .Progress.Progress }}, {{.IndexData.TUMLiveContext.User}});
     {{end}}
     {{if $stream.Silences}}
     watch.skipSilence({{$stream.GetSilencesJson}});
--- a/web/template/watch.gohtml
+++ b/web/template/watch.gohtml
@@ -300,11 +300,9 @@
 <script>
     {{if $stream.Recording}}
     watch.initPlayer(true, false, false, {{.IndexData.TUMLiveContext.User.GetPlaybackSpeeds.GetEnabled}}, {{$stream.LiveNow}}, {{$stream.GetThumbIdForSource .Version}}, {{$stream.ThumbInterval}}, {{$stream.Model.ID}});
+    watch.watchProgress({{$stream.Model.ID}}, {{ .Progress.Progress }}, {{not (empty .IndexData.TUMLiveContext.User)}});
     {{else}}
     watch.initPlayer(true, false, false, {{.IndexData.TUMLiveContext.User.GetPlaybackSpeeds.GetEnabled}}, {{$stream.LiveNow}});
-    {{end}}
-    {{if and $stream.Recording}}
-    watch.watchProgress({{$stream.Model.ID}}, {{ .Progress.Progress }}, {{.IndexData.TUMLiveContext.User}});
     {{end}}
     {{if $stream.Silences}}
     watch.skipSilence({{$stream.GetSilencesJson}});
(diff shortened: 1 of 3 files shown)


web/ts/global.ts
```diff
@@ -251,6 +251,18 @@ export type Section = {
...
+export function getQueryParam(name: string): string {
+    const query = window.location.search.substring(1); // remove '?'
+    const params = query.split("&");
+    for (let i = 0; i < params.length; i++) {
+        const pair = params[i].split("=");
+        if (pair[0] == name) {
+            return pair[1];
+        }
+    }
+    return undefined;
```
just-max: `URL` can do this directly
```suggestion
    return new URL(window.location.href).searchParams.get(name) ?? undefined;
```

web/ts/TUMLiveVjs.ts
```diff
@@ -210,23 +210,29 @@ export const watchProgress = function (streamID: number, lastProgress: number) {
...
-            player.on("canplaythrough", () => {
-                // Can be executed multiple times during playback
-                if (!iOSReady) {
-                    player.currentTime(lastProgress * duration);
-                    iOSReady = true;
-                }
+        // check if query contains parameter 't' and set timestamp accordingly
+        const queryT = getQueryParam("t");
+        if (queryT !== undefined) {
+            player.currentTime(Number(queryT));
```
just-max: This should probably include the same special handling as for iOS, which can be encapsulated in a function to be called from both places.

just-max (changes requested):
Thanks, just what I wanted :heart_eyes:

web/ts/global.ts
MatthiasReumann: Thanks, that's great!

web/template/watch.gohtml
```diff
@@ -303,8 +303,8 @@
     {{else}}
     watch.initPlayer(true, false, false, {{.IndexData.TUMLiveContext.User.GetPlaybackSpeeds.GetEnabled}}, {{$stream.LiveNow}});
     {{end}}
-    {{if and $stream.Recording .IndexData.TUMLiveContext.User}}
-    watch.watchProgress({{$stream.Model.ID}}, {{ .Progress.Progress }});
+    {{if and $stream.Recording}}
+    watch.watchProgress({{$stream.Model.ID}}, {{ .Progress.Progress }}, {{.IndexData.TUMLiveContext.User}});
```
just-max: I would suggest not sending the whole user data if we don't need it. No strong opinion though, this works fine as-is.
```suggestion
    watch.watchProgress({{$stream.Model.ID}}, {{ .Progress.Progress }}, {{not (empty .IndexData.TUMLiveContext.User)}});
```

just-max (approved):
Thanks, works nicely!

just-max (approved):
Looks good, thanks for implementing this!

alexanderstephan (approved):
Cool feature! :100: Code also looks good to me :+1:

joschahenningsen:
Amazing, now I can share https://live.rbg.tum.de/w/it-sec/7105?t=4411 :heart_eyes:

web/ts/TUMLiveVjs.ts
```diff
@@ -481,6 +488,21 @@ function debounce(func, timeout) {
...
+// iPhone/iPad need to set the progress again when they actually play the video. That's why loadedmetadata is
+// not sufficient here.
+// See https://stackoverflow.com/questions/28823567/how-to-set-currenttime-in-video-js-in-safari-for-ios.
+// Returns updated iOSReady value
+function setTimeOnIOS(iOSReady: boolean, seconds: number): boolean {
+    player.on("canplaythrough", () => {
+        // Can be executed multiple times during playback
+        if (!iOSReady) {
+            player.currentTime(seconds);
+            return true;
```
just-max, line 500: In relation to #752 : this doesn't look right. From what I can tell, the `return true;` is meant to be used to update the value of `iOSReady` in `watchProgress`. Returning it from the callback doesn't do that. Thus, `iOSReady` is always false and the player keeps settings the current time (in the case of #752 , to the end of the video). Can't test if that's actually the problem right now.

PR #636, opened 2022-08-11
Remove Course.LiveEnabled

### LINKED ISSUE #391: Course.LiveEnabled should be removed
We stream all courses, streams that should only be recorded can be streamed privately and are recorded as well. The livestream then simply serves as a monitor for the recording.

MatthiasReumann:
## Changes 
- Closes #391 
- Updates testdata
- Eliminates `LiveEnabled` from all places

## Further information 
SQL drop column command:
```sql
ALTER TABLE courses DROP COLUMN live_enabled;
```


joschahenningsen:
> `ALTER TABLE courses DROP COLUMN live_enabled;`

This could be a [migration](https://github.com/joschahenningsen/TUM-Live/blob/dev/dao/migrations/202207240.go)

MatthiasReumann:
> > `ALTER TABLE courses DROP COLUMN live_enabled;`
> 
> This could be a [migration](https://github.com/joschahenningsen/TUM-Live/blob/dev/dao/migrations/202207240.go)

I've added the migration. I hope I got the naming right 😅 How do I make sure that the migrations executes - or in other words - where do I call the migration?

joschahenningsen:
You can reference it here https://github.com/joschahenningsen/TUM-Live/blob/f80d12b754a30effed3f7bfc26533187351790d7/dao/migrator.go#L34 :)

joschahenningsen (approved):

joschahenningsen (changes requested):
I think `course.StreamEnabled` becomes `streamCtx.stream` after sending the request to the worker. 

https://github.com/joschahenningsen/TUM-Live/blob/fa9bfbf2c0a13dc9446a2325efb204d99cfea31a/worker/worker/request_handlers.go#L433

This should also be removed along with the logic for just recording without streaming.

MatthiasReumann:
> I think `course.StreamEnabled` becomes `streamCtx.stream` after sending the request to the worker.
> 
> https://github.com/joschahenningsen/TUM-Live/blob/fa9bfbf2c0a13dc9446a2325efb204d99cfea31a/worker/worker/request_handlers.go#L433
> 
> This should also be removed along with the logic for just recording without streaming.

Thanks! 

> the logic for just recording without streaming.

Thanks! How about the `PublishStream` attribute in the `StreamRequest`? I guess that should be deleted too.

joschahenningsen:
> How about the PublishStream attribute in the StreamRequest?

Yup that should be removable as well then.

Am Di., 16. Aug. 2022 um 15:35 Uhr schrieb matthias <
***@***.***>:

worker/worker/request_handlers.go
```diff
@@ -214,13 +211,8 @@ func HandleStreamRequest(request *pb.StreamRequest) {
 	regularStreams.addContext(streamCtx.streamId, streamCtx)
 
 	//only record
```
joschahenningsen:
```suggestion
```

joschahenningsen (approved):

PR #489, opened 2022-05-30
Separate link/s for data privacy statement and imprint + markdown

MatthiasReumann:
- Added `texts` endpoint
- Added Texts Admin Tab
- Some UI changes 

### UI

image

On the mobile devices this merges with the new side nav:

image

Admin Tab:

image

### INSERT Statement for About, Imprint, and Data Privacy (Needs to be inserted after deployment)
```
insert into info_pages
    (id, created_at, updated_at, name, raw_content, type)
values
    (1, NOW(), NOW(), 'Privacy', '# Datenschutzerklärung

Für die Technische Universität München ist Datenschutz ein wichtiges Anliegen. Wir möchten, dass Sie wissen,
wann wir welche Daten speichern und wie wir sie verwenden. Personenbezogene Daten werden auf diesem zentralen
Web-Server der Universität (www.tum.de) nur im technisch notwendigen Umfang erhoben. Die Datenverarbeitung
unterliegt den geltenden datenschutzrechtlichen Bestimmungen, insbesondere der Datenschutz-Grundverordnung
(DSGVO) dem Bayerischen Datenschutzgesetz (BayDSG) und dem Telemediengesetz (TMG).Nachfolgend informieren wir Sie über Art, Umfang und Zweck der Erhebung und Verwendung personenbezogener Daten.
Diese Informationen können jederzeit von unserer Webseite abgerufen werden.

## Verwendete personenbezogene Daten
Bei einer Anmeldung werden die Folegenden personenbezogenen Daten erhoben:

- Anonymer personenbezogener Identifier ("Obfuscated ID")
- LRZ ID
- Matrikelnummer
- Name
- Kursanmeldungen aus TUMOnline

Die zum Login verwendeten Daten werden ausschließlich zur identifizierung an den TUM-weiten identity Provider
gesendet und im Anschluss verworfen.

## Protokollierung
Wenn Sie diese Internetseite aufrufen, übermitteln Sie über Ihren Internetbrowser Daten an unseren Webserver.
Die folgenden Daten werden während einer laufenden Verbindung zur Kommunikation zwischen Ihrem Internetbrowser
und unserem Webserver temporär in einer Logdatei aufgezeichnet:

- IP-Adresse des anfragenden Rechners</li>
- Datum und Uhrzeit des Zugriffs</li>
- Name, URL und übertragene Datenmenge der abgerufenen Datei
- Zugriffsstatus (angeforderte Datei übertragen, nicht gefunden etc.)
[...]

PR #646, opened 2022-08-30
Remove old about page

joschahenningsen:
#489 introduced custom info pages. This about page has been unused since.


MatthiasReumann (approved):
👍🏻 🚀

PR #558, opened 2022-06-29
Enhancement: Enable Chat Individually per Stream

MW360:
closes #557 

Adds checkboxes to admin view to enable chats individually per stream.

Implemented behaviour:
* If chats are globally enabled for the course chats will be enabled based on the individual per stream settings.
* If chats are globally disabled, all chats get disabled for the course regardless of their individual settings.
* The preset isChatEnabled setting for adding new lectures is derived from the current global course wide isChatEnabled setting
* If chat is disabled the backend blocks incoming API requests regarding chat

PR #648, opened 2022-09-01
fix websocket connection when chat is disabled 

joschahenningsen:
In #558, a mechanism was introduced that aborts all websocket requests when the chat is not enabled.

Since the Websocket is also used for other stuff (e.g. capture viewer counts and broadcast them) this should not happen.


alexanderstephan (approved):
Oh no... :(

Ignoring makes sense, changes look good to me.

PR #665, opened 2022-09-12
Custom Branding

### LINKED ISSUE #607: Make Branding Customizable
If a non-TUM user wants to deploy TUM-Live, they should be able to mount their own branding config into the docker container to change page title, icons etc. 

This includes: 
- [ ] Header Icon
- [ ] Meta information (e.g. page title and description)
- [ ] Service-worker meta information
- [ ] favicon
- [ ] ?

MatthiasReumann:
# Changes 
- Update Dockerfile
- Add API Route `/logo.svg` and `/manifest.json`
- Add `branding.go` which reads a `branding.yaml` file for the customization of titles, etc.
- Update README.md

# How it works 

Mount `branding` folder. Specify the folder in `config.yaml`. The custom logo and favicon should be visible.

If a `branding.yaml` is provided at the same location as `config.yaml`, it will be used for titles, etc.

If the `branding` folder doesn't exist the default TUM-Live configuration is used.


Dockerfile
```diff
@@ -3,6 +3,11 @@ FROM node:18 as node
 WORKDIR /app
 COPY web web
 
+# Copy branding (if exists)
+COPY 'brandin[g]/manifest.json' /app/web/assets/
+COPY 'brandin[g]/favicon.ico' /app/web/assets/
+COPY 'brandin[g]/logo.svg' /app/web/assets/img/
```
joschahenningsen: Would it be possible to mount these with a volume like the branding.yaml?
End users should not be required to build the app themselves imo. 

Changes look great otherwise!

MatthiasReumann: Good idea. Will look into it!

joschahenningsen (approved):

PR #664, opened 2022-09-11
Update go, remove deprecations and house-keeping

alexanderstephan:
- Update to 1.19, including the images used in Dockerfiles
- Staticcheck runs now for every (relevant) project in this repo
- Update staticcheck (looks like this is not detected by Dependabot, we missed out on a few versions)
- `io/ioutil` is deprecated (actually since 1.16), it can be replaced with functions from `os` and `io`
- Fix some other stuff found by staticcheck

PR #677, opened 2022-09-22
remove obsolete record.go

joschahenningsen:
In #636 this file became obsolete and can now be removed.


joschahenningsen:
Seems like this will be fixed in #664

PR #681, opened 2022-09-25
Improve robustness of watched counter

### LINKED ISSUE #680: Course page broken when no videos exist.
`Uncaught ReferenceError: watchedAll is not defined` seems to be the cause here.

image

alexanderstephan:
Fixes https://github.com/joschahenningsen/TUM-Live/issues/680.

The cause here was not that in fact there are no VoDs, but rather private VoDs. By filtering these out beforehand the problem should be solved.


joschahenningsen (commented):
Amazing, thanks a ton! :heart:

joschahenningsen (approved):

PR #683, opened 2022-09-27
Also fetch students for upcoming courses

joschahenningsen:
We currently have some revision courses that are scheduled for the upcoming semester (starting October 1st.). The enrollments for these courses are currently not automatically updated. This fixes the issue by always fetching students for courses in this **and** the upcoming semester.


MatthiasReumann (approved):
LGTM 👍🏻

PR #696, opened 2022-10-02
Influxdb viewer stats

mono424:
Pushing viewer stats into the influxdb instead of mysql.
 
To test:
- setup a local influxdb
- retrieve api token from web interface (Load Data > API Token)
- add api token to `config.yaml`


tools/stats/stats.go
```diff
@@ -0,0 +1,32 @@
...
+	client    influxdb2.Client
+	liveStats api.WriteAPIBlocking
+}
+
+var Client *Stats
+
+func InitStats(client influxdb2.Client) {
+	Client = &Stats{
+		client:    client,
+		liveStats: client.WriteAPIBlocking("rbg", "live_stats"),
```
joschahenningsen: This should either be non blocking or writing the stats has to happen in a separate go routine. I prefer the first option.

api/chat.go
```diff
@@ -573,8 +574,7 @@ func CollectStats(daoWrapper dao.DaoWrapper) func() {
 			}
 			if s, err := daoWrapper.GetStreamByID(context.Background(), fmt.Sprintf("%d", sID)); err == nil {
 				if s.LiveNow { // store stats for livestreams only
-					s.Stats = append(s.Stats, stat)
-					if err := daoWrapper.AddStat(stat); err != nil {
+					if err := stats.Client.AddStreamStat(fmt.Sprintf("%d", s.CourseID), stat); err != nil {
```
joschahenningsen: Here the live viewers are counted. Can we also add a point for every vod view? (https://github.com/joschahenningsen/TUM-Live/blob/5aa33d394ed3b639e703fcdfc9eac4521bee5bd6/api/chat.go#L634)

joschahenningsen: Also the `AddVodView` dao stuff can then be removed.

joschahenningsen (changes requested):
Looks good so far! The next step would be to get the stats back from influxdb to display them (e.g. at `/admin/course/123/stats`). 
Are you familiar with the [flux](https://docs.influxdata.com/influxdb/v2.4/query-data/flux) query language? If you want I can write a few useful queries for you to use :)

mono424:
> Looks good so far! The next step would be to get the stats back from influxdb to display them (e.g. at `/admin/course/123/stats`). Are you familiar with the [flux](https://docs.influxdata.com/influxdb/v2.4/query-data/flux) query language? If you want I can write a few useful queries for you to use :)

Not yet, but I love to look into that. You would still appreciate the few useful queries 💯

joschahenningsen:
Let's start with this for testing :)

### Live viewers for a course:
```flux
from(bucket: "live_stats")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "viewers")
  |> filter(fn: (r) => r["course"] == "224")
  |> filter(fn: (r) => r["live"] == "true")
  |> group(columns: ["stream"])
  |> max()
  |> group()
  |> aggregateWindow(every: 1d, fn: median, createEmpty: false)
```

Produces [2022-10-04_09 36_influxdb_data.csv](https://github.com/joschahenningsen/TUM-Live/files/9703685/2022-10-04_09.36_influxdb_data.csv)

Which looks like this:

image


This can be fetched and cleaned up for chart.js like this:
```go
	query, err := c.QueryAPI("rbg").Query(context.Background(), "from(bucket: \"live_statistics\")"+
		"|> range(start: -1y, stop: now())"+
		"|> filter(fn: (r) => r[\"_measurement\"] == \"viewers\")"+
		"|> filter(fn: (r) => r[\"course\"] == \"224\")"+
		"|> filter(fn: (r) => r[\"live\"] == \"true\")"+
		"|> group(columns: [\"stream\"])"+
		"|> max()"+
		"|> group()"+
		"|> aggregateWindow(every: 1d, fn: median, createEmpty: false)")
	if err != nil {
		panic(err)
	}
	var xAxis = make([]string, 0)
	var yAxis = make([]int, 0)
	for query.Next() {
		xAxis = append(xAxis, query.Record().Time().Format("2006-01-02"))
		strVal, ok := query.Record().Value().(string)
		if !ok {
			yAxis = append(yAxis, 0)
			continue
		}
		n, err := strconv.Atoi(strVal)
		if err != nil {
			yAxis = append(yAxis, 0)
			continue
		}
		yAxis = append(yAxis, n)
	}
```

web/yarn.lock
```diff
@@ -0,0 +1,3310 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
joschahenningsen: can this be deleted?

.idea/runConfigurations/build_dev.xml
```diff
@@ -5,7 +5,8 @@
     <scripts>
       <script value="build-dev" />
     </scripts>
-    <node-interpreter value="project" />
+    <node-interpreter value="node" />
+    <package-manager value="yarn" />
```
joschahenningsen: this can be reverted

dao/statistics.go
```diff
@@ -1,25 +1,13 @@
 package dao
```
joschahenningsen: Can you regenerate the mock dao please?

dao/statistics.go
```diff
@@ -30,133 +18,9 @@ func NewStatisticsDao() StatisticsDao {
...
 }
 
-// AddStat adds a new statistic entry to the database
-func (d statisticsDao) AddStat(stat model.Stat) error {
-	return DB.Create(&stat).Error
-}
-
-//GetCourseNumStudents returns the number of students enrolled in the course
+// GetCourseNumStudents returns the number of students enrolled in the course
 func (d statisticsDao) GetCourseNumStudents(courseID uint) (int64, error) {
```
joschahenningsen: This could now be part of the course dao. Then we might delete this file.

joschahenningsen (changes requested):

joschahenningsen:
T'm afraid this needs to be closed. Sorry for letting this go stale, I suppose the best way to move forward here is integrate our grafana stack here at some point.

PR #710, opened 2022-10-07
Remove Google Cast toggle

alexanderstephan:
Simply removes the Google Cast setting as the feature is probably not coming any time soon.

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -23,7 +23,7 @@ func ConfigRealtimeRouter(router *gin.RouterGroup) {
 	RegisterRealtimeChatChannel()
 }
 
-//ConfigGinRouter for non ws endpoints
+// ConfigGinRouter for non ws endpoints
 func ConfigGinRouter(router *gin.Engine) {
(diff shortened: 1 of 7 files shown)


joschahenningsen (changes requested):
Agree 👍 can you also remove the route for setting this, the setting type and write a quick migration that drops all settings with this type so they don't cause problems later?

alexanderstephan:
Thanks for the hint, I completely missed that part. :pray:  Do I have to generate anything in this case?

joschahenningsen:
You can write a migration by hand, have a look at https://github.com/joschahenningsen/TUM-Live/blob/dev/dao/migrations/202201280.go

alexanderstephan:
I used an unscoped delete to keep the DB clean, I hope it's not overkill.

joschahenningsen:
> I used an unscoped delete to keep the DB clean, I hope it's not overkill.

Nope thats perfect here!

joschahenningsen (approved):

PR #714, opened 2022-10-10
Small UI bug fixes

MatthiasReumann:
## Changes 
- Fix search input field (was a bit misaligned after #697)
- Don't show Poll UI when not logged in (Don't send `http.StatusOK` on error)

### DIFF
--- a/api/chat.go
+++ b/api/chat.go
@@ -470,7 +470,7 @@ func (r chatRoutes) getActivePoll(c *gin.Context) {
 	}
 	poll, err := r.ChatDao.GetActivePoll(tumLiveContext.Stream.ID)
 	if err != nil && err == gorm.ErrRecordNotFound {
-		c.JSON(http.StatusFound, nil)
+		c.JSON(http.StatusNotFound, nil)
 		return
 	}
 	if err != nil {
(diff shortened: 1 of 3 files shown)


api/chat.go
```diff
@@ -463,14 +463,14 @@ func (r chatRoutes) getActivePoll(c *gin.Context) {
...
-			Status:        http.StatusOK,
+			Status:        http.StatusBadRequest,
 			CustomMessage: "not logged in",
 		})
 		return
 	}
 	poll, err := r.ChatDao.GetActivePoll(tumLiveContext.Stream.ID)
 	if err != nil && err == gorm.ErrRecordNotFound {
-		c.JSON(http.StatusOK, nil)
+		c.JSON(http.StatusFound, nil)
```
joschahenningsen: StatusFound is intended for redirects. Is this intended?

joschahenningsen (changes requested):

api/chat.go
MatthiasReumann: Nope. Should be NotFound. Thanks for pointing out!

joschahenningsen (approved):
Thanks!

PR #718, opened 2022-10-11
Avoid inits

### LINKED ISSUE #685: Remove side effects from code caused by `init()`
Our code utilizes the init() function go automatically runs at package initialization. Most people recommend avoiding this https://github.com/uber-go/guide/blob/master/style.md#avoid-init as it makes the code unclear and less testable. For instance: The worker could try communicate with TUM-Live when tests are run because it's initialized using `init()`.

> Avoid depending on the ordering or side-effects of other init() functions. While init() ordering is well-known, code can change, and thus relationships between init() functions can make code brittle and error-prone.

This is an issue as well in our dao packages. When the daos are initialized the DB variable might not be set yet, causing calls to `somedao.db` to fail.

MatthiasReumann:
## Changes 
- Closes #685
- [Worker] `cfg.Init()` --> `cfg.SetConfig()` called in `cmd/worker.go`
- [TUM-Live] `config.Init()` --> `config.LoadConfig()` called in `cmd/tumlive.go`
- [TUM-Live] `api.Init()` --> `api.ServeWorkerGRPC()` called in `cmd/tumlive.go`


alexanderstephan (commented):
Just had a look here out interest. The changes make sense to me, but I saw there are some more `inits()` not addressed by this PR. Are these OK? I think they could be also potentially removed.

- This looks a bit problematic to me: https://github.com/joschahenningsen/TUM-Live/blob/095ca48c197349fcbcc06f8213d32912186d242e/tools/branding.go#L28
- This one could be fine: https://github.com/joschahenningsen/TUM-Live/blob/815492a680744efc568197e6cb626268313d0a69/tools/realtime/channel_store.go#L13

MatthiasReumann:
> Just had a look here out interest. The changes make sense to me, but I saw there are some more `inits()` not addressed by this PR. Are these OK? I think they could be also potentially removed.
> 
> * This looks a bit problematic to me: https://github.com/joschahenningsen/TUM-Live/blob/095ca48c197349fcbcc06f8213d32912186d242e/tools/branding.go#L28
> * This one could be fine: https://github.com/joschahenningsen/TUM-Live/blob/815492a680744efc568197e6cb626268313d0a69/tools/realtime/channel_store.go#L13

Ah these must have been added after the merge of some later pull requests. At least the branding one is new. 
I will remove those two too. 👍🏻

MatthiasReumann:
> Just had a look here out interest. The changes make sense to me, but I saw there are some more `inits()` not addressed by this PR. Are these OK? I think they could be also potentially removed.
> 
> * This looks a bit problematic to me: https://github.com/joschahenningsen/TUM-Live/blob/095ca48c197349fcbcc06f8213d32912186d242e/tools/branding.go#L28
> * This one could be fine: https://github.com/joschahenningsen/TUM-Live/blob/815492a680744efc568197e6cb626268313d0a69/tools/realtime/channel_store.go#L13

Removed the branding init function. The second one doesn't is fine, yes.

I've also added a list of `initializers` and the function `initAll` in `cmd/tumlive.go`. `initAll` basically iterates the global variable `initializers` calls the function at the beginning of execution.

alexanderstephan (approved):
The iterator is a really good idea! Looks all good from side. I would just want to wait if this is what @joschahenningsen had in mind.

PR #719, opened 2022-10-12
Don't try streaming lectures of deleted courses

### LINKED ISSUE #390: streams with deleted courses cause trouble
When importing courses as 'opt in', the courses field `deleted_at` is set until someone requests streaming the course. The lectures don't have this field set. When automatically starting streams or viewing the calendar those streams are selected even though there is no course associated with them. This causes trouble for the workers and is confusing. Ideally, the streams should also be deleted until opting in.

joschahenningsen:
Since we let people opt in to streaming, some courses exist where `course.deleted_at` is set (until opting in) but the streams are not deleted. Currently we would try to stream such lectures. 

fixes #390


alexanderstephan (approved):
Seems good. 👍  We should not forget to update the course deletion function at some point.

PR #724, opened 2022-10-17
fix wrong calendar events

joschahenningsen:
We have two issues with our calendar:

- events of deleted courses are still shown. This is mainly a problem with opt-in courses where the streams exist but the course is marked as deleted.
- streams of courses that don't have lecturers are not shown to admins.

This PR fixes both issues in the SQL query


MatthiasReumann (approved):
LGTM 👍🏻

PR #728, opened 2022-10-19
Fix chat misplacement

MatthiasReumann:
### Motivation and Context
`<a>`-tags are not wrapped properly in the chat window. Therefore the chat becomes wider and moves below the player.
Can be seen at [Einführung in die Rechnerarchitektur](https://live.rbg.tum.de/w/WiSe22ERA/22918).

### Description
Since Tailwind-CSS doesn't provide the class, add `.overflow-wrap-anywhere` to the `main.css` file and apply it to the message container.

### Steps for Testing
1. Log in
2. Navigate to a Livestream that has the chat enabled
3. Post one of the following links:
    1. https://chat.whatsapp.com/DKIH69Z0RkTLSv8neUb5GP
    2. https://zulip.in.tum.de/#narrow/stream/1216-ERA---Allgemein/topic/Abonement.20der.20ERA.20Streams/near/694571
    3. Or any other long link
4. The chat should stay at its place 

### Screenshots
image

### DIFF
--- a/web/assets/css/main.css
+++ b/web/assets/css/main.css
@@ -406,3 +406,7 @@ a.fc-event, a.fc-event:hover {
 .tum-live-markdown a{
     border-bottom: 1px solid black;
 }
+
+.overflow-wrap-anywhere {
+    overflow-wrap: anywhere;
+}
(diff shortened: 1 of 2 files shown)


joschahenningsen (approved):
Neat catch!

PR #730, opened 2022-10-19
log course opt-in in audits

joschahenningsen:
### Motivation and Context

When a lecturer opts in we currently don't know when this happened or who it was. This would be very helpful.

### Description
This PR adds audit creation to the opt in endpoint.

### Steps for Testing

Code review should be enough.

Prerequisites:
- a course that is not yet opted in
- a admin user who can view audits

1. open the opt in link for the course
2. click `enable streaming`
3. log in as admin (if not yet logged in)
4. find action in audits

### Screenshots

n/a


AndreasPaul (approved):
:+1:

PR #735, opened 2022-10-20
allow deselecting lecture halls in calendar

### LINKED ISSUE #732: Filter calendar in admin area for lecture halls
Provide an easy way to find the next lecture in a certain lecture hall that may have technical issues for example

Co 14:33 Die SMP war auf "Pult 1 DMC 3" eingestellt,  hab auf "Content" umgestellt, jetzt geht's
Joscha Henningsen 14:34 Mussen wir das dann wieder zurück stellen? Bei GDB und GBS ging es vorher
Co  14:36 Gute Frage, weiss nicht wie's dazu gekommen ist. Da muessen wir im 0001 schauen ob das oeffter passiert
       14:37 🙈 und schon wieder faellt mir ein Feature ein ... im Kalender waer's cool, wenn man nach Hoersaelen filtern koennte

joschahenningsen:
### Motivation and Context

To have a quick glance about the events that are in a certain lecture hall it would be helpful if we were able to filter the events in the calendar.
fixes #732 

### Description

This PR adds a checkbox for each lecture hall below the calendar. If a checkbox is checked we pass the id to the api which then includes that lecture hall in the result. The changes (logic wise) are done in SQL.

### Steps for Testing

Prerequisites:
- Lecturers
- An admin

1. generate events for different courses in different lecture halls.
2. view the schedule as lectuer and verify that 
  2.1 filtering by lecture hall works.
  2.2 The lecturer only sees their events.  
4. test that filtering also works for admins.

### DIFF
--- a/api/lecture_halls.go
+++ b/api/lecture_halls.go
@@ -37,7 +37,7 @@ func configGinLectureHallApiRouter(router *gin.Engine, daoWrapper dao.DaoWrapper
 	adminsOfCourse.Use(tools.AdminOfCourse)
 	adminsOfCourse.POST("/switchPreset/:lectureHallID/:presetID/:streamID", routes.switchPreset)
 
-	router.GET("/api/hall/all.ics", routes.lectureHallIcal)
+	router.GET("/api/schedule.ics", routes.lectureHallIcal)
 }
 
 type lectureHallRoutes struct {
--- a/api/lecture_halls.go
+++ b/api/lecture_halls.go
@@ -12,6 +12,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"net/http"
 	"strconv"
+	"strings"
 	"text/template"
 	"time"
 )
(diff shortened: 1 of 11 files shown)


alexanderstephan (changes requested):
Cool feature! I tested all the listed steps and everything works flawlessly. :+1: 

I noticed only two small potential problems.
- Checkbox text is not readable presumable a missing CSS class. Or maybe there is a problem with my setup right now. :thinking: 
grafik

- After updating the filters there are always to identical `GET` requests. Is this intended? I noticed small very short flickering, maybe this is related? I did not have too much time to look into it, but want to have it mentioned here. :stuck_out_tongue:
grafik

joschahenningsen:
I pushed a few tiny fixes, this is what it looks like on my end now: 

https://user-images.githubusercontent.com/44805696/197293618-63b5ddf1-c97f-495c-8634-7036d2e81fbf.mp4

I couldn't find out why your url didn't change, can you verify the latest commit?

joschahenningsen:
> I noticed small very short flickering

I think the fullCalendar needs some time to render the events unfortunately :/

alexanderstephan:
> I think the fullCalendar needs some time to render the events unfortunately :/

Hm, not really a problem, but might be worth to look whether there is an issue for this for fullCalendar.

> I pushed a few tiny fixes, this is what it looks like on my end now:

Looks the same for me, looks good now!

> I couldn't find out why your url didn't change, can you verify the latest commit?

I mean the URL did change after editing, but the request is sent twice. I recorded another gif to make sure we mean the same thing. Do you also have this?

Peek 2022-10-22 00-00

joschahenningsen:
@alexanderstephan nice catch! Apparently `AddEventSource` already triggers `RefetchEventSources` internally so the call was redundant.

api/lecture_halls.go
```diff
@@ -209,13 +210,28 @@ func (r lectureHallRoutes) lectureHallIcal(c *gin.Context) {
 		})
 		return
 	}
+	err = c.Request.ParseForm()
+	if err != nil {
+		_ = c.Error(tools.RequestError{Status: http.StatusBadRequest, CustomMessage: "Bad Request", Err: err})
+		return
+	}
+	lectureHallsStr := strings.Split(c.Request.Form.Get("lecturehalls"), ",")
```
alexanderstephan, line 218: One last thing I noticed: This does not work if all lecture halls are deselected, resulting in the following error:
grafik

alexanderstephan (changes requested):
Looks good! I just found one more minor bug.

joschahenningsen:
Fixed that as well, thanks :) @alexanderstephan

alexanderstephan (approved):
Works perfectly! :) 🎉

AndreasPaul:
A de-select all feature would be usefull

PR #736, opened 2022-10-21
fix stream stop ui

joschahenningsen:
### Motivation and Context

Currently the stream stop ui (popup) is hidden behind the player. 

### Description

This issue is fixed by setting the z index of the popup higher than the one of the player.

### Steps for Testing

Prerequisites:
- 1 Lecturer
- 1 Livestream

1. Navigate to a Livestream
2. Click stop button and verify that the ui is visible.


joschahenningsen:
The failing test will be fixed in #735

alexanderstephan (approved):
Works as expected now!

PR #743, opened 2022-10-22
Add additional Makefile targets and phony targets

alexanderstephan:
### Motivation and Context
As I try to rely on a terminal based workflow, I found that the current Makefiles do not support everything that is needed for daily development. Therefore I added some more targets which I think are useful.

I saw that no `.PHONY` is used which is kinda bad practice as we don't really name the targets after file names, see https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html.

### Steps for Testing

Execute all the targets locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+.PHONY: all
 all: npm_dependencies go_dependencies bundle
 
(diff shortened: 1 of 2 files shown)


joschahenningsen (approved):
Very nice, thanks!

PR #744, opened 2022-10-23
fix ui of lecture hall settings

joschahenningsen:
### Motivation and Context

In #713 an issue was introduced that made the lecture hall settings section overflow if there are too many presets.

### Description

This PR removes the class that introduces `display: grid` which doesn't work here. 
Also the `select` for the source settings is now styled properly.

### Steps for Testing

Have a look at the Lecture hall settings and verify that it looks alright with a few presets.

### Screenshots

Before: 
image

After: 
image


alexanderstephan (approved):
Nice catch, totally forgot about this...  
The fix works as expected! :100:

PR #745, opened 2022-10-24
Bookmarks UI

MatthiasReumann:
### Motivation and Context
Bookmarks are a cool way to save _"klausurrelevantes"_. If many students use the bookmark feature we might even be able to generate a video section recommendations for the course admins from it.

### Description
- Adds bookmark UI to watch page

### Steps for Testing
1. Log in (Bookmarks should not be visible when not logged in!)
2. Navigate to a Livestream
3. Create a bookmark 
4. Edit a bookmark
5. Delete bookmarks by clicking on the `EDIT` button

### Screenshots

https://user-images.githubusercontent.com/29633518/201600630-0d89e363-78c1-49f0-ab64-6a09c0c7b4ef.mov


web/ts/TUMLiveVjs.ts
```diff
@@ -485,6 +485,14 @@ export function attachCurrentTimeEvent(videoSection: VideoSections) {
     });
 }
 
+export function currentTimeToHMS() {
```
mono424, line 484: I don't really understand why we are not taking just a timestamp and safe that, am I missing something?

mono424 (commented):
Looks dope and extremely useful! 🚀
Just wondering if a timestamp would be easier to handle and more dynamic than the three time attributes 👍

web/ts/TUMLiveVjs.ts
MatthiasReumann: When I first implemented video sections I started with timestamps but then I transitioned into the 3-tuple of `(h,m,s)` because I ran into all sort of problems. For one, it's easier to sort numbers. 😅

MatthiasReumann:
> Looks dope and extremely useful! 🚀
> Just wondering if a timestamp would be easier to handle and more dynamic than the three time attributes 👍

What's your take on the UI? I think it could also make sense to move the dialog and the list of bookmarks to the right hand side of the player (just like the chat). The current "description-box" does only contain content created by the lecturers themselves - Bookmarks change that - and I'm not sure if I'm a fan of it.

MatthiasReumann:
⚠️ Since I want to try some things I will put this PR back to [Draft] Status.

mono424:
I really love it! The only thing I think is a little annoying (maybe subjective tho) is that after adding a bookmark you are not directly on the bookmark overview but you have to click the button on the top right... On my opinion you can just get thrown back into the list and maybe have a small overlay if you want to add additional feedback :)

mono424 (approved):
Code looks good! Tested locally everything worked as expected! Awesome feature! 👍

MatthiasReumann:
> I really love it! The only thing I think is a little annoying (maybe subjective tho) is that after adding a bookmark you are not directly on the bookmark overview but you have to click the button on the top right... On my opinion you can just get thrown back into the list and maybe have a small overlay if you want to add additional feedback :)

I've removed the "success" message so that it displays the bookmark list immediately. Makes more sense, I think. 🤔 Thanks!

mono424:
Now works as expected! Very nice! 👍

alexanderstephan (changes requested):
Looks great! I only have some minor suggestions.

web/template/partial/stream/bookmarks.gohtml
```diff
@@ -0,0 +1,153 @@
...
+                           placeholder="" required
+                           class="rounded px-4 py-3 mt-1 text-4 focus:outline-none border-0 bg-gray-50 w-full dark:bg-gray-600"/>
+                    <span class="px-2 my-auto font-semibold text-5">:</span>
+                    <input x-model="bookmark.request.Minutes"
+                           type="number" name="bookmark-timestamp-minutes"
+                           id="bookmark-timestamp-minutes"
+                           placeholder="" required
+                           class="rounded px-4 py-3 mt-1 text-4 focus:outline-none border-0 bg-gray-50 w-full dark:bg-gray-600"/>
+                    <span class="px-2 my-auto font-semibold text-5">:</span>
+                    <input x-model="bookmark.request.Seconds"
```
alexanderstephan, line 67: As far as I can tell, there is no (client-side) validation here, see e.g.
grafik
I think this is a good safety measure.

web/template/partial/stream/bookmarks.gohtml
```diff
@@ -0,0 +1,153 @@
+{{/* 1st parameter: 'streamID' */}}
+{{define "bookmarks-modal"}}
+    <div x-data="{showAddMenu: false}"
+         class="relative h-full border rounded-lg dark:border-gray-600">
```
alexanderstephan, line 4: I think the modal looks great! However, when I am viewing it on mobile, it takes quite a large amount of space. I think it would be better to collapse it, e.g. adjusting the height to the height used by the bookmarks.
grafik

web/template/partial/stream/bookmarks.gohtml
MatthiasReumann: Definitely. Great catch!

web/template/partial/stream/bookmarks.gohtml
MatthiasReumann: As I will work on the mobile version as a whole soon, I think the best solution for now is a dialog:
image

alexanderstephan: Looks great!

alexanderstephan (approved):
💯

PR #746, opened 2022-10-24
`Admin`: Impersonate users

joschahenningsen:
### Motivation and Context

Sometimes we get issue reports that are very hard to reproduce and only happen to some users. It'd help to log in as the user reporting the issue.

### Description

This PR adds user impersonation for admins. It works by simply starting a session for the required user after authorizing as an administrator.

### Steps for Testing

Prerequisites:
- 1 Lecturer
- 1 Student
- 1 Admin

1. Log in as the admin
2. Navigate to user management
3. Try starting a session as a lecturer or student by clicking the `impersonate` button.


AndreasPaul (approved):
Very cool and usefull feature!

PR #751, opened 2022-10-27
remove pause feature

joschahenningsen:
### Motivation and Context

This feature is too specific to our lecture halls and therefore only caused trouble without bringing any real benefit. People can just end their streams. 

### Description

This PR removes all code related to the pause feature.

### Steps for Testing

1. Check that nothing broke the UI?


MatthiasReumann (approved):
LGTM 👍🏻 

One thing I noticed is that the column (obviously) still exists in the `tum-live-starter.sql` file. Do you know if that will cause any problems or does gorm simply ignore (drop?) the missing column?

PR #753, opened 2022-10-28
Add KaTeX

### LINKED ISSUE #679: Allow LaTeX snippets in chat
Unsure if this is really something we **should** implement depending on the complexity of a potential solution, but messages like this:

image

Would look much better if we allowed creating latex snippets. IMO this would require a more sophisticated chat input which I'm not sure we actually want.

alexanderstephan:
### Motivation and Context

- Fixes https://github.com/joschahenningsen/TUM-Live/issues/679. 

### Description

- Adds [KaTex](https://katex.org/) as a dependency
- Uses the [autorender extension]( https://katex.org/docs/autorender.html) to LaTex Code within chat messages.
- Adds [copy-tex](https://github.com/KaTeX/KaTeX/tree/main/contrib/copy-tex) to make rendered code copyable.
### Notes
- I tried to make it as performant as possibly by only including dependencies when needed. Furthermore, KaTeX code is only searched in the most elementary `<span>` that contains the chat message.
- I left full width equations enabled as they look surprisingly good out of the box.

Maybe this could be worth a server notification when this is deployed? I think otherwise people probably won't notice it.

### Steps for Testing

1. Log in.
2. Navigate to a livestream with chat enabled.
3. Post random equations such as `$\sum_{i=1}^{10} t_i$`.
4. Try to copy the rendered math and try to paste it somewhere. Check if it's your input.
5. Check if everything renders nicely without any noticeable delays or errors.

### Screenshots

grafik


web/template/partial/stream/chat/chat_message.gohtml
```diff
@@ -27,7 +27,16 @@
                     lg:hover:bg-gray-100 lg:dark:hover:bg-gray-700
                 {{end}}">
                 <div class="flex justify-between">
-                    <span class="chatMsg w-5/6 overflow-wrap-anywhere my-auto" x-html="m.message"></span>
+                    <span class="chatMsg w-5/6 overflow-wrap-anywhere my-auto" x-init="$nextTick(() => { renderMathInElement($el, {
```
MatthiasReumann, line 30: Instead of doing everything inline in the html file you could utilize the preprocessors of the `Chat` object like so:

```typescript 
import katex from "katex";
...
 preprocessors: ((m: ChatMessage) => ChatMessage)[] = [
        ...
        (m: ChatMessage) => {
            const msg = removeDelimiters(m.message); 
            m.message = katex.renderToString(msg, {throwOnError: false});
            return m;
        },
        ...
    ];
...
```
Then the auto-render dependency isn't necessary anymore. 

However, there will be no delimiters anymore. For example instead of `$\frac{1}{2}$`,  `\frac{1}{2}`  will suffice.

To make copy-tex work again a `removeDelimiters` function which extracts the content in between two delimiter symbols (e.g. `$<math>$`) would be needed.

MatthiasReumann: Instead of doing everything inline in the html file you could utilize the preprocessors of the `Chat` object like so:

```typescript 
import katex from "katex";
...
 preprocessors: ((m: ChatMessage) => ChatMessage)[] = [
        ...
        (m: ChatMessage) => {
            const msg = removeDelimiters(m.message); 
            m.message = katex.renderToString(msg, {throwOnError: false});
            return m;
        },
        ...
    ];
...
```
Then the auto-render dependency isn't necessary anymore. 

However, there will be no delimiters anymore. For example instead of `$\frac{1}{2}$`,  `\frac{1}{2}`  will suffice.

To make copy-tex work again a `removeDelimiters` function which extracts the content in between two delimiter symbols (e.g. `$<math>$`) would be needed.

MatthiasReumann (commented):
Looks great. 🌟 

I left one remark regarding the javascript code in the html file.

web/template/partial/stream/chat/chat_message.gohtml
alexanderstephan: Hm, I actually think that actually would be a better solution for the chat, thanks. 👍

The reason I went the the autorender extension is that it would be easily extensible to different parts of the site. E.g. using it in other places where markdown is used. I need to take a look whether the preprocessor logic also can be applied there, then I see no problems with this approach.

alexanderstephan: Hm, I actually think that actually would be a better solution for the chat, thanks. 👍

The reason I went the the autorender extension is that it would be easily extensible to different parts of the site. E.g. using it in other places where markdown is used. I need to take a look whether the preprocessor logic also can be applied there, then I see no problems with this approach.

MatthiasReumann (approved):
As discussed last Wednesday, this PR can be merged without the preprocessor changes. 🚀

PR #755, opened 2022-10-30
Feature/734 fix approve message bug

### LINKED ISSUE #734: Approving chat messages removes the complete chat for a few seconds
**Describe the bug**
As an admin, whenever I approve a chat message during a livestream, all chat messages disappear for a few seconds. They usually reappear after ½ to 10 seconds, but sometimes they never reappear again until the site is reloaded (in one case, new chat messages appeared but all other messages remained lost until reloaded).
Disturbingly, the messages also were lost for all other viewers, with them reappearing at a different times for everyone.

**To Reproduce**
1. Create and start a livestream
2. Enter a chat message as a regular user (so that it isn't approved automatically)
3. Approve message as an admin

 - OS: Happend for Windows (10), Ubuntu (20.04) and MacOS (version unknown)
 - Browser: all Firefox (mine is 105.0.1)

mono424:
Fixes #734


web/ts/chat/Chat.ts
```diff
@@ -340,6 +344,30 @@ export class Chat {
...
+            if (createdAt === newMessageCreatedAt) {
+                const newRenderVersion = this.messages[i].renderVersion + 1;
+                this.messages.splice(i, 1, { ...m, renderVersion: newRenderVersion });
+                break;
+            } else if (createdAt > newMessageCreatedAt) {
+                this.messages.splice(i, 0, m);
+                break;
+            }
+        }
+        console.log(this.messages);
```
alexanderstephan:
```suggestion
```

dao/chat.go
```diff
@@ -117,17 +108,7 @@ func (d chatDao) GetAllChats(userID uint, streamID uint) ([]model.Chat, error) {
...
-			if chats[i].UserLikes[j].ID == userID {
-				chats[i].Liked = true
-				break
-			}
-		}
-		chats[i].AddressedToIds = []uint{}
-		for _, user := range chats[i].AddressedToUsers {
-			chats[i].AddressedToIds = append(chats[i].AddressedToIds, user.ID)
-		}
+		normalizeChat(&chats[i], userID)
```
alexanderstephan: Nice refactoring! :+1:

dao/chat.go
```diff
@@ -199,3 +180,30 @@ func (d chatDao) CloseActivePoll(streamID uint) error {
...
+	var chat *model.Chat
+
+	err := d.db.Preload("Replies").Preload("UserLikes").Preload("AddressedToUsers").Find(&chat, "id = ?", id).Error
+	if err != nil {
+		return chat, err
+	}
+
+	normalizeChat(chat, userID)
+	return chat, nil
+}
```
alexanderstephan:
```suggestion
func (d chatDao) GetChat(id uint, userID uint) (*model.Chat, error) {
	var chat model.Chat

	err := d.db.Preload("Replies").Preload("UserLikes").Preload("AddressedToUsers").Find(&chat, "id = ?", id).Error
	if err != nil {
		return &chat, err
	}

	normalizeChat(&chat, userID)
	return &chat, nil
}
```

Maybe this is more idiomatic? Using the `&` operator on a pointer looks a bit weird to me.

api/chat.go
```diff
@@ -280,12 +280,20 @@ func (r chatRoutes) handleApprove(ctx tools.TUMLiveContext, msg []byte) {
...
 		return
 	}
+
 	err = r.ChatDao.ApproveChat(req.Id)
 	if err != nil {
 		log.WithError(err).Error("could not approve chat")
+		return
+	}
+
+	chat, err := r.ChatDao.GetChat(req.Id, 0)
```
alexanderstephan, line 294: I am not too familiar with the chat. Maybe this could use a comment to explain why the `userID` has to be `0` or, alternatively, a constant with a descriptive name.

alexanderstephan (changes requested):
Thanks for fixing this! :tada: The changes make sense to me and it works as expected when I test it locally with a relatively small amount of users.

I added some comments about minor stuff. However, it would be good if there is another reviewer, who is more familiar with the chat.

dao/chat.go
mono424: Oh I see its double, I am pretty sure I can just leave out the `&`. Good catch! 👍

api/chat.go
mono424: This actually quite interesting. @joschahenningsen do you have suggestion here, because we use the userId for `normalizeChat`. Its needed to set a flag `Liked` on the chat message (not sure why this is not handled on frontend). At our particular case the messages are approved so users have not liked it anyways, that means it should work with just settings the userId to 0, but its kind of ugly. Any good ideas to make it cleaner?

dao/chat.go
```diff
@@ -199,3 +180,30 @@ func (d chatDao) CloseActivePoll(streamID uint) error {
...
+	err := d.db.Preload("Replies").Preload("UserLikes").Preload("AddressedToUsers").Find(&chat, "id = ?", id).Error
+	if err != nil {
+		return &chat, err
+	}
+
+	normalizeChat(&chat, userID)
+	return &chat, nil
+}
+
+func normalizeChat(chat *model.Chat, userID uint) {
```
joschahenningsen: Let's call this something like `prepareChat` and add a descriptive comment what it does to clear things up.

```suggestion
// prepareChat sets Liked to true if the user with userID liked the message and and adds the ids of the addressed users to it for further usage in the fronted. 
func prepareChat(chat *model.Chat, userID uint) {
```

dao/chat.go
```diff
@@ -199,3 +180,30 @@ func (d chatDao) CloseActivePoll(streamID uint) error {
...
+// GetChat returns a chat message with the given id, uses the userId to normalize the chat.
+func (d chatDao) GetChat(id uint, userID uint) (*model.Chat, error) {
+	var chat model.Chat
+
+	err := d.db.Preload("Replies").Preload("UserLikes").Preload("AddressedToUsers").Find(&chat, "id = ?", id).Error
+	if err != nil {
+		return &chat, err
+	}
+
+	normalizeChat(&chat, userID)
```
joschahenningsen:
```suggestion
	prepareChat(&chat, userID)
```

dao/chat.go
```diff
@@ -199,3 +180,30 @@ func (d chatDao) CloseActivePoll(streamID uint) error {
 func (d chatDao) GetChatsByUser(userID uint) (chats []model.Chat, err error) {
 	return chats, d.db.Find(&chats, "user_id = ?", userID).Error
 }
+
+// GetChat returns a chat message with the given id, uses the userId to normalize the chat.
```
joschahenningsen:
```suggestion
// GetChat returns a chat message with the given id, uses the userId to add user specific status the chat like the liked status.
```

web/ts/chat/Chat.ts
```diff
@@ -340,6 +344,29 @@ export class Chat {
 
     isMessageToBeFocused = (index: number) => this.messages[index].ID === this.focusedMessageId;
 
+    patchMessage(m: ChatMessage): void {
```
joschahenningsen, line 348:
```suggestion
    // patchMessage adds the message to the list of messages at the position it should appear in based on the send time.
    patchMessage(m: ChatMessage): void {
```

web/ts/chat/Chat.ts
```diff
@@ -379,6 +406,7 @@ type ChatMessage = {
     visible: true;
     deleted: boolean;
     isGrayedOut: boolean;
+    renderVersion: number;
```
joschahenningsen, line 410: what exactly is a renderVersion? I can only find it being `0` here, could we use a boolean?

api/chat.go
joschahenningsen: Adding a comment that explains what these values do and why a zero is fine to use here would be nice, looks good otherwise.

joschahenningsen (changes requested):

web/ts/chat/Chat.ts
mono424: The issue is that alpine.js was not rerendering the messages for admin layout as the key was not changing after a patch. To indicate that the there is a different version of the message now, I increment the renderVersion if a message gets patched, which already exists. So theoretically later on it would support multiple patches of a message with corresponding layout update. A boolean could work too, by toggling it between true and false but this would be difficult to understand I think.

alexanderstephan (approved):
Looks good to me! 👍 Looking forward to test this out. 😁

joschahenningsen (approved):

PR #756, opened 2022-10-31
fix downloads for students

joschahenningsen:
### Motivation and Context

Students can't download lectures currently. Also, the current code is very hard to read. 

### Description

This PR should make the function more understandable and allows students to download their lectures.

### Steps for Testing

Prerequisites:
- Lecturer
- Students
- Admin
- 1 VoD

1. Test downloading a video that has downloads enabled with all users and visibilities. It should work whenever a user is allowed to see the course.
2. Test downloading a video that has downloads disabled with all users and visibilities. It should only work if the user is allowed to manage the course.

### DIFF
--- a/api/download.go
+++ b/api/download.go
@@ -22,6 +22,11 @@ type downloadRoutes struct {
 	dao.DaoWrapper
 }
 
+var dlErr = tools.RequestError{
+	Status:        http.StatusForbidden,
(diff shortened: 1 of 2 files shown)


MatthiasReumann (approved):
Thanks for cleaning up my mess 😅 

Works as expected. 💯 

btw: the new PR-template was a great idea. It makes testing much easier!

Issue #742, opened 2022-10-21
Rename video sections


alexanderstephan:
**Is your feature request related to a problem? Please describe.**
When creating video sections, I sometimes change the name of the section e.g. when I see that the section also covers another aspect.

**Describe the solution you'd like**
Have a small pen icon or something like that to edit at least. Even better would be if you could adjust the time as well.

**Describe alternatives you've considered**
Deleting it again is of course an option, but if we want more people to use this feature it has to be as convenient as possible.

PR #757, opened 2022-10-31
Edit Video Sections

MatthiasReumann:
### Motivation and Context
Closes Issue: #742 

### Description
- Added put request to video section endpoint.
- Updated the video section admin UI.
- Removed the `VideoSectionClient` class since it is a burden to work with it.

If the timestamp of a video section is changed a new thumbnail will not be generated.

### Steps for Testing
1. Log in as admin
2. Course Edit > Edit Course
3. Update video section name and timestamp
4. Head to the watch page and check if everything works as expected. 

### Screenshots
<div style="display: flex">
Screenshot 2022-10-31 at 07 52 29

Screenshot 2022-10-31 at 07 52 19
</div>


web/ts/video-sections.ts
```diff
@@ -189,47 +147,129 @@ export class VideoSectionsAdmin {
...
+            await this.fetch(); // load sections again to avoid js-sorting
             this.newSections = [];
         });
         this.unsavedChanges = false;
     }
 
     removeExistingSection(id: number) {
-        this.client.delete(id).then(async () => {
-            await this.fetch();
+        console.log("delete: " + id);
```
alexanderstephan:
```suggestion
```

web/ts/global.ts
```diff
@@ -4,6 +4,10 @@ export * from "./notifications";
 export * from "./user-settings";
 export * from "./start-page";
 
+export async function getData(url = "") {
+    return await fetch(url, { method: "GET" });
```
alexanderstephan: It's technically not neeed as far as I know, but maybe we could also specify the content type just to be more explicit and consistent with `putData`.

alexanderstephan (changes requested):
Really nice work, this is exactly what I had in mind! 👍 

Just some small questions/suggestions. I also noticed that the field to enter the name of lecture got a bit smaller, which is possibly related to this PR. 

grafik

Comparision how it looks on `dev` for me:
grafik

web/ts/video-sections.ts
MatthiasReumann: Removed 👍🏻

web/ts/global.ts
MatthiasReumann: As far as I know is Content-Type not necessary for GET requests since the request body is empty anyways, however, I removed the unnecessary options object ~~`{ method: "GET" }`~~.

MatthiasReumann:
> Really nice work, this is exactly what I had in mind! 👍
> 
> Just some small questions/suggestions. I also noticed that the field to enter the name of lecture got a bit smaller, which is possibly related to this PR.
> 
> grafik
> 
> Comparision how it looks on `dev` for me: grafik

Interestingly this bug only appears in Firefox. Somehow `w-fit` or `width: fit-content;` doesn't work there. 🧐

web/ts/global.ts
alexanderstephan: Yeah, `GET` seems to be default, I also just checked in the console. :+1:

alexanderstephan:
> Interestingly this bug only appears in Firefox. Somehow w-fit or width: fit-content; doesn't work there. monocle_face

Hm, that's weird. :thinking: As long as it works now... that's kinda web development in a nutshell. :joy:

alexanderstephan (approved):
Looks all good now!  :tada: However, I am not sure if you still want to add something since you didn't request a review.

MatthiasReumann:
> Looks all good now! 🎉 However, I am not sure if you still want to add something since you didn't request a review.

All good - mergeable on my side!

PR #758, opened 2022-11-01
Fix description markdown

### LINKED ISSUE #729: Description markdown not rendered properly
**Describe the bug**
The Markdown in video descriptions gets reduced to plain text in the video descriptions.

image

image

MatthiasReumann:
### Motivation and Context
- Fixes #729 

### Description
- Increase cutoff length to 256
- Remove truncated variables but keep `truncateHtml` function
- Update UI (a bit)
- Add list css

### Steps for Testing
1. Log in as admin
2. Navigate to a Livestream
3. Update description with lists, etc.
4. Should be changed

### Screenshots

https://user-images.githubusercontent.com/29633518/199230476-800fc3af-7474-44a4-9f05-51384568ddc7.mov

### DIFF
--- a/api/courses.go
+++ b/api/courses.go
@@ -90,7 +90,7 @@ type coursesRoutes struct {
 
 const (
 	WorkerHTTPPort = "8060"
-	CutOffLength   = 100
+	CutOffLength   = 256
 )
 
 type uploadVodReq struct {
@@ -644,8 +644,7 @@ func (r coursesRoutes) updateDescription(c *gin.Context) {
 	}
 	wsMsg := gin.H{
 		"description": gin.H{
-			"full":      stream.GetDescriptionHTML(),
-			"truncated": tools.Truncate(stream.GetDescriptionHTML(), 100),
+			"full": stream.GetDescriptionHTML(),
 		},
 	}
 	if msg, err := json.Marshal(wsMsg); err == nil {
--- a/api/courses.go
+++ b/api/courses.go
@@ -645,7 +645,7 @@ func (r coursesRoutes) updateDescription(c *gin.Context) {
 	wsMsg := gin.H{
 		"description": gin.H{
 			"full":      stream.GetDescriptionHTML(),
-			"truncated": tools.Truncate(stream.GetDescriptionHTML(), 150),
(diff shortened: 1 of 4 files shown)


joschahenningsen (changes requested):
Looks great, thanks! :heart:  I noticed a few quirks though:

- When the description is short (nothing to expand) the cursor is still a pointer when hovering the description
-  Expanding is dependent on the number of characters, could it be the number of lines instead to prevent quirks like this?


https://user-images.githubusercontent.com/44805696/199276033-fdcdd699-0141-4e70-b69e-5fe38f781c08.mp4

We could also go for something like `if (numLines>5 || numChars>100){ // resizable }` but you are  the expert here :)

MatthiasReumann:
> Looks great, thanks! ❤️ I noticed a few quirks though:
> 
> * When the description is short (nothing to expand) the cursor is still a pointer when hovering the description
> * Expanding is dependent on the number of characters, could it be the number of lines instead to prevent quirks like this?
> 
>  out.mp4 
> 
> We could also go for something like `if (numLines>5 || numChars>100){ // resizable }` but you are the expert here :)

Number of lines makes sense. Will have a look into it!

MatthiasReumann:
> Looks great, thanks! ❤️ I noticed a few quirks though:
> 
> * When the description is short (nothing to expand) the cursor is still a pointer when hovering the description
> * Expanding is dependent on the number of characters, could it be the number of lines instead to prevent quirks like this?
> 
>  out.mp4 
> 
> We could also go for something like `if (numLines>5 || numChars>100){ // resizable }` but you are the expert here :)

What do you actually think about going back to using `overflow-y: clip;` again, combined with the cutoff-length?

https://user-images.githubusercontent.com/29633518/199302924-c4c79463-444c-4f1b-a99b-fb7ed12d1640.mov


https://user-images.githubusercontent.com/29633518/199302930-2bddd568-c015-4cb2-b71e-6f37b1214945.mov

joschahenningsen:
Sure, the simplest solution is always the best if it works!

joschahenningsen (approved):
Thanks!

Issue #782, opened 2022-11-12
Un-approve a chat message


Jojaa350:
**Is your feature request related to a problem? Please describe.**
This might be nit-picking, but I feel like it should be possible to un-approve a chat message without deleting it entirely.
 
**Describe the solution you'd like**
Upon approving a message, the approval-button should change into a new retraction-button (instead of disappearing), which, if clicked, simply reverts the approval of the message such that it disappears for everybody except the moderators and author.

PR #786, opened 2022-11-14
Add un-approve/retract message

MatthiasReumann:
### Motivation and Context
- Discussed in issue #782 

### Description
Adds ws event "retract" to backend and frontend. 

### Steps for Testing
- 1 Lecturer
- 1 Students
- 1 Livestream

1. Log in
2. Navigate to a Livestream that has moderate-chat enabled.
3. Write a message as a student
4. Approve as admin
5. Retract as admin

### Screenshots
image

### DIFF
--- a/api/chat.go
+++ b/api/chat.go
@@ -70,6 +70,8 @@ func RegisterRealtimeChatChannel() {
 				routes.handleResolve(tumLiveContext, message.Payload)
 			case "approve":
 				routes.handleApprove(tumLiveContext, message.Payload)
+			case "retract":
+				routes.handleRetract(tumLiveContext, message.Payload)
 			default:
 				log.WithField("type", req.Type).Warn("unknown websocket request type")
 			}
@@ -219,7 +221,7 @@ func (r chatRoutes) handleCloseActivePoll(ctx tools.TUMLiveContext) {
 }
 
 func (r chatRoutes) handleResolve(ctx tools.TUMLiveContext, msg []byte) {
-	var req resolveReq
+	var req wsIdReq
 	err := json.Unmarshal(msg, &req)
 	if err != nil {
 		log.WithError(err).Warn("could not unmarshal message delete request")
@@ -246,7 +248,7 @@ func (r chatRoutes) handleResolve(ctx tools.TUMLiveContext, msg []byte) {
 }
 
 func (r chatRoutes) handleDelete(ctx tools.TUMLiveContext, msg []byte) {
-	var req deleteReq
+	var req wsIdReq
 	err := json.Unmarshal(msg, &req)
 	if err != nil {
 		log.WithError(err).Warn("could not unmarshal message delete request")
@@ -271,7 +273,7 @@ func (r chatRoutes) handleDelete(ctx tools.TUMLiveContext, msg []byte) {
 }
 
 func (r chatRoutes) handleApprove(ctx tools.TUMLiveContext, msg []byte) {
-	var req approveReq
+	var req wsIdReq
 	err := json.Unmarshal(msg, &req)
 	if err != nil {
 		log.WithError(err).Warn("could not unmarshal message approve request")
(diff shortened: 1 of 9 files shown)


api/chat.go
```diff
@@ -307,8 +309,45 @@ func (r chatRoutes) handleApprove(ctx tools.TUMLiveContext, msg []byte) {
...
+	if err != nil {
+		log.WithError(err).Error("could not retract chat")
+		return
+	}
+
+	/* UserId should be the user who gets the message, to add dynamic user specific flags (e.g. Liked)
+	 * to the message payload. In this case the Message is freshly approved so no users should have interacted
+	 * with that message so far, so we pass 0 instead of a userId.
+	 */
+	chat, err := r.ChatDao.GetChat(req.Id, 0)
```
MatthiasReumann, line 335: I'm not really sure which userID I should set here. I don't necessarily want to re-implement the self-inflicted "DDOS attack", so could you please help me out @mono424? 😅

mono424: If I understand right the message will be removed from all users so you could just use `0` here. The function only adds a `Liked` flag, which is individual for every user, but if the message disappears anyways this is probably not important.

But if you just want to remove the message maybe it's easier to just send an identifier of the message, in that case, you would not even need to call the `getChat()` function i guess :)
Than you could just do
```
broadcast := gin.H{
	"retract": req.Id,
}
```
or am I missing something ?

edit: Oh I see the Moderators should still see the message, after the retraction ... But maybe thats something better handled in frontend? If you want to use the GetChat method you would need to loop through all users and send them individual messages for their id which is kinda disgusting. On my opionion the 'Liked'-Flag should be attached in frontend anyways to get rid of the id parameter of `getChat()` 😅

MatthiasReumann: Thanks! So for the meantime I will keep using `0` and maybe we can refactor this in an another PR.

mono424: Can a moderator approve a message, than retract it and than approve it again? If so does the Likes be still present on the message? Because if yes I think there could a bug:
1. Message gets approved
2. UserA likes the messages
3. Message gets retracted
4. Message gets approved again
5. ! User does not see his like anymore

Is that an issue ? If not than nothing to worry about :)

MatthiasReumann: Yup, approving, retracting, and approving it again is possible. Likes are however still visible. Basically what retract does is set `visible` to false.

mono424: Yes but the `Liked` flag will not be set right I believe. Because in the approve message function we also use `r.ChatDao.GetChat(req.Id, 0)` because we believe there are no likes on the post. If there are already likes on the post the `Liked` flag will be still false for everone.

MatthiasReumann: Yup. A possible (and probably the easiest one) solution would be to delete all likes of the message when retracting, right?

mono424: yes I think that would be the best solution ! 👍

mono424: edit: maybe not the best but the fastest to implement! 😄

mono424 (commented):
Looks great! Pretty nice feature too!

I ran into a small issue, to reconstruct:
1. Open two sessions where you can see the chat
2. Send a message as admin (should be visible on the second chat)
3. Retract the message as admin ( !! still visible on the second chat with a little mark (only visible to you) !! )

Expected: The chat message hides in the other chat.

MatthiasReumann:
> Looks great! Pretty nice feature too!
> 
> I ran into a small issue, to reconstruct:
> 
> 1. Open two sessions where you can see the chat
> 2. Send a message as admin (should be visible on the second chat)
> 3. Retract the message as admin ( !! still visible on the second chat with a little mark (only visible to you) !! )
> 
> Expected: The chat message hides in the other chat.

Great catch! Admin messages shouldn't be retractable, since they aren't approved in the first place.

mono424:
Nice Admin-Retract Bug is fixed! 👍 

There is still a little issue with the implementation, I attached a Video for clarification :)

https://user-images.githubusercontent.com/17506411/202205434-34396629-e308-40d5-9372-af54d95de01c.mov

MatthiasReumann:
> Nice Admin-Retract Bug is fixed! 👍
> 
> There is still a little issue with the implementation, I attached a Video for clarification :)
> 
>  Bildschirmaufnahme.2022-11-16.um.15.20.52.mov

Mhm. Thanks for pointing out - I will look into it!

mono424:
> > Nice Admin-Retract Bug is fixed! 👍
> > There is still a little issue with the implementation, I attached a Video for clarification :)
> > Bildschirmaufnahme.2022-11-16.um.15.20.52.mov
> 
> Mhm. Thanks for pointing out - I will look into it!

The `visible` flag is only used to tag the messages in the frontend with the "only visible for your"-tag. I think the easiest way is to change the onRetract message handler in the frontend to remove the message completely if you are not the author nor an admin.

MatthiasReumann:
> > > Nice Admin-Retract Bug is fixed! 👍
> > > There is still a little issue with the implementation, I attached a Video for clarification :)
> > > Bildschirmaufnahme.2022-11-16.um.15.20.52.mov
> > 
> > 
> > Mhm. Thanks for pointing out - I will look into it!
> 
> The `visible` flag is only used to tag the messages in the frontend with the "only visible for your"-tag. I think the easiest way is to change the onRetract message handler in the frontend to remove the message completely if you are not the author nor an admin.

I've added a filter function to solve that problem. Great advice thanks! 🚀 

However, in order to achieve that I added the `userId` to the chat-message response. I'm not sure if this will cause privacy problems. My guess is that it's fine.
(?) @joschahenningsen

joschahenningsen:
No that's fine, we already disclose user IDs through the @-ing users feature.

mono424:
Now works as expected! Very nice! 👍

mono424 (approved):

joschahenningsen:
:heart:

PR #791, opened 2022-11-19
Only consider recordings for hiding watched streams

alexanderstephan:
### Motivation and Context

We also load future streams into the VoD list. This breaks the behavior of hiding months or showing the `All VoDs watched` message.

### Description
By also checking whether a stream is a recording this should be fixed, as the recordings are what the user sees in VoD list.
I am not sure why also get future lectures but this should work regardless of what other lectures we load.

### Steps for Testing

Prerequisites:
- 1 Livestream with VoDs and upcoming streams
- Any account will do

1. Log in
2. Navigate to the course page of a stream with some VoDs
3. Check that the most recent month that should have additional future lectures is hidden
4. Check that `All VoDs watched` appears when all VoDs are marked as watched.

### DIFF
--- a/web/course.go
+++ b/web/course.go
@@ -5,10 +5,9 @@ import (
 	"fmt"
 	"github.com/joschahenningsen/TUM-Live/model"
 	"github.com/joschahenningsen/TUM-Live/tools"
+	log "github.com/sirupsen/logrus"
 	"html/template"
 	"net/http"
-
-	log "github.com/sirupsen/logrus"
 
 	"github.com/getsentry/sentry-go"
 	sentrygin "github.com/getsentry/sentry-go/gin"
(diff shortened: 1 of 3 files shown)


MatthiasReumann (approved):
Looks good 👍🏻 

As you pointed out, it might be worth considering refactoring the course page so that VoD's, and future streams are two separate REST calls that return already structured data. 
Currently, my sloppy implementation I should say, sorts and does all kind of crazy stuff in the .gohtml template and it always takes a while to understand what's actually going on. 😅

alexanderstephan:
> As you pointed out, it might be worth considering refactoring the course page so that VoD's, and future streams are two separate REST calls that return already structured data.

Good idea, I'll try update this!

joschahenningsen (approved):

PR #794, opened 2022-11-21
Add katex to pop-out

MatthiasReumann:
### Motivation and Context
Latex doesn't work in the popout chat. The dependencies are missing.

### Description
Add dependencies.

### Steps for Testing
- 1 Lecturer
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. Pop out chat
4. Write latex

### Screenshots
image


joschahenningsen (approved):
Good catch!

PR #795, opened 2022-11-21
Remove server messages

### LINKED ISSUE #793: Remove server messages
**Describe the problem**
The server sends messages to the chat right now such as:
1. `Welcome to the chatroom! Please be nice to each other and stay on topic if you want this feature to stay active.`
2. `The broadcaster disabled anonymous messaging for this stream.`

I would argue that this is not needed anymore. The chat can be moderated and was never really abused from my experience.
Also, people should be able to see whether they can send anonymous messages by the icon. I don't think it is necessary to announce this to users as I would expect the default to be not anonymous.

These server messages cause sometimes weird behavior e.g. it appears as if there was a new message sent when watching a stream. Furthermore, in case of disconnects these messages can stack up which looks rather ugly.

MatthiasReumann:
### Motivation and Context
Issue: #793 

### Description
Remove the code that sends the  server messages.

### Steps for Testing
- 1 Livestream with Chat

1. Log in
2. Navigate to a Livestream
3. No server messages should be sent

### DIFF
--- a/api/wsHub.go
+++ b/api/wsHub.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 )
 
 var wsMapLock sync.RWMutex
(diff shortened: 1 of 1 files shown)


joschahenningsen:
I'd keep the server message logic and annotate it for staticcheck 

`//lint:ignore U1000 Ignore unused function `

alexanderstephan (approved):
Works perfectly, thanks!

PR #812, opened 2022-11-30
add test users to testserver comment

joschahenningsen:
### Steps for Testing

1. Is there a comment telling you which users are available?
2. do they work?

### DIFF
--- a/.github/workflows/pr-open.yml
+++ b/.github/workflows/pr-open.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ dev ]
-    types: [ opened ]
+    types: [ opened, reopened ]
 
 jobs:
   comment:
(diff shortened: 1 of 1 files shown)


alexanderstephan:
Looks good! The test server works fine, I also tested all user types, which also seemed to work fine. I also see the comment. Is there anything else to do here?

joschahenningsen:
No that's all. I noticed that let's encrypt rate limits based on the base domain (tum.de) so we might have to wait a minute longer from time to time.

joschahenningsen:
Feel free to approve :)

alexanderstephan:
> No that's all. I noticed that let's encrypt rate limits based on the base domain (tum.de) so we might have to wait a minute longer from time to time.

Ah, that's totally ok. :)

alexanderstephan (approved):
:partying_face:

PR #839, opened 2022-12-06
Add golangci-lint

alexanderstephan:
### Motivation and Context

Our CI only catches a small portion of issues. Problems such as code duplication or more in-depth security-checks are lacking.

### Description

- [golangci-lint](https://github.com/golangci/golangci-lint) combines the most popular and useful linters for go.
- A corresponding CI action is added.  This PR is only for setting up the CI itself.
-  The linter is configured in a way that only issues for upcoming PRs are displayed. Fixing the linting issues will be done in multiple follow up PRs.


MatthiasReumann (approved):
Great idea! 🌟

I feel like we need to merge this to actually see the changes and improvements. If any problems arise we can always rollback to the old workflow.

PR #841, opened 2022-12-11
Add copying of up-to-date timestamps via context menu

### LINKED ISSUE #851: Automatically add the current timestamp when using the share button
 I'm always frustrated when sharing a specific timestamp of a lecture.

Already add the current timestamp, when pressing the share button

Fill in the current timestamp in the optional timestamp field of the share submenu

alexanderstephan:
### Motivation and Context

- The context menu looks a bit empty right now.
- The timestamp is only loaded when the watch page is initially loaded. One might expect that the timestamp is always up-to-date

### Description

- Extended the video context menu by two new entries:
  - `Copy URL` copies the URL of current VoD
  - `Copy URL at current time` copies the URL of the VoD. It also includes the current time of VoD.
- New CSS class for menu entries called `.tl-context-entry`
- The share menu also now takes the current time into account. The time is fetched every time the share menu is open.

Minor change:

Use `npm i --include=dev` instead of `npm i --dev` since the expected behavior would usually be that regular dependencies are installed alongside the dev dependencies.

### Steps for Testing

There is a surprising amount of edge cases, so here is a rather long list of testing steps. Maybe you can think of additional cases. Possibly also related to https://github.com/joschahenningsen/TUM-Live/pull/603.

### Prerequisites:
- 1 User
- 1 VoD

1. Log in
2. Navigate to a VoD and start it.
3. Right click on the video player
4. Select `Copy URL`
5. Observe that is copied the URL without a timestamp
6. Right click and again and select `Copy URL at current time`
7. Observe that a timestamp is encoded in the URL for the current time. Try if the link works.
8. Click on the share button.
9. Select `Include timestamp`
10. Check that it shows the current URL in text field.
11. Close the window, let the video play and repeat. This time check that it fetched the time when the share button was pressed.
12. Unselect the `Include timestamp` again. Observe that the time argument is deleted. Also make sure that the time is **not** updated in this case.
13. 🎉 

### Screenshots

image

### Possible follow ups

Animations when the url is copied via the context menu would be cool. The context menu als does not support dark mode yet which might be also desirable.


joschahenningsen:
> Use npm i --include=dev instead of npm i --dev since the expected behavior would usually be that regular dependencies are installed alongside the dev dependencies.

Did you test if the docker container works as expected?

joschahenningsen (changes requested):
On the testserver it always copies the url with the timestamp it was when I first copied it, even if the time has changed in the mean time, any idea why?

alexanderstephan:
> Did you test if the docker container works as expected?

I forgot a `true`. Will be fixed in a second. 

> Did you test if the docker container works as expected?

Will do

joschahenningsen:
> Use npm i --include=dev instead of npm i --dev since the expected behavior would usually be that regular dependencies are installed alongside the dev dependencies.
 
This should however mean that our binary will be significantly larger if compiled this way. When the makefile is only used locally this should not be an issue. I'd like to keep unneeded dependencies out of the production docker build tho.

alexanderstephan:
> This should however mean that our binary will be significantly larger if compiled this way. When the makefile is only used locally this should not be an issue. I'd like to keep unneeded dependencies out of the production docker build tho.

As far as I know, it isn't used in the Dockerfile. 🤔

alexanderstephan:
I noticed that the page layout is encoded in the URL e.g. `https://841.test.live.mm.rbg.tum.de/w/t2/6/SPLIT?t=307`. Not sure, if everyone will like this From my side that is okay, just wanted to mention it.

alexanderstephan:
I guess that's a good thing, actually. I could imagine this being useful when someone e.g. wants to highlight what happened on camera.

joschahenningsen (approved):
Great addition!

PR #852, opened 2022-12-27
Fix duplicated handler

alexanderstephan:
### Motivation and Context
I just set up a new dev environment and noticed that TUM-Live wouldn't start due to a duplicated handler.

### Description
Set the correct handler name to remove duplicates.

### Steps for Testing

1. Try `run-dev` in goland.
2. See, if the TUM-Live starts.


joschahenningsen (approved):
Thanks :D

PR #883, opened 2023-01-17
Add autocomplete

MatthiasReumann:
### Motivation and Context
Screenshot 2023-01-17 at 07 18 49

### Description
Add `autocomplete=""` to password input field.

### Steps for Testing
1. Navigate to Login page
2. Click "Internal Account"
3. There shouldn't be an error in the console
4. 🚀

### DIFF
--- a/web/template/login.gohtml
+++ b/web/template/login.gohtml
@@ -41,7 +41,7 @@
                 <div class="text-sm">
                     <label for="password" class="block text-5">Password</label>
                     <input type="password" name="password" id="password" required
-                           class="tl-input mt-3"
+                           class="tl-input mt-3" autocomplete=""
                            placeholder="Password"/>
                 </div>
                 <button type="submit"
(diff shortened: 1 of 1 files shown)


web/template/login.gohtml
```diff
@@ -41,7 +41,7 @@
                 <div class="text-sm">
                     <label for="password" class="block text-5">Password</label>
                     <input type="password" name="password" id="password" required
-                           class="tl-input mt-3"
+                           class="tl-input mt-3" autocomplete=""
```
joschahenningsen: Why not use 
```suggestion
                           class="tl-input mt-3" autocomplete="current-password"
```
As suggested by the browser?

joschahenningsen (changes requested):
While we're at it we could add `autocomplete="username"` to the username field

web/template/login.gohtml
MatthiasReumann: Thanks. Apparently I can't read. 🫠

joschahenningsen (approved):

PR #884, opened 2023-01-17
Add class 'Time'

MatthiasReumann:
### Motivation and Context
This PR is part of the watch page refactoring. Seconds to Hours/Minutes/Seconds is a common use case for TUM-Live, therefore wrap this into an easily reusable class. Furthermore utilize `Date` instead of custom solutions (See `currentTimeToHMS`).

### Description
- Add TypeScript class `Time`.
- Remove `friendlyTimestamp` from video sections and bookmarks
- Use `Time` in `ShareURL` class
- Update Tests

### Steps for Testing
- 1 Account
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. Create Bookmarks
4. Click on Video Sections
5. Check if sharing still works
6. Check if the highlighting of video sections still works
7. Check if the admin side of things for video section still work
8. 🌟

(I added some sections to https://884.test.live.mm.rbg.tum.de/w/t2/6)

### DIFF
--- a/api/bookmarks.go
+++ b/api/bookmarks.go
@@ -113,20 +113,7 @@ func (r bookmarkRoutes) GetByStreamID(c *gin.Context) {
 		return
 	}
 
-	response := []gin.H{}
-	for _, bookmark := range bookmarks {
-		response = append(response, gin.H{
-			"ID":                bookmark.ID,
-			"hours":             bookmark.Hours,
-			"minutes":           bookmark.Minutes,
-			"seconds":           bookmark.Seconds,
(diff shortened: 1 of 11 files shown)


alexanderstephan (approved):
Looks great! Much cleaner to unify the timestamp logic.
We could also update the [watch-info stuff](https://github.com/joschahenningsen/TUM-Live/blob/dev/web/template/partial/stream/watch-info.gohtml) soon, which is very alpine heavy and also involves a lot of timestamp logic.

PR #886, opened 2023-01-18
Move timestamp to bottom of thumbnail

### LINKED ISSUE #571: Move thumbnail timestamp to bottom edge
It sometimes is somewhat difficult to quickly read the timestamps displayed in the seek thumbnails (see screenshots below), because they are white text on a transparent background (with a tiny black shadow) placed above the usually also white contents of the lecture slides.

At least for the presentation-plus-camera video option, there will always be a black bar at the bottom of the video (as a result of how the individual video streams for the presentation and the camera are placed next to each other), so it might be a good idea to move the timestamp to the bottom of the thumbnail, into that always-black space.

This obviously will not really improve things for the other two video options (presentation-only and camera-only), though even there i'd guess that for e.g. the presentation mode the top part of the slides will usually have more content (and therefore make reading the timestamp more difficult) than the bottom part of the slides.

- Just keep it as-is
- Keep the timestamp at the top but add some kind of grey background (instead of the current black shadow thing) so that it is clearly legible

---

Screenshot 1:

Screenshot 2:

alexanderstephan:
### Description

Fixes https://github.com/joschahenningsen/TUM-Live/issues/571 by moving the timestamp to the bottom of the thumbnail where typically the contrast is better due to black bars. Also sorted the CSS so any tooltip styling is at the same location.

### Steps for Testing

Probably best to test this locally. This requires database or code changes that are not suitable for production.

Prerequisites:
- 1 Lecturer/Student/Admin
- 1 VoD
- Hardcode a thumbnail URL and interval here: https://github.com/joschahenningsen/TUM-Live/blob/520b0cd41c04cc4b791551cb982d5d14c8eff707/web/ts/TUMLiveVjs.ts#L61
- The URL can refer to any thumbnail such as https://live.rbg.tum.de/api/stream/20065/thumbs/26524

1. Navigate to a Livestream
2. Hover over the time
3. Observe, that the time is on the bottom of the thumbnail.
4. Remove thumbnail and repeat. It should still look the same.


MatthiasReumann (approved):
Looks great. 🌟

PR #890, opened 2023-01-20
allow getting stream from edge server with admin token, add authentication tests

joschahenningsen:
### Motivation and Context

For automated scripts it's useful to be able to get streams without a jwt.

### Description

If the environment variable `ADMIN_TOKEN` is set and a provided jwt matches the token, the request is handled instead of aborted.

### Steps for Testing

Code review should be enough.


worker/edge/README.md
```diff
@@ -23,3 +23,4 @@ The following configuration options are available via environment variables:
 - `CERT_DIR`: If specified, the edge node will use the files ending with `fullchain.pem` and `privkey.pem` in this directory to serve https connections on port `8443`.
 - `VOD_DIR`: A directory that is statically served under the path `/vod` to access vods generated by `vod-service`. Defaults to `/vod`
 - `MAIN_INSTANCE`: The url where your TUM-Live instance is available. Used for public key exchange. Defualts to `http://localhost:8081`
+- `ADMIN_TOKEN`: Can be used in place of the ?jwt query parameter to authenticate for streams. No default.
```
alexanderstephan:
```suggestion
- `ADMIN_TOKEN`: Can be used in place of the `?jwt` query parameter to authenticate for streams. No default value set.
```

worker/edge/README.md
```diff
@@ -23,3 +23,4 @@ The following configuration options are available via environment variables:
 - `CERT_DIR`: If specified, the edge node will use the files ending with `fullchain.pem` and `privkey.pem` in this directory to serve https connections on port `8443`.
 - `VOD_DIR`: A directory that is statically served under the path `/vod` to access vods generated by `vod-service`. Defaults to `/vod`
 - `MAIN_INSTANCE`: The url where your TUM-Live instance is available. Used for public key exchange. Defualts to `http://localhost:8081`
```
alexanderstephan:
```suggestion
- `MAIN_INSTANCE`: The url where your TUM-Live instance is available. Used for public key exchange. Defaults to `http://localhost:8081`
```

worker/edge/edge.go
```diff
@@ -110,43 +113,56 @@ type JWTPlaylistClaims struct {
 	Playlist string
 }
 
+func validateToken(w http.ResponseWriter, r *http.Request) bool {
+	token := r.URL.Query().Get("jwt")
+	if token == "" {
+		w.WriteHeader(http.StatusForbidden)
+		_, _ = w.Write([]byte("Forbidden"))
```
alexanderstephan:
```suggestion
		http.Error(w, "Missing JWT", http.StatusForbidden)
```
What about using [`http.Error`](https://pkg.go.dev/net/http#Error) for all error responses?

worker/edge/edge.go
```diff
@@ -48,6 +48,8 @@ var allowedOrigin = "*"
 
 var mainInstance = "http://localhost:8081"
 
+var adminToken = ""
```
alexanderstephan, line 51: Preference, but a struct type like `edgeConfig` that groups all the proxy configurations after the first `var` block could be an alternative here. This could then be used as a receiver type for some of the methods.

worker/edge/edge_test.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+const testURL = "http://localhost/vod/test.mp4/playlist.m3u8"
+
+func BenchmarkValidateToken(b *testing.B) {
+	str, err := prepareJWT(time.Hour, testURL)
+	if err != nil {
+		b.Fatal(err)
+	}
+	r, _ := http.NewRequest("GET", testURL+"?jwt="+str, nil)
```
alexanderstephan:
```suggestion
	r, _ := http.NewRequest(http.MethodGet, testURL+"?jwt="+str, nil)
```
Alternatively, `http` constants could be used here. But doesn't really matter.

alexanderstephan (changes requested):
Looks really good! 🍻
I only have small suggestions/remarks.

worker/edge/edge.go
joschahenningsen: Yes, will do when refactoring the jwt into livestreams as well.

PR #891, opened 2023-01-20
Feature/unified player bar in splitview

mono424:
### Motivation and Context
In split view we have currently two player bars, where the bigger player's player bar is active and the other one is disabled.
To enable the sound controls we currently mute the smaller player and unmute the bigger player with the player bar.
As the players are not perfectly synchronous, if the PresentationSlide-Player is bigger, the sound can be a little off to the camera player.

### Description
Only the camera players sound and controlbar is used, so the sound is not off. Also the playerbar is aligned over 
the whole video to give the same user expereience than in the old player.

### Steps for Testing
Prerequisites:
- 1 Student
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. Open Split Player
4. Check if all the controls and seeking is working and the controlbar is aligned correctly
5. Check if subtitles are displayed correctly

### Screenshots
Bildschirm­foto 2023-01-20 um 14 51 19
nshots


joschahenningsen (approved):
Looks amazing and feels great, I'll leave merging to @MatthiasReumann because you know the split view best!

web/ts/splitview.ts
```diff
@@ -52,15 +56,38 @@ export class SplitView {
         return [100 - this.camPercentage, this.camPercentage];
     }
 
-    private toggleControlBars(camPercentage: number) {
-        let i = 0,
-            j = 1;
-        if (camPercentage > 50) {
-            (i = 1), (j = 0);
+    private cloneEvents(srcElem: HTMLElement, destElem: HTMLElement, events: string[]) {
```
MatthiasReumann: I think this function could also be useful elsewhere, so we might want to move it to the `global.ts` file.

MatthiasReumann (changes requested):
Looks awesome 🌟

However, I noticed a peculiar bug (MacOS: Safari, Chrome) that the control bar in fullscreen is misaligned (This is only 1/4 of my screen - imagine the rest):
Screenshot 2023-01-22 at 07 54 33

Furthermore, as you already mentioned @mono424, we might want to style the splitter differently in a follow-up PR.

web/ts/splitview.ts
mono424: Good catch! Also kind of bad is that you can only fullscreen one of the two players. You would probably expect to fullscreen the whole split view ... may also something for a followup!

mono424: @MatthiasReumann https://github.com/joschahenningsen/TUM-Live/pull/895 lets the user make the splitview fullscreen

MatthiasReumann (approved):
Fullscreen looks great now, thanks! 🚀

Could you check after merging `split-view.ts` if the subtitles still work properly? 

<details>
<summary>Insert statement for testing (Replace the stream_id accordingly)</summary>
<p>

```sql
insert into subtitles (created_at, updated_at, stream_id, content, language) values (Now(), Now(), 1337, 'WEBVTT

1
00:00:00.000 --> 00:00:07.000
I''m gonna shift the nozzle to the empty side of the stand.

2
00:00:30.000 --> 00:00:48.200
Now at home self isolating you to a positive COVID test but those two things have in common

3
00:00:48.200 --> 00:00:51.000
is that you cannot be here to give this lecture.

4
00:00:51.000 --> 00:00:54.840
So instead I will do the lecture today.

5
00:00:54.840 --> 00:01:00.680
So last time what we were trying to do is we were trying to actually complete the proof that

6
00:01:00.680 --> 00:01:07.280
when I have used this or when I have annotated a control flow graph with assertions that

7
00:01:07.280 --> 00:01:12.960
are locally consistent that if I have done this and I know that the assertion at the beginning

8
00:01:12.960 --> 00:01:18.320
of a path holds then at the end of the path this assertion that is annotated there will

9
00:01:18.320 --> 00:01:20.040
hold.

10
00:01:20.040 --> 00:01:25.560
And let''s quickly recap what we define here we define an execution trace and this trace

11
00:01:25.560 --> 00:01:31.040
we call Pi and one of these execution traces corresponds to a path in the control

12
00:01:31.040 --> 00:01:38.200
flow graph and it basically starts at some initial program point u0 and at program point

13
00:01:38.200 --> 00:01:43.320
u0 you have this state sigma 0.

14
00:01:43.320 --> 00:01:50.000
So here and then it leads to some program point um and at this program point um you have

15
00:01:50.000 --> 00:01:59.720
this final state sigma m and then one can simply represent the trace as a sequence of

16
00:01:59.720 --> 00:02:04.720
first of all these topos of program points.

17
00:02:04.720 --> 00:02:07.360
Where''s my mouse here.

18
00:02:07.360 --> 00:02:13.920
These topos of program points in states and then you have these labels.

19
00:02:13.920 --> 00:02:19.560
So these correspond to the things that happen in the control flow graph and you simply

20
00:02:19.560 --> 00:02:24.280
have a sequence of these.

21
00:02:24.280 --> 00:02:30.240
Now okay this is this and then here was the theorem that we were trying to prove.

22
00:02:30.240 --> 00:02:35.620
So we had some sort of mini java program p and we had some path pi that was an execution

23
00:02:35.620 --> 00:02:42.040
trace that starts in some program point u and leads to this program point v and we had

24
00:02:42.040 --> 00:02:47.420
assumed that the all program points in p are annotated with assertions which are locally

25
00:02:47.420 --> 00:02:53.660
consistent and then we had set the program point u is annotated with a and the program point

26
00:02:53.660 --> 00:03:00.980
v is annotated with b and if these things hold then we want to conclude that if the initial

27
00:03:00.980 --> 00:03:10.900
state of pi satisfies a then the final state will satisfy b and then what we had done

28
00:03:10.900 --> 00:03:20.940
is we had started and then we had decided that we want to do this proof by induction namely

29
00:03:20.940 --> 00:03:28.140
on the length m of this execution trace and we had also done the first step and the first

30
00:03:28.140 --> 00:03:34.580
step is if the path has length 0 and then sort of trivial the end point of the execution

31
00:03:34.580 --> 00:03:42.060
is equal to the start point and then this sigma 0 is equal to sigma m and the assertions

32
00:03:42.060 --> 00:03:46.380
that we have annotated are both the same so a is the same as b and then the claim sort

33
00:03:46.380 --> 00:03:56.900
of trivial holds and then we were gearing up to get into more complicated steps so

34
00:03:56.900 --> 00:04:02.500
by the induction step and what we needed there was this proposition here that if I have

35
00:04:02.500 --> 00:04:13.340
some arithmetic expressions t and e then it is the same whether I first update the state

36
00:04:13.340 --> 00:04:20.620
to set x to the value that is expression e would evaluate to in the state sigma and then

37
00:04:20.620 --> 00:04:27.140
evaluate this expression t on this entire thing or alternatively it is also the same if

38
00:04:27.140 --> 00:04:35.660
I simply go in and take my expression and wherever I find an x in my original expression

39
00:04:35.660 --> 00:04:43.820
t I simply pop in this e and I think this was the point where you stopped last week so

40
00:04:43.820 --> 00:04:48.500
to convince you of this we will actually give you a small example where this does hold

41
00:04:48.500 --> 00:04:54.060
and then the actual proof you will see in the homeworks this next week so let us consider

42
00:04:54.060 --> 00:05:02.460
this example expression where t is x plus y and e is 2 times z and we have also have

43
00:05:02.460 --> 00:05:12.860
a state and our state sigma in this case is that x is 5 y is minus 1 and z is 21 no

44
00:05:12.860 --> 00:05:20.060
this direction okay and then let us first maybe tackle the left hand side of this

45
00:05:20.060 --> 00:05:31.820
equality that we claim here and what we do here is we update our state here and first of all

46
00:05:31.820 --> 00:05:38.180
we check what e evaluates to in the state sigma and now we know that z here is 21 we know

47
00:05:38.180 --> 00:05:59.460
this from here so this expression e this thing here and then what we do is we go here and we

48
00:05:59.460 --> 00:06:06.420
update our state such that we set this x to 42 and what this does is we take the state

49
00:06:06.420 --> 00:06:13.300
from here and then we simply swap in this 42 here so we end up with a new state where x is 42

50
00:06:13.300 --> 00:06:21.540
y is minus 1 and z is 21 and then we take our expression t which was this x plus y and we

51
00:06:21.540 --> 00:06:29.620
evaluated on the state that we have so for x we use 42 and for y we use minus 1 and then we

52
00:06:29.620 --> 00:06:36.500
sort of obviously end up with the value 41 down here so this was the one direction now it''s

53
00:06:36.500 --> 00:06:42.100
seen in the other direction namely we take this expression t and then we replace the occurrences

54
00:06:42.100 --> 00:06:50.420
of x with e so when I look at the expression t it has exactly one occurrence of x namely this

55
00:06:50.420 --> 00:06:56.820
one and so I go in and replace this occurrence of x with this expression that I had 2 times z

56
00:06:56.820 --> 00:07:05.140
so then the new expression here becomes 2 times c plus y and then I evaluate this on the original

57
00:07:05.140 --> 00:07:11.860
state sigma and when I look up in sigma then I see that the value of z is 21 in my sigma

58
00:07:12.980 --> 00:07:18.980
and the value of y is minus 1 in my sigma and then I see that in this case there''s also

59
00:07:18.980 --> 00:07:28.020
e variance to 41 this was just a small example to convince you the general proof as I said will

60
00:07:28.020 --> 00:07:35.860
be in the homework then we have a further proposition let''s maybe take a part of this proposition

61
00:07:35.860 --> 00:07:48.180
means so it says in a state sigma that is updated by setting the value of x to whatever the expression

62
00:07:48.180 --> 00:07:58.980
e evaluates in this state and this new resulting state satisfies this assertion here for t1 and t2

63
00:07:58.980 --> 00:08:10.100
some expression if and only if the original state satisfies this formula here where I have

64
00:08:10.100 --> 00:08:19.060
um in t1 replaced all occur it says of x with e and I''ve also done the same in t2 so actually

65
00:08:19.060 --> 00:08:23.540
there''s not much going on here there''s just a little bit of formula you''re shuffling but let''s maybe

66
00:08:23.540 --> 00:08:30.900
do this do this proof here and look at the individual steps for this small case so we have this

67
00:08:30.900 --> 00:08:41.220
here which is our right side and then we say well this obviously only holds if what t1

68
00:08:41.220 --> 00:08:48.100
evaluates to in this in this state is less than what t2 evaluates to in this state so in some sense

69
00:08:48.100 --> 00:08:58.900
I pull this evaluation of these expressions I pull that in so I have this only if that or this should be

70
00:08:58.900 --> 00:09:05.460
less than that and then the next step this should look awful if a media to you because we have

71
00:09:05.460 --> 00:09:12.500
basically proven something about stuff that looks a bit like this already before so if we go back

72
00:09:12.500 --> 00:09:19.700
one slide we have just yeah we have just proven that those two things are the same or we have

73
00:09:19.700 --> 00:09:24.260
not really proven but I have said that always see an example and you will prove it in our homework

74
00:09:24.260 --> 00:09:31.860
so let''s assume it for now and then what we can do here is we use this thing that we have just

75
00:09:31.860 --> 00:09:40.500
not proven and plug it in here and then we can basically do this thing that we did before in reverse

76
00:09:41.220 --> 00:09:51.220
so if this expression that we have here evaluated on sigma is less than this expression evaluated

77
00:09:51.220 --> 00:10:06.340
on sigma here then obviously sigma satisfies the assertion that this expression here is less than the other one

78
00:10:09.860 --> 00:10:13.460
so these are some preliminary propositions three ones to get out of the way

79
00:10:13.460 --> 00:10:21.940
Marco are there relevant questions up to now okay so everybody''s happy that''s always nice so

80
00:10:21.940 --> 00:10:31.220
then let''s maybe continue oh okay so we have only seen this for this less for now and of course

81
00:10:31.220 --> 00:10:36.660
we need to prove this for every formula so formulas do often not just consist of an expression

82
00:10:36.660 --> 00:10:41.540
is less than another expression but usually there''s more things going on than just less

83
00:10:41.540 --> 00:10:46.580
there''s usually some conjunctions implications other things so one actually needs to prove this for

84
00:10:46.580 --> 00:10:52.820
every formula and we will here just quickly sketch how one could do this and the idea would be

85
00:10:52.820 --> 00:11:01.300
that one doesn''t induction of the structure of the formula so this is one further of these necessary

86
00:11:01.300 --> 00:11:05.620
prerequisites out of the way and we will heavily use this in the rest of the proof

87
00:11:05.620 --> 00:11:14.500
so now let''s continue so we actually maybe we can go I''ll go back again to see so that you see

88
00:11:14.500 --> 00:11:23.780
what we are trying to do and name so this and we want to do this induction on the length m

89
00:11:23.780 --> 00:11:29.700
and we have already proven for the base case where m is zero that it works so now what we want

90
00:11:29.700 --> 00:11:39.780
to do is we want to look at the inductive case and so let''s go here and here we have the

91
00:11:39.780 --> 00:11:46.580
induction hypothesis that the statement already holds for m minus one so if I have a path of length

92
00:11:46.580 --> 00:11:55.460
m then it already holds for all the steps except the last one and then I call the assertion at this

93
00:11:55.460 --> 00:12:04.500
last point u m minus one we have proven it I call that b prime and then as a first case let us

94
00:12:04.500 --> 00:12:12.260
deal with statements so we look at what happens when the thing that happens between step m minus one

95
00:12:12.260 --> 00:12:17.220
and step m is a statement and the first statement is a very nice statement it''s namely the

96
00:12:17.220 --> 00:12:26.660
empty statement so nothing happens here so we know that the weakest recondition for some formula b

97
00:12:27.700 --> 00:12:33.380
and the empty statement is simply b itself again and then we also know that b prime

98
00:12:34.900 --> 00:12:41.380
implies b and this we get from local consistency because we had assumed that the annotations are

99
00:12:41.380 --> 00:12:47.940
locally consistent so we know that this thing here holds and then it is sort of obvious because

100
00:12:47.940 --> 00:12:56.500
the state does not change if I do nothing it''s the same so both for sigma m minus one and sigma m

101
00:12:57.540 --> 00:13:05.540
they will both satisfy this formula b here so it''s maybe continue with a more interesting case

102
00:13:05.540 --> 00:13:13.380
namely the case where we do a right at first glance it might seem more interesting actually

103
00:13:13.380 --> 00:13:18.900
there is not so much more interesting things going on because once again a right does not

104
00:13:18.900 --> 00:13:25.540
modify the local state so we once again have that the state before at step i minus one the sigma

105
00:13:25.540 --> 00:13:30.580
m minus one simply stays the same when we go to sigma m so it''s completely the same nothing

106
00:13:30.580 --> 00:13:36.820
interesting happened the weakest precondition once again is simply b and then we have the same argument

107
00:13:36.820 --> 00:13:44.900
is before that b prime implies b and then we also have this argument once again that sigma m minus one

108
00:13:44.900 --> 00:13:52.660
is the same as sigma m and both of them satisfy b so the problem is something more interesting

109
00:13:52.660 --> 00:13:58.820
so actually let us see something more interesting namely we have an assignment where we assign

110
00:13:58.820 --> 00:14:05.940
if we can increase the font of this thing yes we have an assignment where we assign an expression

111
00:14:05.940 --> 00:14:12.420
e to this variable x so first of all let us think about what the state after this thing looks

112
00:14:12.420 --> 00:14:20.660
like so the state sigma m will be I take the state before and then I evaluate this expression m

113
00:14:20.660 --> 00:14:30.580
m is expression e on the state that I had before and then I update x so this is all that happens

114
00:14:30.580 --> 00:14:43.060
here so our new state here will look like this and then once again because of local consistency

115
00:14:43.060 --> 00:14:52.180
we had that the annotation that we have at this program point before this b prime it needs to

116
00:14:52.180 --> 00:15:15.300
imply the weakest precondition of this assignment so this thing here
', 'en');
```

</p>
</details>

mono424:
Subtitles work but not in a unified way, okay for now and fix in a followup or is mandatory for merge? What do you think @MatthiasReumann @joschahenningsen ? :)
Bildschirm­foto 2023-01-31 um 13 36 40

MatthiasReumann (approved):
🚀

joschahenningsen (approved):

PR #897, opened 2023-01-23
Resume at 1x speed when caught up with live stream

### LINKED ISSUE #667: Beta Stream: Change playback speed to 1.0x when the live content is reached
Beta Stream feature: I'm always frustrated when I start a lecture 30 min late, then set the playback speed to > 1.0 (e.g.) to catch up with the live stream. After some time, the 30-min buffer to the live stream is watched and the client tries to stream the live content with a play speed > 1.0. Because of this, every few seconds the client has to buffer (because the is no newer content than live) and then continues the playback with a speed > 1.0.

When reaching the live stream and the client has to buffer to get new content, reduce the playback speed to 1.0. Then the live stream can be watched without any hiccups.

Maybe a small notification could inform the user about the too high playback speed for the live stream. But this is not really a great solution.

**Additional context**
Beta Stream of any active live stream

xarantolus:
### Motivation and Context

When watching a live stream a few minutes behind at e.g. 2x speed, you'll catch up to live at some point. Currently you'll land in a loop of loading and then watching at 2x speed, rinse and repeat. 

### Description

Instead of keeping the current (too fast) playback rate, the rate will be adjusted when we reach the current live time. This leads to no disruptions when watching.
This fixes #667.

When compared to someone that always watched at 1x speed, the one that caught up will end up with a delay of 10-20 seconds. I tried to reduce this by trying out different values for the [`trackingThreshold`](https://docs.videojs.com/livetracker), but that didn't appear to help.
It might also make sense to wait until we buffer and then set to 1x if at the live edge, however that will result in a small interruption.

### Steps for Testing

Prerequisites:
- 1 Livestream

1. Change the `streams.playlist_url` in the database as described [here](https://github.com/joschahenningsen/TUM-Live/issues/667#issuecomment-1387440714) (thanks :))
2. Log in
3. Navigate to that Livestream
4. Switch to the beta stream
5. Seek back e.g. 30 seconds and set the playback rate to 2x
6. Wait until you've caught up with live
7. The playback rate should automatically switch back to 1x

### DIFF
--- a/web/ts/TUMLiveVjs.ts
+++ b/web/ts/TUMLiveVjs.ts
@@ -80,6 +80,14 @@ export const initPlayer = function (
     player.on("ratechange", function () {
         window.localStorage.setItem("rate", player.playbackRate());
     });
+
+    // When catching up to live, resume at normal speed
+    player.liveTracker.on("liveedgechange", function (evt) {
+        if (player.liveTracker.atLiveEdge() && player.playbackRate() > 1) {
+            player.playbackRate(1);
+        }
+    });
+
     player.ready(function () {
         player.airPlay({
             addButtonToControlBar: true,


MatthiasReumann (approved):
That's awesome, thanks 🌟

Works great.

PR #898, opened 2023-01-25
Add subtitles w/o admin

MatthiasReumann:
### Motivation and Context
* Add subtitles model and dao
* Add /api/stream/:id/subtitles/:lang endpoint + tests
* Dynamically load subtitles on the frontend

### Steps for Testing
Insert subtitles to database for stream

- 2 Livestream(s) one with, one without subtitles

1. Navigate to the Livestream with subtitles
2. Subtitles should be activateable
3. Navigate to the livestream without subtitles
4. No subtitle button should be visible

### Further Information
Not sure, but #891 might break the subtitles in split view. (@mono424)


joschahenningsen (approved):
Works great, thanks a ton! :heart:

PR #929, opened 2023-02-21
Reply UI

### LINKED ISSUE #925: Improve visibility of reply mode
I've seen the problem that people overlook that they are reply mode or even the other way around, that people send a normal message even if they intend to send a reply.
Especially when there is something typed in the chat, the reply mode text is hidden.
There is an indicator around the chat message, but it is not very visible and intuitive.

One could enlarge the text input field and pin the message at the bottom of the chat.

Leave it like it is and maybe just even more bold/aggressive colors

MatthiasReumann:
### Motivation and Context
- Resolves #925 

### Description
- Update UI
- Move katex options to function `getKatexOptions` to avoid duplication

### Steps for Testing
Prerequisites:
- 1 Account
- 1 Livestream with chat messages

1. Log in
2. Navigate to a Livestream
3. Reply to chat messages
4. 🚀

### Screenshots
https://user-images.githubusercontent.com/29633518/220309280-6dc82740-742d-4bca-88aa-003ef783ffee.mov

### DIFF
--- a/web/ts/chat/NewChatMessage.ts
+++ b/web/ts/chat/NewChatMessage.ts
@@ -1,17 +1,16 @@
 import { sendMessage } from "../watch";
 import { getCurrentWordPositions } from "./misc";
+import { ChatMessage } from "./Chat";
 
 export class NewChatMessage {
     message: string;
-    replyTo: number;
+    reply: Reply;
     anonymous: boolean;
     addressedTo: ChatUser[];
 
     constructor() {
-        this.message = "";
-        this.replyTo = 0;
         this.anonymous = false;
-        this.addressedTo = [];
+        this.clear();
     }
 
     send(): void {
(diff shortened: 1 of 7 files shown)


web/ts/chat/NewChatMessage.ts
```diff
@@ -21,10 +20,25 @@ export class NewChatMessage {
...
     }
 
+    setReply(m: ChatMessage) {
+        this.reply = {
+            message: m,
+            id: m.ID,
+        };
+    }
+
+    cancelReply() {
```
alexanderstephan, line 31: I think the `cancelReply()` logic might be broken.
There seem to be some ways to get undefined state.

This can be triggered in two ways.
1. If escape or the cross symbol is pressed.
2. After a reply is sent.

If another message is sent now in the normal chat, the error occurs.
Error: `Uncaught TypeError: e.reply is undefined`

alexanderstephan (dismissed):
I absolutely love the UI! 🚀 

I only found a bug, which I mentioned in the code review itself.
I tried to reproduce it with two different setups, so I hope it is reproducible on your side and not a setup problem on my end.

Besides just a note: Was `.idea/sqldialects.xml` deleted on purpose?

MatthiasReumann:
> I absolutely love the UI! 🚀
> 
> I only found a bug, which I mentioned in the code review itself. I tried to reproduce it with two different setups, so I hope it is reproducible on your side and not a setup problem on my end.
> 
> Besides just a note: Was `.idea/sqldialects.xml` deleted on purpose?

Thanks for the feedback! Will look into it.

No, `.idea/sqldialects.xml` wasn't deleted on purpose.

joschahenningsen (approved):

PR #934, opened 2023-02-23
fix saml sp selection

joschahenningsen:
### Motivation and Context

#932 didn't work quite as expected because the /shib endpoint rejects requests for the wrong domains when using a default sp to parse it.

### Description

This PR adds a list of service providers from which the right one is selected for each request depending on the request host. 

### Steps for Testing

Code review.


MatthiasReumann (approved):
LGTM 👍🏻

PR #935, opened 2023-02-26
Added id to sort_order_button

Valentin-Metz:
### Motivation and Context
Enables automated tools to determine sort order of the page.
### Description
All this should do is add a unique id to the button, as finding it by XPATH is apparently not reliable.

### Screenshots
[Button is displaying the text "DESC" here]:

image


joschahenningsen (approved):
👌

Issue #796, opened 2022-11-22
Subtitles


joschahenningsen:
This feature tracks the progress of implementing auto-generated subtitles and a related search functionality.

Related: https://github.com/TUM-Dev/TUM-Live-Voice-Service

joschahenningsen:
Status: We already have a functioning micro-service that does the transcribing. The back-end functionality to call the micro-service and handle the data is still missing. Also the front-end wasn't adapted yet to support .srt subtitles but this should be easy. 

We might want to consider using whisper instead of vosk in the future (thanks to the hackatum team "Wordless"!) but this can be tracked in a separate issue at https://github.com/TUM-Dev/TUM-Live-Voice-Service

Drawbacks: whisper needs much more resources so we might want to make it configurable which service to use.

simon300000:
(commenting here because it maybe more a management problem)

Is it possible to distribute the workload when using whisper?

Some students have gaming computer/laptop with dedicated GPU, is it possible that we can “crowdfund” computing resources? So volunteer opens an executable which pulls audio from Tum-live and upload the SRT file.

To motivate people we might have some kind of leaderboard or even gifts at end of each semester, I think many would like to help.

CommanderStorm:
Thanks for the offer, but this won't be necessary. (Furthermore, I am not convinced that this would be an option in terms of data-protection and reliability)
We have enough compute (the BBB nodes are currently not as busy as they used to 😉). This would be via CPUs, which are slower than real-time (not good, but manageable as lectures are not streamed 24/7).
RBG is currently also looking into buying GPUs.

This will definitely happen, but it will possibly take a few weeks ;) 
@joschahenningsen Correct me if I am distributing outdated info 😉

joschahenningsen:
> This will definitely happen, but it will possibly take a few weeks ;)

Exactly, I'll get a few powerful servers but it will take some time as we're a little understaffed right now

PR #955, opened 2023-03-05
Search through subtitles

joschahenningsen:
### Motivation and Context
This PR implements searching through the subtitles of an individual stream, allowing students to quickly find a topic in them. 
Course-wide search could be a follow up.

Related to #796 

### Description

This PR adds
- A cron job that exports all subtitles to meilisearch once a day
- a search bar that calls a new api which internally asks meilisearch for related results
- a popup on the watch page that displays the results and allows jumping to the timestamp of the subtitle

### Steps for Testing

Prerequisites:
- 1 vod with subtitles
- a running meilisearch instance
```
docker run -it --rm \
    -p 7700:7700 \
    getmeili/meilisearch:v0.30 \
```
- that is configured in `config.yml`
```yaml
meili:
  host: http://localhost:7700
  apiKey: MASTER_KEY
```

1. Log in
2. Navigate to `/admin/maintenance` and run the meili export cron job
3. navigate to vod and test the search functionality 

### Screenshots

https://user-images.githubusercontent.com/44805696/222977320-f8e8ee8f-119a-4e56-8d70-82d6500c4a22.mp4

### DIFF
--- a/api/search.go
+++ b/api/search.go
@@ -13,6 +13,9 @@ import (
 func configGinSearchRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 	routes := searchRoutes{daoWrapper}
 	router.GET("/api/search/streams", routes.searchStreams)
+	g := router.Group("/api/search/subtitles")
+	g.Use(tools.InitStream(daoWrapper))
+	g.GET("/:streamID", routes.searchSubtitles)
 }
 
 type searchRoutes struct {
@@ -64,4 +67,10 @@ func (r searchRoutes) searchStreams(c *gin.Context) {
 		"duration": endTime.Sub(startTime).Milliseconds(),
 		"results":  response,
 	})
+}
+
+func (r searchRoutes) searchSubtitles(c *gin.Context) {
+	s := c.MustGet("TUMLiveContext").(tools.TUMLiveContext).Stream
+	q := c.Query("q")
+	c.JSON(http.StatusOK, tools.SearchSubtitles(q, s.ID))
 }
--- a/api/search.go
+++ b/api/search.go
@@ -12,10 +12,13 @@ import (
 
 func configGinSearchRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 	routes := searchRoutes{daoWrapper}
-	router.GET("/api/search/streams", routes.searchStreams)
-	g := router.Group("/api/search/subtitles")
-	g.Use(tools.InitStream(daoWrapper))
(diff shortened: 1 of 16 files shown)


mono424:
This unbelievably sexy!!!!!!

api/search.go
```diff
@@ -13,6 +13,9 @@ import (
 func configGinSearchRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 	routes := searchRoutes{daoWrapper}
 	router.GET("/api/search/streams", routes.searchStreams)
+	g := router.Group("/api/search/subtitles")
```
MatthiasReumann: Minor:
We could theoretically group it as `/api/search` and then `.GET("/streams")` 
and `.GET("subtitles")`. Makes the code cleaner, I guess.

web/ts/TUMLiveVjs.ts
```diff
@@ -584,6 +584,14 @@ export function jumpTo(hours: number, minutes: number, seconds: number) {
     }
 }
 
+export function jumpToMs(ms: number) {
```
MatthiasReumann: Could you maybe merge the exisiting `jumpTo` function with this one by 
changing the parameter to an object? e.g.:

```typescript 
export function jumpTo(setting: object) {
  for (let j = 0; j < players.length; j++) {
         players[j].ready(() => {
           if ('ms' in object) { ... }
           else { ... }
       }
    }
}
```

MatthiasReumann (changes requested):
I'm not sure how but you read my mind, I was about to suggest that feature next Wednesday! 
Works great, Looks great 💯


What's your take on this: Do you think we could already implement a TUM-Live "complete" search using Meili? e.g. for the _new_ start page.

Added some minor request changes.

api/search.go
joschahenningsen: Done :ok: Tho I was thinking about removing the ocr search considering the quality of the subtitles, this is something we can discuss on Wednesday :)

web/ts/TUMLiveVjs.ts
joschahenningsen: Yup, good catch :+1:

joschahenningsen:
> Do you think we could already implement a TUM-Live "complete" search using Meili? e.g. for the new start page.

Definitely a feature I want! But currently not really feasible in my opinion until we have more subtitles. If a user had a search now they would always only find results in the one lecture with subtitles which might be confusing.

MatthiasReumann (approved):
Great thanks ⚡

PR #960, opened 2023-03-19
Data Store

mono424:
### Motivation and Context
Loading data from the API and storing it was very entangled with the controllers for the UI. This makes it harder if different components need to load the same data resulting in loading sections multiple times. This could also cause inconsistencies.

### Description
A global store that stores the most up-to-date data, and can be subscribed to by different components, would solve the issues. This keeps the current data state consistent across all components that are using it. Also using a customizable cache that minimizes fetches.

### Steps for Testing
Prerequisites:
- 1 Admin
- 1 Livestream

1. Log in
2. Navigate to a Course Admin panel
3. Check if section editing works as expected
4. Go to a live stream that has sections
5. See if sections are displayed correctly
6. Add / Edit / Remove Bookmarks and chek if UI responds correctly

### Screenshots
Bildschirmfoto 2023-03-30 um 00 27 01


mono424:
@MatthiasReumann @joschahenningsen Please tell me wether you like the approach or not, its more or less just a proposal to make the data more accessible in the frontend ✌️

mono424:
Added the possibility of a cache layer behind the providers, currently disabled, probably super unnecessary, could remove it again to reduce complexity

web/ts/bookmarks.ts
```diff
@@ -19,17 +21,12 @@ export class BookmarkList {
...
-        });
+        await DataStore.bookmarks.delete(this.streamId, id);
     }
 
     async fetch() {
-        this.list = (await Bookmarks.get(this.streamId)) ?? [];
-        this.list.forEach((b) => {
-            b.update = updateBookmark;
-            b.friendlyTimestamp = new Time(b.hours, b.minutes, b.seconds).toString();
+        await DataStore.bookmarks.subscribe(this.streamId, (data) => {
```
MatthiasReumann: If I add a new bookmark, the bookmark list on the watch page isn't updated and the callback `(data) => {}` is executed many times. Interestingly the more bookmarks exist the often the callback is called 😅

This simple test shows that
```typescript
await DataStore.bookmarks.subscribe(this.streamId, (data) => {
    console.log(data);
    this.list = data;
});
```

`data` holds only the old bookmarks without the new one.

web/ts/video-sections.ts
```diff
@@ -132,11 +131,8 @@ export class VideoSectionsAdmin {
     }
 
     async fetch() {
-        VideoSections.get(this.streamId).then((list) => {
-            this.existingSections = list;
-            list.forEach(
-                (s) => (s.friendlyTimestamp = new Time(s.startHours, s.startMinutes, s.startSeconds).toString()),
-            );
+        await DataStore.videoSections.subscribe(this.streamId, (data) => {
```
MatthiasReumann: This callback does also only hold the old data, however, it only is called once.

MatthiasReumann (changes requested):
Awesome! Not re-loading content all the time in TS will make the page much more response 🤩

I noticed however a bit of a bug as described in the comments. Once fixed I think this PR is good to go.

web/ts/bookmarks.ts
MatthiasReumann: Updating and deleting bookmarks still works fine btw.

mono424:
@MatthiasReumann As you already know I ran into trouble with AlpineJS calling `init()`-Function twice. I fixed it now with some singleton like behaviour... :S

Further, I found an issue with Course Admin Page. You cannot edit or delete Chapters, but I have the same behavior in`dev`-Branch. Should be fixed in a follow up i guess :)

MatthiasReumann (approved):
Seems to be fine now. Thanks! 💯

PR #963, opened 2023-03-28
Get thumbnails for livestreams 

alexanderstephan:
### Description
Implements https://github.com/joschahenningsen/TUM-Live/issues/957

A cron job is created that fetches thumbnails every minute and writes them to `tmp` as a jpeg with the stream id as the file name. By using an admin token, workers should have access to the HLS stream.
I annotated the worker Docker files with `amd64` so the `docker-compose` setup works on ARM macs (tesseract has issues with ARM) such that this PR was easier to test for me.

### Steps for Testing
1. Use local `docker-compose` setup.
2. Go to the maintenance page in the settings and run `fetchLivePreviews`.
3. Observe that the log has no errors and the file is created.


alexanderstephan:
@MatthiasReumann This should already work. Live thumbnails are written to `tmp` every other minute. Does this format work for you?
It should also fix the generation of section previews. I would be glad if you could give it a try!

api/worker_grpc.go
```diff
@@ -778,6 +780,73 @@ func notifyWorkersPremieres(daoWrapper dao.DaoWrapper) {
...
+		HLSUrl:   s.PlaylistUrl,
+	}
+	resp, err := client.GenerateLiveThumbs(context.Background(), &req)
+	if err != nil {
+		log.WithError(err).Warn("Can't generate live preview")
+		sentry.CaptureException(err)
+		return err
+	}
+
+	file, err := os.Create(filepath.Join("tmp", fmt.Sprintf("%d.jpeg", s.ID)))
```
joschahenningsen: `os.TempDir()` returns the temporary directory depending on the os the server runs on. Wo knows which weird ppl might run Tum live on windows sometime 😀

alexanderstephan: Ah, very cool!

worker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20-alpine3.17 as builder
+FROM amd64/golang:1.20-alpine3.17 as builder
```
joschahenningsen, line 1: Why is this change necessary?

worker/cfg/cfg.go
```diff
@@ -46,6 +47,7 @@ func SetConfig() {
 	LrzUploadUrl = os.Getenv("LrzUploadUrl")
 	MainBase = os.Getenv("MainBase")             // eg. live.mm.rbg.tum.de
 	VodURLTemplate = os.Getenv("VodURLTemplate") // eg. https://stream.lrz.de/vod/_definst_/mp4:tum/RBG/%s.mp4/playlist.m3u8
+	AdminToken = os.Getenv("AdminToken")
```
joschahenningsen: I would have TUM-Live construct a valid playlist url (with jwt or admin token) and send it to the worker. This makes the worker less dependent on the type of the video (e.g. is it hosted by us or the lrz? Does it need the jwt?)

api/worker_grpc.go
```diff
@@ -778,6 +780,70 @@ func notifyWorkersPremieres(daoWrapper dao.DaoWrapper) {
...
+}
+
+func getLivePreviewFromWorker(s *model.Stream, workerID string, client pb.ToWorkerClient) error {
+	req := pb.LiveThumbRequest{
+		WorkerID: workerID,
+		HLSUrl:   s.PlaylistUrl,
+	}
+	resp, err := client.GenerateLiveThumbs(context.Background(), &req)
+	if err != nil {
+		log.WithError(err).Error("Can't retrieve raw live preview data from worker")
```
joschahenningsen: return `fmt.Errorf("something: %w", err)` and only log further up the call stack.

joschahenningsen (changes requested):
Looks great! Just a few nits :)

worker/Dockerfile
alexanderstephan: Tesseract apparently does not build on ARM. I added this prefix to make this explicit. I can also remove it since it's technically redundant (if the platform is specified in `docker-compose.yaml`).

worker/cfg/cfg.go
alexanderstephan: I think it needs the JWT, the edge server does not really care where request comes from at the moment.
But I agree, avoiding this dependency in the worker is better. Will change that.

worker/api/api_server.go
```diff
@@ -73,6 +74,26 @@ func (s server) GenerateThumbnails(ctx context.Context, request *pb.GenerateThum
 	return &pb.Status{Ok: true}, nil
 }
 
+// GenerateLiveThumbs generates a preview image of the most recent stream state.
+func (s server) GenerateLivePreview(ctx context.Context, request *pb.LivePreviewRequest) (*pb.LivePreviewResponse, error) {
+	if request.WorkerID != cfg.WorkerID {
+		log.Info("Rejected request to generate live thumbnails")
+		return nil, errors.New("unauthenticated: wrong worker id")
+	}
+	cmd := exec.Command("sh", "-c",
```
MatthiasReumann, line 83: Do you perhaps know the file size and resolution of the resulting image?

joschahenningsen (approved):
LGTM as long as this works for @MatthiasReumann 🍻

joschahenningsen (approved):

Issue #539, opened 2022-06-21
Eliminate hardcoded file paths


MatthiasReumann:
**Is your feature request related to a problem? Please describe.**

1. Almost all paths are hardcoded into endpoints. This introduces code duplication and potentially errors.
2. It always takes some time to figure out what the correct path to store files is. This is a bit tedious and can be avoided.

**Describe the solution you'd like**
```golang 
type PathProvider interface {
    Provide(...) string // returns path
    ....
}
```

Concrete `PathProvider`'s could store all the additional information needed to create the path. For example, every course could have a `CoursePathProvider` and `.provide()` would return the folder in which all the files for the course will be stored. 

**Additional context**
Backwards-Compatibility because of inconsistent path names is a big issue here. However, I am not sure if and how many inconsistent paths exist in the first place. One might need to investigate that before implementing it.

Mjaethers:
I would like to work on this for Open Source Lab.
Could it be assigned to me?

joschahenningsen:
Sounds great @Mjaethers! Fyi, hopefully, the worker will be rewritten until Oct. 1st. So don't put to much effort into that just yet :)

alexanderstephan:
@Mjaethers The new worker code is available at https://github.com/TUM-Dev/gocast/tree/enh/runner. Please try to make it fit there. If in doubt, maybe contact @joschahenningsen again.

PR #964, opened 2023-03-28
Redesign of / page

### LINKED ISSUE #597: Start page redesign
The start page lags a bit behind in terms of styling. I like that it is so minimal, but it could be a bit more visually grouped and better fit in with the other design reworks (e.g. use of containers with small borders, different font styles, ...)

As this is a rather important rework, we can collect ideas here. After all it is the thing that the people see when they first visit the site, so want to make a good impression. :D 

It would be also cool rework the hiding of courses since it is a bit unconsistent. Maybe it can be done similar to pinning courses.

### LINKED ISSUE #737: `Live now` vs `Live`
**Describe the bug**
For indicating to users, that something is live there are two types of iconographys

**To Reproduce**
1. see `Live` (in red bubble) vs `Live now` (without bubble) 
Screenshot_20221021_140036_Chrome

**Expected behavior**
Since both are live, both should get the red bubble

 - OS: linux
 - Browser chrome
 - Version latest

 - Device: Samsung Galaxy S53
 - OS: android
 - Browser chrome
 - Version latest

MatthiasReumann:
### Motivation and Context
Redesign and reimplement the start page using best practices of Alpine, Tailwind, and TypeScript.

### Description
- New UI
- 99% REST. Little to none go-templates (no more `x-data="json.stringify({{data}})"`!)
- TUM-Live CSS classes
- Proposing updated TypeScript folder structure:
    - `views/`: Whole pages (used only once)
    - `components/`: Single UI Elements (used more than once, e.g. a `header` module)
    - `utilities/`: Reusable Classes, Functions, Enums, etc. that can be included in components and views
    - `api/`: API-Wrappers, DTOs, objects
    - `entry/`: Combination of views and components to bundle functionality as close as possible
- Add `tools/pathprovider/pathprovider.go` to reduce code duplication and a place to put present and future file-path stuff.
    - Partially addresses #539  
    
### Help needed 🙋‍♂️
There is probably a better way to built multiple css files than my current approach. I simply appended another `npx` command to the existing one.

```
webpack --config webpack.dev.js & npx tailwindcss build -i assets/css/main.css -o ./assets/css-dist/main.css --watch & npx tailwindcss build -i assets/css/home.css -o ./assets/css-dist/home.css --watch
```

❓Are there any changes needed in the production config (webpack)? 

### Potential follow-ups
I propose the following Follow-Ups in order to decrease the size of this PR:

- Custom thumbnail upload
- Move API objects (`Bookmarks`, `VideoSections`) to separate file in `api/`. 
    - Currently they reside in `data-store/`
    - See: `api/` for examples

### Steps for Testing

Head to `/` and click on everything you can in all the browsers you have installed.


CommanderStorm:
> - [ ] How to create a navigatum link given a lecture hall (id, name, ...)? (@CommanderStorm)

I don't fully know what you mean by this.
When you get to migrating [`tum.live/about`](https://tum.live/about), you can simply copy the links.
Otherwise, I am a bit lost what you are unclear about… 😅

MatthiasReumann:
> > * [ ]  How to create a navigatum link given a lecture hall (id, name, ...)? (@CommanderStorm)
> 
> I don't fully know what you mean by this. When you get to migrating [`tum.live/about`](https://tum.live/about), you can simply copy the links. Otherwise, I am a bit lost what you are unclear about… 😅

Never mind. Error on my side, I guess. 😅 Mapping URLs to lecture halls will just work fine. Thanks!

api/courses.go
```diff
@@ -99,6 +110,261 @@ type uploadVodReq struct {
...
+			sentry.CaptureException(err)
+			_ = c.Error(tools.RequestError{
+				Err:           err,
+				Status:        http.StatusInternalServerError,
+				CustomMessage: "can't retrieve course",
+			})
+		}
+		return
+	}
+	var response Response
```
mono424: Nothing blocking but maybe nicer like that:
```
response := Response{Course: course}
```

And then just:
```
if tumLiveContext.User != nil {
[...]
}
```

model/course.go
```diff
@@ -204,6 +204,19 @@ func (c Course) GetNextLecture() Stream {
 	return earliestLecture
 }
 
+// GetLastLecture returns the most recent lecture of the course
+func (c Course) GetLastLecture() Stream {
```
mono424, line 234: This function needs the streams to be sorted right? I see that `GetCourseForLecturerIdByYearAndTerm` gives back the courses with sorted streams already, but it may be good to mention it in the function description or somewhere to prevent confusion (if someone uses the function in the future but loads the courses differntly)? What do you think? :)

web/router.go
```diff
@@ -58,6 +61,8 @@ func ConfigGinRouter(router *gin.Engine) {
 func configGinStaticRouter(router gin.IRoutes) {
 	router.Static("/public", tools.Cfg.Paths.Static)
 
+	router.Static("/thumbnails/live", pathprovider.TUMLiveTemporary)
```
mono424: I am thinking, are the filenames of the thumbnails unpredictable? Otherwise, you may could see thumbnails of private streams.

mono424:
This looks amazingly clean, tried it locally seemed to work fine!! Love it, I added some comments nothing big tho :)
We could think about using the data store abstraction and also could split API and the data store into separate
folders to have a better separation, maybe `api` and `store` or something? But all that seems to be for a followup, so I am cool with current PR 👍 Can't wait to see it live!! 💯

mono424 (approved):
Looks fine from my side :)

api/courses.go
```diff
@@ -34,6 +36,15 @@ func configGinCourseRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 
 	api := router.Group("/api")
 	{
+		api.GET("/courses/live", routes.getLive)
+		api.GET("/courses/public", routes.getPublic)
+		api.GET("/courses/users", routes.getUsers)
```
joschahenningsen, line 41: Looks great so far! Some of these requests however are leaking sensitive information, e.g. ips of lecture halls. Maybe we can have a look at this together @MatthiasReumann ?

joschahenningsen (changes requested):

model/course.go
MatthiasReumann: Yup. A comment would certainly help. 👍🏻

web/router.go
MatthiasReumann: The filenames are `<streamid>.jpeg`, so quite the contrary 😅 I will add a middleware in order to avoid this. Thanks!

api/courses.go
MatthiasReumann: Great catch. Forgot about that. 
Yes. 👍🏻

web/router.go
MatthiasReumann: I've changed the thumbnail route from `/thumbnails/live` to `/api/stream/:id/thumbs/live` in order to utilize the already existing permission checks for streams.

web/ts/api/courses.ts
```diff
@@ -0,0 +1,136 @@
...
+            case "FMI 00.07.014":
+                return "https://nav.tum.de/room/5607.EG.014";
+            case "room_00_08_038":
+                return "https://nav.tum.de/room/5608.EG.038";
+            case "IRH101":
+                return "https://nav.tum.de/room/5620.01.101";
+            case "IRH102":
+                return "https://nav.tum.de/room/5620.01.102";
+            default:
+                return "#";
```
CommanderStorm: I think this could be a better fallback (with importing `fmt`):
```suggestion
                return fmt.Sprintf("https://nav.tum.de/search?q=%s", this.Name);
```

joschahenningsen: IMO we should add a new field to the `lecture_halls` table, e.g. `string ExternalLink` instead of mapping them in the codebase.

CommanderStorm:
Some nitpicks:
- when no semeser is selected, this part of the UI looks a bit broken, I think selecting the most recent one by default is a save bet:
  image
- [`About`](https://964.test.live.mm.rbg.tum.de/about), [`Data Privacy`](https://964.test.live.mm.rbg.tum.de/privacy), [`Imprint`](https://964.test.live.mm.rbg.tum.de/imprint) and `We ❤️  OpenSource`  lead to an empty page
- These scroll bars are kind of not necessary, but no user will notice given that tum.live has a lot more content in prod
   image

Thanks for the feedback! 💙

1 and 2 are test server related. The _about_, _imprint_, _data privacy_ pages aren't stored in the database. The current semester is actually automatically selected if no Year/Term is specified, however, SS23 isn't in the database either 😅

The scroll bars are unfortunately only an issue on certain operating systems. For example, on MacOS the start page looks like:
image

I will style the scroll bar to be invisible. Thanks for the catch!

api/courses.go
```diff
@@ -99,6 +110,259 @@ type uploadVodReq struct {
...
+	c.JSON(http.StatusOK, livestreams)
+}
+
+func (r coursesRoutes) getPublic(c *gin.Context) {
+	tumLiveContext := c.MustGet("TUMLiveContext").(tools.TUMLiveContext)
+
+	year, term := tum.GetCurrentSemester()
+	year, err := strconv.Atoi(c.DefaultQuery("year", strconv.Itoa(year)))
+	if err != nil {
+		_ = c.Error(tools.RequestError{
```
CommanderStorm, line 172: Just a comment
This whole ["not aborting on errors is a good thing"](https://stackoverflow.com/questions/69948784/how-to-handle-errors-in-gin-middleware) thing Gin is doing looks cursed as heck.
=> is ideomatic should be kept this way, but I find this weird as hack

web/assets/css/home.css
```diff
@@ -0,0 +1,195 @@
...
+    @apply pt-1 lg:pt-0 lg:pl-2
+}
+
+.tum-live-footer {
+    @apply bg-gray-100 dark:bg-secondary-light py-4 px-8;
+}
+
+.tum-live-footer a {
+    @apply text-sm font-light text-5 md:py-0 py-1
+}
```
CommanderStorm, line 225: Not a strong opinion, but I think this could be simplified this way

```suggestion
footer {
    @apply bg-gray-100 dark:bg-secondary-light py-4 px-8;
}

footer a {
    @apply text-sm font-light text-5 md:py-0 py-1
}
```

web/assets/css/home.css
```diff
@@ -0,0 +1,195 @@
...
+
+.tum-live-menu > header {
+    @apply py-2 px-4 border-b dark:border-gray-800
+}
+
+.tum-live-menu > header:has(.close) {
+    @apply flex justify-between items-center
+}
+
+.tum-live-menu > header .close{
```
CommanderStorm:
```suggestion
.tum-live-menu > header .close {
```

web/assets/init.js
```diff
@@ -0,0 +1,59 @@
...
+    const shouldBeDark =
+        localStorage.themeMode === "dark" ||
+        (localStorage.themeMode === "system" && mediaQueryPrefersDarkScheme.matches);
+    if (document.documentElement.classList.contains("dark") !== shouldBeDark) {
+        document.documentElement.classList.toggle("dark");
+    }
+}
+const setThemeMode = (mode) => {
+    localStorage.themeMode = mode;
+};
```
CommanderStorm: I dont see the utility in this function => may be inlined

web/assets/init.js
```diff
@@ -0,0 +1,59 @@
...
+const getThemeMode = () => localStorage.themeMode;
+
+// transition from old dark theme system
+if ("darkTheme" in localStorage) {
+    /* TODO: send notification? */
```
CommanderStorm: I am not a fan of `TODO` comments.
Imo the whole if clause could be dropped completely without a huge loss to the user

web/assets/init.js
```diff
@@ -0,0 +1,59 @@
...
+updateTheme();
+mediaQueryPrefersDarkScheme.addEventListener("change", () => updateTheme());
+
+document.addEventListener("alpine:init", () => {
+    Alpine.store("theme", {
+        init() {
+            this.activeMode = getThemeMode();
+            // show the currently active theme (light/dark) on the switcher when system theme mode is active
+            const updateSystemModeSwitcherIconId = () =>
+                (this.modes.system.faSwitcherIconId = mediaQueryPrefersDarkScheme.matches ? "moon" : "sun");
```
CommanderStorm: Should this be coupled to `this.modes.light.faIconId`/`this.modes.dark.faIconId`

web/assets/init.js
```diff
@@ -0,0 +1,59 @@
...
+            this.activeMode = mode;
+            setThemeMode(mode);
+            updateTheme();
+        },
+        modes: {
+            light: { name: "Light", faIconId: "sun" },
+            dark: { name: "Dark", faIconId: "moon" },
+            system: { name: "System", faIconId: "desktop", faSwitcherIconId: "" },
+        },
+        activeMode: "",
```
CommanderStorm: I think `activeTheme` or better `theme` could be a better naming for this functionality

web/template/home.gohtml
```diff
@@ -0,0 +1,446 @@
+{{- /*gotype: github.com/joschahenningsen/TUM-Live/web.IndexData*/ -}}
+{{$user := .TUMLiveContext.User}}
+{{$currentYear := .CurrentYear}}
+{{$currentTerm := .CurrentTerm}}
```
CommanderStorm: They are unused
```suggestion
```

web/template/home.gohtml
```diff
@@ -0,0 +1,446 @@
...
+    <script src="static/assets/init.js"></script>
+    <script src="/static/assets/ts-dist/home.bundle.js?v={{if .VersionTag}}{{.VersionTag}}{{else}}development{{end}}"></script>
+
+    <link rel="manifest" href="/static/assets/manifest.json">
+    <link href="/static/node_modules/@fortawesome/fontawesome-free/css/all.min.css" rel="stylesheet">
+    <link href="/static/assets/css/icons.css?v={{if .VersionTag}}{{.VersionTag}}{{else}}development{{end}}"
+          rel="stylesheet">
+    <link href="/static/assets/css-dist/home.css?v={{if .VersionTag}}{{.VersionTag}}{{else}}development{{end}}"
+          rel="stylesheet">
+    <style>[x-cloak] {
```
CommanderStorm, line 29: This is here and not with the rest of the css to prevent Flash of Unstyled Content, right?
Otherwise this should be added to the main stylesheet 😉

web/template/home.gohtml
```diff
@@ -0,0 +1,446 @@
...
+                    </template>
+                </article>
+            </article>
+        </template>
+        <template x-if="view === home.Views.Main">
+            <article id="main-view" class="relative min-h-full">
+                {{if .TUMLiveContext.User}}
+                    {{if .TUMLiveContext.User.Name}}
+                        <h1 id="greeting" class="md:text-2xl text-xl font-bold mb-4 px-3">
+                            {{.TUMLiveContext.User.GetPreferredGreeting}} {{.TUMLiveContext.User.GetPreferredName}},
```
CommanderStorm:
```suggestion
                {{if $user}}
                    {{if $user.Name}}
                        <h1 id="greeting" class="md:text-2xl text-xl font-bold mb-4 px-3">
                            {{$user.GetPreferredGreeting}} {{$user.GetPreferredName}},
```

CommanderStorm (commented):
I like this change (I am not qualified, this is why this is ony a comment not an aproval)
I have added some inline comments, but nothing major

web/template/home.gohtml
MatthiasReumann: There is no `main.css `anymore. `home.css` is the only stylesheet provided to the home page.

api/courses.go
MatthiasReumann: Luckily, in this case a `return` statement is missing. 😅 If the error occurs the handler function will return with an `Error` attached to its context, abort accordingly and return 400.

web/assets/init.js
MatthiasReumann: Actually on the new home page no icons are needed anymore. Therefore, we can delete all icon related code.

web/ts/api/courses.ts
MatthiasReumann: Agreed. Should we add the column in this PR or in a follow-up?

web/assets/css/home.css
MatthiasReumann: The problem with `footer` alone is that the `<footer>` tag can be used multiple times for different purposes. (For example as page-footer or menu-footer) Therefore introducing a class makes sense to me.

web/template/home.gohtml
joschahenningsen: Yes, this is in the dom to prevent flickering while the stylesheet loads.

MatthiasReumann: Forget what I said. I thought you were referencing line 30. 🥲

MatthiasReumann:
### Update 24.4.2023

- Server Notifications
- Pagination for recent VODs to reduce massive one time server load (e.g. ~100 thumbnails) and improve responsiveness
- Add `ExternalURL` to lecture hall model

image

web/ts/api/courses.ts
MatthiasReumann: Did it in this one. Ids don't necessarily have to be the same, I guess.

<details>

<summary>Here's the SQL insert statement:</summary>

```sql
update lecture_halls
set external_url = 'https://nav.tum.de/room/5602.EG.001'
where id = 1;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5604.EG.011'
where id = 2;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5606.EG.011'
where id = 3;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5620.01.101'
where id = 4;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5620.01.102'
where id = 5;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5608.EG.038'
where id = 6;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5613.EG.009A'
where id = 7;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5607.EG.014'
where id = 9;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5510.EG.001'
where id = 10;
update lecture_halls
set external_url = 'https://nav.tum.de/room/5510.02.001'
where id = 11;
```
</details>

joschahenningsen (approved):
Amazing, looking forward to seeing this on prod :)

Issue #789, opened 2022-11-16
audio only mode


SiKreuz:
Some people might want to listen to a lecture while they are on their way to the university or just getting from one place to another. Playing the stream in the background consumes a lot of the mobile internet. Instead it would be nice to have an audio-only mode.
This reduces data consumption and increases stability while on the Ubahn or similar.

Issue #882, opened 2023-01-16
Uploading two VoDs with the same start date overrides the older upload


JohannesStoehr:
**Describe the bug**
Uploading multiple VoDs one after the other but using the same start date, leads to all VoDs with that start date showing the same video (the one that was uploaded last)

**To Reproduce**
1. Go to Admin page
2. Upload one VoD with start date x
3. Upload another VoD with start date x
4. View videos of these VoDs using their respective link
5. See the same video twice

**Expected behavior**
The VoDs should show their respective video

**Desktop (please complete the following information):**
 - OS: macOs
 - Browser: Safari
 - Version: 16.2

DominikRemo:
Also encountering this problem on chrome:

https://github.com/user-attachments/assets/7b2f8f2d-432e-45d1-b2dd-4f7818e2e99e

Editing a VoD and re-uploading the mp4 does not help as it overwrites all other VoDs with same start date.

While the videos are overwritten, thumbnails in the edit view are still updated correctly:
![Image](https://github.com/user-attachments/assets/1ac10bc9-a020-48bc-b9d2-c2ebd5c0a748)

OS: MacOS
Browser: Chrome
Version: gocast@v1.4.37

PR #1005, opened 2023-04-24
Add video to audio transcoding

MatthiasReumann:
### Motivation and Context
Addresses #789 

### Description
Add ffmpeg command to transcode video to audio for an audio only watch mode.

### Steps for Testing
- 1 Admin

1. Log in
2. Upload VOD
3. In `cfg.StorageDir` should be a .m4a file next to a video file.


worker/worker/transcode.go
```diff
@@ -194,3 +194,24 @@ func reportProgress(stream *StreamContext, p chan int32) error {
 		}
 	}
 }
+
+func transcodeAudio(ctx *StreamContext) error {
+	S.startTranscodingAudio(ctx.getStreamName())
```
MatthiasReumann, line 199: I'd love to hear you opinions on starting and ending (with defer) inside the "working" instead of inside the handler function. Any drawbacks?

worker/worker/request_handlers.go
```diff
@@ -486,6 +496,16 @@ func (s StreamContext) getTranscodingFileName() string {
 		s.getStreamName())
 }
 
+func (s StreamContext) getAudioTranscodingFileName() string {
+	return fmt.Sprintf("%s/%d/%s/%s/%s/%s.m4a",
+		cfg.StorageDir,
+		s.teachingYear,
+		s.teachingTerm,
+		s.courseSlug,
+		s.startTime.Format("2006-01-02_15-04"),
```
MatthiasReumann: This probably introduces the bug describes in #882 also to audio files. Should we change this to 
```golang
return fmt.Sprintf("%s/%d/%s/%s/%d/%s.m4a",
 		cfg.StorageDir,
 		s.teachingYear,
 		s.teachingTerm,
 		s.courseSlug,
 		s.streamId, 
                s.streamName)
``` 
similar to the large thumbnails for the new start page?

worker/worker/transcode.go
joschahenningsen: Inside the "working" function would actually be the correct way. Currently, the status is only cleared after all jobs are done.

Unrelated; the workload checks are a little overengineered imo, we could also just add 1 for every stream, and it's post-processing steps the worker is handling

worker/worker/request_handlers.go
joschahenningsen: Changing the filename to streamID-Version.mp4/m4a seems like a good idea

worker/worker/transcode.go
MatthiasReumann: Agree. I think we talked about this last week, the worker probably needs a bit of refactoring 😬

joschahenningsen: I think  I'll try to rework the worker

worker/worker/request_handlers.go
```diff
@@ -361,6 +366,11 @@ func HandleUploadRestReq(uploadKey string, localFile string) {
 		}
 	}
 
+	err = transcodeAudio(&c)
+	if err != nil {
+		log.WithError(err).Error("Error transcoding audio")
+	}
```
joschahenningsen, line 400: Depending on your use case with the audio, is uploading to the edge server for streaming required?

worker/worker/request_handlers.go
```diff
@@ -241,6 +241,11 @@ func HandleStreamRequest(request *pb.StreamRequest) {
 	S.endTranscoding(streamCtx.getStreamName())
 	notifyTranscodingDone(streamCtx)
 
+	err = transcodeAudio(streamCtx)
```
joschahenningsen: I think we should only extract the audio of the "COMB" stream, otherwise we'll end up with 3 audio tracks of the same lecture.

MatthiasReumann: Agree.

worker/worker/request_handlers.go
MatthiasReumann: Eventually audio files should be served the same way as video files are. So, I guess yes? Maybe we can discuss this in the next meeting.

Of course we still need to implemented a proper audio player page on the front end (Ultimately I don't think reusing the watch page as audio page is a good idea 😅)

worker/worker/request_handlers.go
MatthiasReumann: Is there a better way of achieving this besides an if-statement? e.g.:
```golang
if streamCtx.streamVersion == "COMB" {
...
}
```

joschahenningsen: For now, I think we can go with this. When we refactor the worker, we can implement a way to figure out which version of the stream is the main one.

MatthiasReumann: I think it would be quite cool to specify a custom "pipeline"  (a custom function) for each version of a stream. That way we could potentially add new versions quite easily.

joschahenningsen (approved):
I think when the conflict is resolved this can be merged

PR #1009, opened 2023-04-26
Only show "generate subtitles" button on vods

joschahenningsen:
### Motivation and Context

Currently, there is a button to generate subtitles on lectures without associated videos. This of course does not work.

### Description

The button is now only visible on recordings.

### Screenshots

image


MatthiasReumann (approved):
Thanks!

PR #1015, opened 2023-04-29
fix lecture creation

joschahenningsen:
### Motivation and Context

Currently, lectures can not be created because the file input field is marked as required even when not uploading a vod.

### Description

Now, the required attribute is only set for the fields required for the creation method.

### Steps for Testing

Prerequisites:
- 1 Lecturer

1. create a lecture and verify that the form can be submitted.


alexanderstephan (approved):
Works! 👍

PR #1016, opened 2023-05-01
`Bug` Today's lecture list

MatthiasReumann:
### Motivation and Context
Because of a wrong Javascript method, the "today list" depicts lectures that go live on the same day in a week instead of the same day in a month. 

### Description
- Use correct javascript method.
- Add some padding-bottom to make the next button more visible on mobile browsers.
### Steps for Testing
- Account

1. Log in
2. `/new`
3. Check if the today's lecture list still works

### DIFF
--- a/web/template/home.gohtml
+++ b/web/template/home.gohtml
@@ -278,7 +278,7 @@
             </article>
         </template>
         <template x-if="view === home.Views.Main">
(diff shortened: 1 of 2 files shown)


joschahenningsen (approved):
Good catch!

PR #1017, opened 2023-05-01
fix lecture hall ical

joschahenningsen:
### Motivation and Context

There was a simple typo that is now fixed. Also, the sql statement had an error where the date range with all lecturehalls was ignored.

### DIFF
--- a/tools/config.go
+++ b/tools/config.go
@@ -38,7 +38,7 @@ func initConfig() {
 	viper.AddConfigPath(".")
 	err := viper.ReadInConfig()
 	if err != nil {
-		if err == err.(viper.ConfigFileNotFoundError) {
+		if errors.Is(err, viper.ConfigFileNotFoundError{}) {
 			log.WithError(err).Warn("tools.config.LoadConfig: can't find config file")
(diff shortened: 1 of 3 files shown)


tools/config.go
```diff
@@ -38,7 +38,7 @@ func initConfig() {
 	viper.AddConfigPath(".")
 	err := viper.ReadInConfig()
 	if err != nil {
-		if err == err.(viper.ConfigFileNotFoundError) {
+		if errors.Is(err, viper.ConfigFileNotFoundError{}) {
```
MatthiasReumann, line 41: Is it necessary to instantiate `viper.ConfigFileNotFoundError` here? E.g.: `viper.ConfigFileNotFoundError` vs `viper.ConfigFileNotFoundError{}`

MatthiasReumann (approved):
Thanks! LGTM 🚀

tools/config.go
joschahenningsen: Unfortunately yes, viper doesn't expose the error as a var here, only as a type for some reason

PR #1027, opened 2023-05-09
1006 upload camera lecture slides as vod

### LINKED ISSUE #1006: Upload Camera & Lecture slides as VOD
You cannot upload slides and camera for one VOD.

Add the possibility to upload two video for one VOD, Camera and Slides.

-

**Additional context**
-

mono424:
### Motivation and Context
Currently its not possible to add lecture slides and camera videos as dedicated video files.

### Description
Add input fields to allow the upload of camera and slides video too. Because the upload formular becomes a little
user unfriendly, it is refactored. Also, a jump to new added lectures and is introduced together with a little marker
that marks the new added lectures in the list.

### Steps for Testing
Prerequisites:
- 1 Admin
- 1 Course

1. Log in
2. Navigate to a Course Admin Page
3. Upload a new VoD with dedicated camera and slides video
4. Check if all videos are working
5. Upload a new VoD with combined stream only
6. Check if the VoD with combined stream is working

### Screenshots

https://github.com/joschahenningsen/TUM-Live/assets/17506411/7052c833-3b1d-4a25-acbe-c602929107d2


web/template/partial/course/manage/create-lecture-form.gohtml
```diff
@@ -1,127 +1,216 @@
...
-                <span class="text-sm text-5">End</span>
-                <span x-show="formData.formatedDuration !== ''" class="text-sm font-light opacity-75"
-                      x-text="`( ${formData.formatedDuration} )`"></span>
-                <input class="tl-input" name="end" placeholder="16:00" id="end" x-model="formData.end" x-ref="end"
-                       x-init="flatpickr($refs.end, {enableTime: true, noCalendar: true, dateFormat: 'H:i', time_24hr: true, allowInput: true })"
-                       @change="() => setTimeout(() => recalculateDuration(), 100)"
-                       :required="!formData.vodup"/>
-            </label>
-        </template>
+            <div x-show="currentTab === 0" class="flex py-4">
```
joschahenningsen, line 41: These should probably be templates to avoid weird behavior like the date picker being visible when clicking back

web/template/partial/course/manage/create-lecture-form.gohtml
```diff
@@ -1,127 +1,216 @@
...
+                <ul class="w-full">
+                    <li class="mb-2">
+                        <button
+                            @click="() => updateType(false)"
```
joschahenningsen: IMO it would be nice if the page automatically switched to the second page without clicking continue here.

web/ts/edit-course.ts
```diff
@@ -525,13 +556,76 @@ export function createLectureForm(args: { s: [] }) {
...
+                    this.canContinue = this.formData.title.length > 0 && this.formData.start.length > 0;
+                } else {
+                    this.canGoBack = true;
+                    this.onLastSlide = true;
+                    this.canContinue =
+                        this.formData.title.length > 0 &&
+                        this.formData.start.length > 0 &&
+                        this.formData.end.length > 0;
+                }
+                return;
```
joschahenningsen: This could use some comments

web/ts/edit-course.ts
```diff
@@ -525,13 +556,76 @@ export function createLectureForm(args: { s: [] }) {
...
+                if (this.formData.vodup) {
+                    this.canGoBack = true;
+                    this.onLastSlide = false;
+                    this.canContinue = this.formData.title.length > 0 && this.formData.start.length > 0;
+                } else {
+                    this.canGoBack = true;
+                    this.onLastSlide = true;
+                    this.canContinue =
+                        this.formData.title.length > 0 &&
+                        this.formData.start.length > 0 &&
```
joschahenningsen: A title should not be required

web/ts/edit-course.ts
```diff
@@ -525,13 +556,76 @@ export function createLectureForm(args: { s: [] }) {
...
+                this.canGoBack = false;
+                this.onLastSlide = false;
+                return;
+            }
+
+            if (this.currentTab === 1) {
+                if (this.formData.vodup) {
+                    this.canGoBack = true;
+                    this.onLastSlide = false;
+                    this.canContinue = this.formData.title.length > 0 && this.formData.start.length > 0;
```
joschahenningsen: A title should not be required

web/ts/edit-course.ts
```diff
@@ -620,44 +727,93 @@ export function createLectureForm(args: { s: [] }) {
...
-                if (!e.lengthComputable) {
-                    return;
-                }
-                window.dispatchEvent(
-                    new CustomEvent("voduploadprogress", { detail: Math.floor(100 * (e.loaded / e.total)) }),
-                );
+
+        async uploadVod(): Promise<number> {
+            const uploadPres = this.formData.presFile[0] != null;
+            const uploadCam = this.formData.camFile[0] != null;
```
joschahenningsen: COMB should not be required, the only requirement should be `at least one file is present`

web/ts/edit-course.ts
```diff
@@ -620,44 +727,93 @@ export function createLectureForm(args: { s: [] }) {
...
+                    },
+                );
+
+            if (uploadCam)
+                await uploadFilePost(
+                    `/api/course/${this.courseID}/uploadVODMedia?streamID=${streamID}&videoType=CAM`,
+                    this.formData.camFile[0],
+                    (progress) => {
+                        uploadProgress = { ...uploadProgress, CAM: progress };
+                        window.dispatchEvent(new CustomEvent("voduploadprogress", { detail: uploadProgress }));
```
joschahenningsen: Can you refactor this to its own method to avoid duplication?

api/courses.go
```diff
@@ -363,7 +369,78 @@ func (r coursesRoutes) uploadVOD(c *gin.Context) {
...
+		req.URL.Scheme = u.Scheme
+		req.URL.Host = u.Host
+		req.Host = u.Host
+		req.URL.Path = u.Path
+		req.URL.RawQuery = u.RawQuery
+	}
+	p.ServeHTTP(c.Writer, c.Request)
+}
+
+func (r coursesRoutes) uploadVODMedia(c *gin.Context) {
```
joschahenningsen, line 412: Why do we need another handler? Can you elaborate & add a comment what this is used for?

api/courses.go
```diff
@@ -1113,6 +1190,12 @@ func (r coursesRoutes) createLecture(c *gin.Context) {
 	seriesIdentifier := uuid.NewV4().String()
 	req.DateSeries = append(req.DateSeries, req.Start)
 
+	// To get new ids after insert
+	var oldStreamIds []uint
+	for _, stream := range tumLiveContext.Course.Streams {
+		oldStreamIds = append(oldStreamIds, stream.ID)
+	}
+
```
joschahenningsen, line 1199: What exactly does this do?

model/upload-key.go
```diff
@@ -2,11 +2,24 @@ package model
...
+const (
+	VideoTypeCombined     VideoType = "COMB"
+	VideoTypePresentation           = "PRES"
+	VideoTypeCamera                 = "CAM"
+)
+
+func (v VideoType) Valid() bool {
+	return v == VideoTypeCombined || v == VideoTypePresentation || v == VideoTypeCamera
+}
+
```
joschahenningsen, line 16: Nice way of doing this! Unsure about this: did you check whether the stream.go model already has a similar enum?

web/template/partial/course/manage/create-lecture-form.gohtml
```diff
@@ -1,127 +1,216 @@
...
-                </option>
-                {{range .LectureHalls}}
-                    <option value="{{.Model.ID}}" class="text-gray-300">
-                        {{.Name}}
-                    </option>
-                {{end}}
-            </select>
-        </label>
+        <div class="form-container-body">
+            <div x-show="currentTab === -1" x-data="{ progress: { COMB: 0, PRES: null, CAM: null } }" @voduploadprogress.window="(e) => (progress = e.detail)">
```
joschahenningsen, line 32: The tabs could potentially be in their own templates if that makes things a little easier to understand

joschahenningsen (changes requested):
Amazing! Left a few thoughts :) :clinking_glasses:

web/template/partial/course/manage/create-lecture-form.gohtml
mono424: @joschahenningsen I tried templates, but the date picker looses the state on going back then, I added `.prevent` to the next and back button. I think that resolved the issue you are experiencing with the date picker :)

web/ts/edit-course.ts
mono424: not sure how to do it in a clean way in a function, because I have the `uploadProgress` state that is caried through the calls, but I abstracted in a way that I have an array of mediaUploads now and iterate through, so the `uploadFilePost`-Part is there only once. 👍

api/courses.go
mono424: My intention was to create an endpoint that lets a user upload VOD Media to an existing VOD without the need of recreating the VOD. This could be used to edit the video on demand media files later too.

api/courses.go
mono424: I save the old ids of the streams of a course, to find the newly created stream ids after the insert. I use them to mark as newly created in the frontend.
There could not figure out how to get back the newly created ids from `CoursesDao.UpdateCourse`. Is there a way to easily get the newly created stream ids from `CoursesDao.UpdateCourse` ?

model/upload-key.go
mono424: Searched for such an enum, but I did not find any :(

mono424:
@joschahenningsen I did some major changes due to your feedback, I hope it makes it simpler. Summary:
I renamed `uploadVOD` to `createVOD` and removed the video upload logic. All videos are not uploaded over the dedicated endpoint `uploadVODMedia`. Now the upload logic only has to be in one of the two endpoints. Also the worker does not have to return the streamID but it can be directly returned from `createVOD`.
Downside, you could theoretically create a vod without any media if you sent the request manually to the API. But thats probably not to bad.

joschahenningsen (approved):
Looks good! Thanks :clinking_glasses:

Issue #889, opened 2023-01-19
Distributed chat


joschahenningsen:
**Is your feature request related to a problem? Please describe.**
We can't currently scale TUM-Live horizontally by replicating it to a number of instances. This is mainly because the chat channels are in memory of a instance. If two users were to connect to two separate instances their chat message would never reach the other instance. This also prevents us from doing rolling updates to avoid downtime.

**Describe the solution you'd like**
We need to introduce sharding of the chat functionality that allows a TUM Live instance to notify other instances about events or develop a chat gateway that users connect to and have the gateway distribute the events. 

Here is an interesting write-up on how twitch handles this: https://blog.twitch.tv/en/2015/12/18/twitch-engineering-an-introduction-and-overview-a23917b71a25#chat

gedw99:
Use nats and Choria ?

Nats can replicate data and chroria uses nats to replicate binaries

gedw99:
It can also keep slides in sync

gedw99:
https://github.com/gabstv/melody is WS based . Nats does WS too . 

I use it with protobufs just like GRPC. Unlike GRPC you get automatic load balancing and failover . No setup required.

joschahenningsen:
nats looks like a nice fit, will investigate whether it fits our needs. Thanks! :+1: 
> chroria uses nats to replicate binaries

Likely not needed in our case, we orchestrate our service with docker swarm.

We already use melody for user facing websocket stuff. Could be good for server <-> server communication as well, we were concerned tho because it didn't seem very well maintained.

gedw99:
Nats can do user facing layer too.

The AAA ( accounts, authentication, authorisation) is extensive too

PR #1037, opened 2023-05-17
Live viewer count

### LINKED ISSUE #1026: Show the number of live viewers next to livestreams

MatthiasReumann:
### Motivation and Context
Resolves #1026 

### Steps for Testing
- 1 Livestream

1. Log in
2. `/new`
3. View count should be visible.

### Screenshots
image

### DIFF
--- a/api/courses.go
+++ b/api/courses.go
@@ -124,7 +124,7 @@ func (r coursesRoutes) getLive(c *gin.Context) {
 		Course      model.CourseDTO
 		Stream      model.StreamDTO
 		LectureHall *model.LectureHallDTO
-		Stat        model.StatDTO
+		Viewers     uint
 	}
 
 	livestreams := make([]CourseStream, 0)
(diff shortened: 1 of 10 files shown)


api/courses.go
```diff
@@ -154,10 +155,14 @@ func (r coursesRoutes) getLive(c *gin.Context) {
 				lectureHall = &lh
 			}
 		}
+
+		var stat model.Stat
+		stat, err = r.StatisticsDao.GetStat(stream.ID)
```
joschahenningsen: We should be able to just get this number from memory. The length of the chat session contains the number of viewers: https://github.com/joschahenningsen/TUM-Live/blob/a1c9209f5a875eab2967035d742ecab0b2e56fac/api/chat.go#LL661C43-L661C43

Would this be easier to implement and more efficient?

MatthiasReumann: Definitely. 👍🏻 However, we need to make sure to update this if we ever shard and distribute the chat as in #889.

joschahenningsen (approved):
Can you regenerate the mocks? Then this looks good to go!

PR #1043, opened 2023-05-22
Show user courses for lecturers

MatthiasReumann:
### Motivation and Context
Currently, account which are registered as `lecturers` don't see the courses they participate in (i.e. their user courses) on the new start page.

### Description
Add missing function call. 

### Steps for Testing
- 1 Lecturer
- 1 Course that he doesn't administer 

1. Log in
2. Course should be listed under **`user courses`**
3. 🕺


joschahenningsen (approved):
Nice, thanks!

Issue #1030, opened 2023-05-10
option to set default playback speed in Settings


MariusAlbrecht:
**Is your feature request related to a problem? Please describe.**
playback speeds always reset to 1 every new stream

**Describe the solution you'd like**
basically the title

**Additional context**
#544 somewhat related

MatthiasReumann:
Resolved in #1047.

PR #1047, opened 2023-05-29
Add `live_rate` parameter and add dynamic rate adjusting

### LINKED ISSUE #544: More playback speed options
Up to 2x, playback speed can be increased in increments of 0.25, but above that it can only be increased in increments of 0.5.
This seems a bit odd, since these larger playback speeds are exactly where you'd want to have the most granularity: the perceived difference between 2x and 2.5x is significantly bigger than between 1.5x and 2x, so it would be nice to have an intermediate step there.

The settings should allow users to enable more playback speeds, such as e.g. 2.25x.
(The best solution would probably be to simply have 0.25 increments all the way up to 3x or 3.5x.)

Don't do anything and keep the current behaviour. This can already be worked around via external browser extensions.

**Additional context**
n/a

### LINKED ISSUE #1031: Changing views resets the playback speed to 1
**Describe the bug**
basically just the title

**To Reproduce**
1. open any stream (or vod probably, didnt test)
2. Set any playback speed other than 1
3. change the view (e.g. from combined to split view)
4. The playback speed will now have reset back to 1

**Expected behavior**
keep the playback speed which was set before

**Screenshots**
n/a

 - Windows 11 Home 22H2 build 22621.1635
 - Firefox 112.0.2 (64-bit)

didnt test, can do if important

**Additional context**
potentially make this toggle-able in the settings as this isn't clearly a bug (but more like unexpected behaviour for me personally)

MatthiasReumann:
### Motivation and Context
- Resolves #1031 and #1030 by adding the key `live_rate` to the LocalStorage.
- Resolves #544 by adding query parameter `?rate={float >= 0}`. This way anyone can adjust the playback-speed to his/her desired value 😬
- Add `PlayerSettings` object to tidy up player settings setting and storing.

### Steps for Testing
- 1 Lecturer
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. Add `?rate=2.3` to query (Playback speed should now be 2.3)
4. Add `?t=1337` to query (Video should jump forwards)
5. Mute, Set Volume, Set Playback speed manually via the video menu. Reload. (Settings should now be the same as before)
6. Any combination of 3, 4, & 5
7. 🌟

You can also check the browser console for those values since I've added some useful logging messages.

### DIFF
--- a/web/ts/TUMLiveVjs.ts
+++ b/web/ts/TUMLiveVjs.ts
@@ -20,6 +20,120 @@ export function getPlayers(): VideoJsPlayer[] {
     return players;
 }
 
+class PlayerSettings {
+    private readonly player: VideoJsPlayer;
+    private readonly isLive: boolean;
+    private readonly isEmbedded: boolean;
+
+    constructor(player: VideoJsPlayer, isLive: boolean, isEmbedded: boolean) {
+        this.player = player;
+        this.isLive = isLive;
+        this.isEmbedded = isEmbedded;
+    }
+
+    initTrackbars(streamID: number) {
+        loadAndSetTrackbars(this.player, streamID);
+    }
+
+    initAirPlay() {
+        // @ts-ignore
+        this.player.airPlay({ addButtonToControlBar: true, buttonPositionIndex: -2 });
+    }
+
+    setVolume() {
+        const volume: number = +PlayerSettings.getFromStorage("volume") ?? this.player.volume();
+        this.player.volume(volume);
+        console.log(`⚫️ set volume: ${volume}`);
+    }
+
+    setMuted() {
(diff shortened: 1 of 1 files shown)


mono424 (approved):
Love how clean it is now. I assume the `console.log()` are intended to stay?
Works well ! 👍

MatthiasReumann:
> Love how clean it is now. I assume the `console.log()` are intended to stay?
> Works well ! 👍

Yes. I think this kind of logging might make it easier to report bugs to us in a replicable way 😅.

mono424:
> > Love how clean it is now. I assume the `console.log()` are intended to stay?
> > Works well ! 👍
> 
> Yes. I think this kind of logging might make it easier to report bugs to us in a replicable way 😅.

Sounds great for me! :)

joschahenningsen (approved):
Thanks!

PR #1050, opened 2023-05-30
Preload next `N` thumbnails on new start page

MatthiasReumann:
### Motivation and Context
Currently we load images, i.e. thumbnails, simply with `img(...)` via CSS. Since we also paginate most content (to avoid floods of image loading) most thumbnails "pop up" after loading (grey default thumbnail → image thumbnail). This can be circumvented by prefetching the next `N` images via javascript.

### Description
Add a `preloader` callback function to the `Paginator` class and call the callback on `.next()`.

### Steps for Testing
- 1 Lecturer
- 2 Students
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. ...

### DIFF
--- a/web/template/home.gohtml
+++ b/web/template/home.gohtml
@@ -655,7 +655,8 @@
                                                 <template x-if="course.NextLecture.IsComingUp">
                                                     <span>
                                                         <span>Next lecture starts in </span><!--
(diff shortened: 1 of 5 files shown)


joschahenningsen (approved):
Nice!

PR #1056, opened 2023-06-06
fix redirect after login

### LINKED ISSUE #1053: Remember what link was clicked on before having to login
I sometimes transfer a link from my computer to a lecture I want to watch on my phone. I am of course greeted by the Login-Page. After successful login, I land on the homepage -> "https://live.rbg.tum.de/" and not on the page I wanted to initially visit.

A lot of services do this, such that they keep track of the link that was clicked on before the login (how? I don't know). For example Moodle. If you click on a link from an E-Mail and are not yet logged in into moodle, it transfers me to the login page, but after successful login it directly transfers me to the original link that was in the E-Mail.
I'd love a solution like this! After login don't put me on the homepage but on the lecture (link) that I wanted to visit, so I do not have to go back and copy/open the link again.

There is no alternative :(

**Additional context**
Super duper ultra mega please implement this. Will make life easier!

joschahenningsen:
### Motivation and Context

The redirect after logins is currently broken. This fixes #1053 

### Description

The redirect target cookie is now properly parsed and set when visiting the login page as opposed to when submitting the login form. This fixes issues with saml logins.

### DIFF
--- a/web/user.go
+++ b/web/user.go
@@ -70,8 +70,7 @@ func HandleValidLogin(c *gin.Context, data *tools.SessionData) {
 }
 
 func getRedirectUrl(c *gin.Context) (*url.URL, error) {
-	ret := c.Request.FormValue("return")
-	ref := c.Request.FormValue("ref")
+	ret := c.Query("return")
 	if ret != "" {
 		red, err := url.QueryUnescape(ret)
 		if err == nil {
@@ -79,11 +78,11 @@ func getRedirectUrl(c *gin.Context) (*url.URL, error) {
 		}
 	}
 
-	if ref == "" {
+	if ret == "" {
 		return url.Parse("/")
 	}
 
-	return url.Parse(ref)
+	return url.Parse(ret)
 }
 
 // loginWithUserCredentials Try to login with non-tum credentials
--- a/web/user.go
+++ b/web/user.go
@@ -34,15 +34,6 @@ func (r mainRoutes) LoginHandler(c *gin.Context) {
 	username := c.Request.FormValue("username")
 	password := c.Request.FormValue("password")
 
-	redirURL := getRedirectUrl(c)
-
-	// Only set cookie if (potentially) needed.
-	if !strings.HasSuffix(redirURL, "/") && !strings.HasSuffix(redirURL, "/login") {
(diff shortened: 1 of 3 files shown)


web/user.go
```diff
@@ -133,6 +124,19 @@ func loginWithTumCredentials(username, password string, usersDao dao.UsersDao) (
 }
 
 func (r mainRoutes) LoginPage(c *gin.Context) {
+	redirUrlStr := "/"
+	redirURL, err := getRedirectUrl(c)
```
MatthiasReumann, line 127: I think this will always be `/` since 
```golang
ret := c.Request.FormValue("return") == ""
ref := c.Request.FormValue("ref") == ""
```
since these are only sent on form.submit:
```html
<form x-data="{ref: global.getLoginReferrer()}" id="loginForm" method="post" class="grid gap-3 py-4 px-5">
                <input x-model="ref" type="hidden" name="ref">
                ....
</form>
```

Maybe I'm also wrong? 🤔

joschahenningsen: It worked when I tested it, but I think we're overcomplicating it here frontend wise. I'll rework it a little.

MatthiasReumann: Does the TUM shibboleth allow a [relay state](https://stackoverflow.com/questions/34350160/what-is-exactly-relaystate-parameter-used-in-sso-ex-saml)? Because if so, we might want to make use of this feature.

joschahenningsen: That's a great impulse, will check it out!

joschahenningsen: Our SAML library already uses the relay state to track requests, hacking in the referrer won't work without significant changes. Also, the relay state can only store up to 80 bytes. Therefore, I decided to stick with our cookie workaround which seems to be an established way of doing this.

joschahenningsen:
I removed the frontend logic of using the `document.referrer` for this case, as it only worked when the browser set it and when using the login form. Now, the referrer cookie is set upon loading the login page when the login page has the `?return=...` query. (Which is the case when a page was accessed that requires login but no session is present)

alexanderstephan (approved):
I can't really test this right now, but the changes look good to me. I think the cookie workaround is perfectly sane.

PR #1058, opened 2023-06-07
Add mail queue

joschahenningsen:
### Motivation and Context

To avoid getting rate limited and to allow retrying failed email sending, we need to store the emails we want to send and send them out buffered. 

### Description

This PR adds a mail table to the database which contains all emails that should be sent, these are handled one by one with a delay that can be configured.


MatthiasReumann (approved):
LGTM 👍 

I guess there isn't an easy way to dry-run this functionality on a large scale (many mails) locally, so we will see if any complaints arise in the future 😬

PR #1061, opened 2023-06-11
Fix VOD not working if path != /vod/

mono424:
### Motivation and Context
Previously, the VOD (Video on Demand) feature was only accessible through the specific VOD-Path `/vod`. If any other path was used, the content delivery edge would respond with a 404 error. This limitation prevented users from customizing their VOD paths according to their preferences. The motivation behind the recent changes was to resolve this issue and allow developers to specify custom VOD paths.

### Description
- Previously, if a path other than `/vod` was used, the system responded with a 404 error. This has been fixed, and the system will now recognize and handle custom VOD paths appropriately.
- JWT Check was only working if the path was `/vod`. Now it's working for custom paths as well.

### Steps for Testing

Prerequisites:
- 1 Lecturer
- 1 Course

1. Log in
2. Navigate to a course management page
3. Create a new Video-on-Demand
4. Check if the Video-on-Demand is accessible
5. **Change the VoD-Path on your local environment** ()
6. Create a new Video-on-Demand
7. Check if the Video-on-Demand is accessible

### DIFF
--- a/worker/edge/edge.go
+++ b/worker/edge/edge.go
@@ -183,7 +183,7 @@ func validateToken(w http.ResponseWriter, r *http.Request, download bool) (claim
 	}
 
 	urlParts := strings.Split(allowedPlaylist.Path, "/")
-	allowedPath := vodPath + "/" + strings.Join(urlParts[2:len(urlParts)-1], "/")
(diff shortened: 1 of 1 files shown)


joschahenningsen (approved):
Nice!

PR #1062, opened 2023-06-12
Unit tests

MatthiasReumann:
### Motivation and Context
In the last few months we've added a lot of new functionality without implementing associated unit tests. This PR addresses this issue partly (Not all functions can be tested).

### Description
Add unit tests.

### Steps for Testing
`go test / test` workflow on GitHub succeeds

### DIFF
--- a/api/progress.go
+++ b/api/progress.go
@@ -176,16 +176,7 @@ func (r progressRoutes) markWatched(c *gin.Context) {
 }
 
 func (r progressRoutes) getProgressBatch(c *gin.Context) {
-	ctx, exists := c.Get("TUMLiveContext")
-	if !exists {
-		_ = c.Error(tools.RequestError{
-			Status:        http.StatusBadRequest,
-			CustomMessage: "context should exist but doesn't",
-		})
-		return
-	}
-
-	tumLiveContext := ctx.(tools.TUMLiveContext)
+	tumLiveContext := c.MustGet("TUMLiveContext").(tools.TUMLiveContext)
 	if tumLiveContext.User == nil {
 		_ = c.Error(tools.RequestError{
 			Status:        http.StatusForbidden,
--- a/api/progress.go
+++ b/api/progress.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"errors"
-	"fmt"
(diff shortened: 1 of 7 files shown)


api/progress.go
```diff
@@ -176,8 +177,16 @@ func (r progressRoutes) markWatched(c *gin.Context) {
 }
 
 func (r progressRoutes) getProgressBatch(c *gin.Context) {
-	tumLiveContext := c.MustGet("TUMLiveContext").(tools.TUMLiveContext)
+	ctx, exists := c.Get("TUMLiveContext")
```
joschahenningsen: Was there an issue with the `MustGet` approach?

api/progress.go
```diff
@@ -196,13 +205,15 @@ func (r progressRoutes) getProgressBatch(c *gin.Context) {
...
-	for i, stringId := range stringIds {
+	ids := make([]uint, 0, len(stringIds))
+	for _, stringId := range stringIds {
+		fmt.Println(stringId)
 		id, err := strconv.Atoi(stringId)
 		if err != nil {
 			continue
 		}
-		ids[i] = uint(id)
+		fmt.Println("ok")
```
joschahenningsen:
```suggestion
```

api/progress.go
```diff
@@ -196,13 +205,15 @@ func (r progressRoutes) getProgressBatch(c *gin.Context) {
 		return
 	}
 
-	ids := make([]uint, len(stringIds))
-	for i, stringId := range stringIds {
+	ids := make([]uint, 0, len(stringIds))
+	for _, stringId := range stringIds {
+		fmt.Println(stringId)
```
joschahenningsen:
```suggestion
```

joschahenningsen (changes requested):
Great, thanks! Just a quick hint

api/progress.go
MatthiasReumann: For the progress API we don't use a middleware as `tools.InitCourse(daoWrapper)` and therefore the handler panics every time the context doesn't exists instead of returning `400`.

joschahenningsen (approved):

PR #1064, opened 2023-06-13
`bug` Delete Chat Replies on Delete

MatthiasReumann:
### Motivation and Context
Currently if a message with replies is deleted only the "base" message is deleted, i.e. replies are kept in the database. In other words `deleted_at` is only not-null for base messages. As a consequence, each time we fetch messages deleted "ghost" replies are transmitted also. 

This PR fixes this issue.

### Description
Add `DeleteReplies` to `ChatDao` and call in delete handler.

### Steps for Testing
- 1 Account
- 1 Livestream with the chat activated

1. Log in
2. Navigate to a Livestream
3. Write a message and reply to that message 
4. Delete the base message
5. Use DataGrip (or similar) to check that both the base message and the replies are deleted


mono424 (approved):
Checked locally with DataGrip, looks great! 🚀

joschahenningsen (commented):
Looks good! Do we ever need to delete a chat without deleting it's replies? If not we can also merge these methods into one.

MatthiasReumann:
> Looks good! Do we ever need to delete a chat without deleting it's replies? If not we can also merge these methods into one.

Probably not. Good idea! 👍🏻

MatthiasReumann:
Please ignore the `go test / test` error. We need to update the workflow to `go1.20` (#1065) 😬

dao/chat.go
```diff
@@ -164,7 +164,9 @@ func (d chatDao) RetractChat(id uint) error {
 
 // DeleteChat removes a chat with the given id from the database.
 func (d chatDao) DeleteChat(id uint) error {
-	return DB.Model(&model.Chat{}).Delete(&model.Chat{}, id).Error
+	return errors.Join(
+		DB.Model(&model.Chat{}).Delete(&model.Chat{}, id).Error,
+		DB.Delete(&model.Chat{}, "reply_to = ?", id).Error)
```
joschahenningsen: This seems a little cleaner

```suggestion
	return DB.Where("id = ? OR reply_to = ?", id, id).Delete(&model.Chat{}).Error
```

joschahenningsen (changes requested):

joschahenningsen (approved):

PR #1067, opened 2023-06-14
ad-hoc streams

### LINKED ISSUE #1059: Start Ad Hoc streams
A lecturer should be able to simply/quickly start a stream without having to schedule it. A button e.g. Start ad-hoc stream -> until x would be nice.

joschahenningsen:
### Motivation and Context

fixes #1059 

### Description

The create-course interface now has a new tab that allows selecting the type of livestreams: Ad hoc or scheduled. 
If the stream is set to ad hoc, the start time of the stream will be set to `now() + 2 min` and workers will immediately be notified to start due streams.

### DIFF
--- a/api/courses.go
+++ b/api/courses.go
@@ -1218,6 +1218,9 @@ func (r coursesRoutes) createLecture(c *gin.Context) {
 			Recording:     req.Vodup,
 			Premiere:      req.Premiere,
 		}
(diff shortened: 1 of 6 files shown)


MatthiasReumann (approved):
Works and looks great. Thanks! 🚀

PR #1069, opened 2023-06-20
allow resetting passwords for internal accounts

joschahenningsen:
### Motivation and Context

We occasionally have tickets from people who need to reset their passwords. Currently we have to handle them by hand.

### Description

This PR adds an interface to the login form, that allows resetting the password. Users will then get an email with a link where they can set a new password.

### Steps for Testing

Prerequisites:
- 1 user in the database with a valid email address

1. Navigate to Log in
2. Click reset password and enter the email of the user. 
3. Take a look at the `emails` table in the database. It should contain a message for the user with a working link to reset the password.


MatthiasReumann (approved):
Can't complain about anything! Works well in both cases, i.e. the provided account exists and if it doesn't exist. 🚀 

Also: Great consideration to not send a  _"This account doesn't exist"_ info message.

PR #1071, opened 2023-06-28
Chat component

### LINKED ISSUE #926: Inconsistent Scrolling Behavior
**Describe the bug**
The livestream chat is only scrolled down when there is a new message and the corresponding blue banner is pressed.

**To Reproduce**
1. Go to a livestream, e.g., https://live.rbg.tum.de/w/GBS/26257?dvr right now
2. See that the chat is at the uppermost position and does not scroll by default

**Expected behavior**
Livestreams should enable auto scroll by default

MatthiasReumann:
### Motivation and Context
Throwing window events and catching them works fine for a few, but with an ever increasing amount of chat-features it becomes tedious (sometimes close to impossible) to maintain and improve the chat.

This PR implements (similar to the new start page) chat and poll components which handle all necessary functions without a single event thrown. (Ok to be truthfully, there is one exception thrown, namely `wsrealtimeconnectionchange` 🥶)

I know, the PR is rather big, but let me try to summarize the changes in a handful of paragraphs, such that the review is less cumbersome. 

### Description
The basic building blocks can be seen in the following schematic:

test drawio

The context for both chat and poll are supplied with either a `ChatWebsocketConnection` or `PollWebsocketConnection` object. Both of which take the static object `SocketConnections.ws` (The actual socket connection) as a constructor parameter. What previously has been achieved with a rather large function in `watch.ts` is now split into smaller feature-grouped (i.e. the chat and poll) objects.

In the parent's `x-data` directive, a `videoInteractionContext` object is created which initializes the web socket connection. This component is similar to the `skeleton` object of the new start page. A kind of meta object so to say. 

The chat input now also resides in a separate component called `chatPromptContext`. Again all messaging functionality (Emojis, @-ing, Replying) is grouped in this component. In order to communicate a press on a __reply__ button in the `chatContext` a `Tunnel` is used. A `Tunnel` is basically a observer pattern communication channel between multiple components and is implemented using the already existing `ValueStream` class (This is actually also used for the `Pin` feature on the new start page.

The chat replay feature is also fully integrated into the `chatContext`. One difference between the implementation is that the chat replay now doesn't work in the pop out chat anymore. I don't see any clean way to communicate between different browser windows besides events. I am sure users will forgive us for this tiny inconvenience. 

The websocket messages for `viewers` and the `description` now resides in the `videoInformationContext`. 

To minimize the footprint of the pop-out chat page, all chat/poll functionality is now webpack-bundled as `interaction.bundle.js`. 

HTML & CSS-wise aren't a lot of changes. I however, re-used classes of the new start page to facilitate a consistent UI experience. One noticeable change however, is that all HTML code now resides in `template/components/chat.gohtml` and not scattered across many different files. Nevertheless, I've rearranged some buttons as well as fields, and added some dropdown menus.

Other than that, a lot of unused functions, classes, and files have either been moved somewhere else or removed.

### Steps for Testing
Prerequisites:
- `1` Admin
- `n` Students
- `1` Livestream

1. Log in
2. Navigate to a Livestream
3. Chat, Create Polls, Delete, Retract, Resolve, Reply, Use the chat replay

### UI

https://github.com/joschahenningsen/TUM-Live/assets/29633518/3185e722-ea43-40e2-821b-1a6bc7195f60


api/chat.go
```diff
@@ -397,10 +406,10 @@ func (r chatRoutes) handleRetract(ctx tools.TUMLiveContext, msg []byte) {
 func (r chatRoutes) handleMessage(ctx tools.TUMLiveContext, context *realtime.Context, msg []byte) {
 	var chat chatReq
 	if err := json.Unmarshal(msg, &chat); err != nil {
-		log.WithError(err).Error("error unmarshaling chat message")
+		log.WithError(err).Error("error unmarshalling chat message")
 		return
 	}
-	if !ctx.Course.ChatEnabled {
+	if !ctx.Course.ChatEnabled && !ctx.Stream.ChatEnabled {
```
MatthiasReumann, line 412: I think this is a bug in the production code, right? Currently, users can't chat if `course.ChatEnabled == false` and `stream.ChatEnabled == true`.

joschahenningsen: Yes, looks like a bug

joschahenningsen (approved):
Looks great, can be merged when conflicts are resolved imo

PR #1072, opened 2023-06-28
use dedicated array for channels

mono424:
### Motivation and Context
If handler has been attached before subscribing to a channel, the channel was subscribed twice.
That's because it took a look at the handlers after the connection, assuming that there has been a subscription
for every handler before.

### Description
Changed to use an extra `subscribedChannels` Array to keep track of the subs on client side to resub on reconnect.

### Steps for Testing
Prerequisites:
- 1 Student
- 1 Livestream

1. Log in
2. Navigate to a Livestream with a Chat
3. Use the chat and verify if it works

### DIFF
--- a/web/ts/socket.ts
+++ b/web/ts/socket.ts
@@ -13,6 +13,7 @@ export class Realtime {
     private debugging = false;
     private ws: WebSocket;
     private handler: object = {};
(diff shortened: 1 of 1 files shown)


joschahenningsen (approved):
Looks good, thanks for the quick fix!

PR #1079, opened 2023-07-11
copy admins when copying course

joschahenningsen:
### Motivation and Context

When copying a course, we want to retain the admins of the course. Likely one of them copied the course in the first place.

### Description

When copying a course we iterate over the admins of the previous course and insert them into the database for the new course

### Steps for Testing

Prerequisites:
- 1 course
- 1 lectuer

1. Make sure you are logged in as a lecturer and copy a course you are admin of. 
2. Check that you are now admin of the newly created course.

### DIFF
--- a/api/courses.go
+++ b/api/courses.go
@@ -1527,6 +1527,11 @@ func (r coursesRoutes) copyCourse(c *gin.Context) {
 	course := tlctx.Course
 	streams := course.Streams
 
+	admins, err := r.DaoWrapper.CoursesDao.GetCourseAdmins(course.ID)
+	if err != nil {
+		log.WithError(err).Error("Error getting course admins")
+		admins = []model.User{}
+	}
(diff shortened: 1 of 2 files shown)


MatthiasReumann (approved):
Looks good to me. Thanks 👍🏻 

I wonder if a feature to "shallow"-copy VODs of a previous semester to another would make sense. For example, if something fails (worker, etc.) dramatically. Could potentially also save some storage space because re-uploads from previous semesters would become redundant 🤔

joschahenningsen:
> I wonder if a feature to "shallow"-copy VODs of a previous semester to another would make sense. 

Yes, great idea! Will work on that :)

alexanderstephan:
I just was about to create an issue for this since I noticed it for the "C-Vorkurs". Thanks for the fix!

PR #1085, opened 2023-07-17
Show public, own and pinned courses without selecting semester

joschahenningsen:
### Motivation and Context

This resolves an issue where the start page does not show the courses on the left when visiting it without specifying the year and term.

### Description

This issue occurs because the expression to set the current year resolves like this:
`this.year = +url.searchParams.get("year") ?? undefined;`
-> `this.year = +null ?? undefined;`
-> `this.year = 0 ?? undefined;`
-> `this.year = 0;`

### Steps for testing

I created a public course on the testserver. It should be visible without selecting a teaching term first.

### DIFF
--- a/web/ts/views/home.ts
+++ b/web/ts/views/home.ts
@@ -170,7 +170,10 @@ class PageState {
     constructor(url: URL) {
         this.url = url;
         this.term = url.searchParams.get("term") ?? undefined;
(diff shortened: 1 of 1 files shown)


MatthiasReumann (approved):
Thanks! 🚀 

U gotta love javascript sometimes 🫠

PR #1092, opened 2023-07-26
display lecture time in vod title, if stream name not present

simon300000:
### Motivation and Context

blank title with `:` looks bad

### Description

I put `Lecture: D M, Y` there

### Steps for Testing

Prerequisites:
- Web Browser

1. Navigate to a VOD without name
2. notice the title is different

### Screenshots
屏幕截图 2023-07-26 211027


joschahenningsen (approved):
Good catch, thanks for the fix! :beers:

PR #1125, opened 2023-08-17
Video Section Admin Bugs

### LINKED ISSUE #1123: Deletion of video sections broken
**Describe the bug**
When have created a video section, it is not possible to delete it anymore.
It fails with `HTTP/2 404 Not Found`. This happens both on Firefox and on Chromium.

**To Reproduce**
1. Go to to any VoD

**Expected behavior**
The deletion requests suceeds, the deleted video section disappears.

 - OS: macOS 13.4.1 (c) 
 - Browser: Firefox
 - Version: 116.0.2 (64-bit)

### LINKED ISSUE #1124: Input validation for video section timestamps not working on Firefox
**Describe the bug**
As the timestamp input requires type `number`, it should not allow characters as input.

**Note**: This works under Chrome as expected, but not on Firefox.

**To Reproduce**
1. Go to VoDs in admin page
2. Go to video sections
3. Insert e.g. an 'A' for minutes
4. Add and save the video section
5. Observe behavior as in the screenshot blow

**Expected behavior**
Either there should be warning when adding the section or it should not be possible to insert anything in input.

**Screenshots**
image

 - OS: macOS 13.4.1 (c) 
 - Browser: Firefox
 - Version: 116.0.2 (64-bit)

MatthiasReumann:
### Motivation and Context
- Resolves #1123 (Allow the deletion of video sections for which no thumbnail exists)
- Resolves #1124 (Firefox still allows letters in `<input type="number">` fields. Manual validation via regex necessary)

### Steps for Testing
- 1 Admin
- 1 Course
- Firefox

1. Log in
2. Navigate to Admin Course Page
3. Type a letter into one of the number fields. Shouldn't work anymore.
4. Create Video Section
5. Publish Video Section
6. Delete Video Section 
7. 🎉

### DIFF
--- a/api/stream.go
+++ b/api/stream.go
@@ -571,17 +571,6 @@ func (r streamRoutes) deleteVideoSection(c *gin.Context) {
 		return
 	}
 
-	file, err := r.FileDao.GetFileById(fmt.Sprintf("%d", old.FileID))
-	if err != nil {
-		log.WithError(err).Error("can not find file")
-		_ = c.Error(tools.RequestError{
-			Status:        http.StatusNotFound,
-			CustomMessage: "can not find file",
-			Err:           err,
-		})
-		return
-	}
-
 	err = r.VideoSectionDao.Delete(uint(id))
 	if err != nil {
 		log.WithError(err).Error("can not delete video-section")
@@ -593,12 +582,15 @@ func (r streamRoutes) deleteVideoSection(c *gin.Context) {
 		return
 	}
 
-	go func() {
-		err := DeleteVideoSectionImage(r.DaoWrapper.WorkerDao, file.Path)
-		if err != nil {
-			log.WithError(err).Error("failed to generate video section images")
-		}
-	}()
+	file, err := r.FileDao.GetFileById(fmt.Sprintf("%d", old.FileID))
+	if err == nil {
+		go func() {
+			err := DeleteVideoSectionImage(r.DaoWrapper.WorkerDao, file.Path)
+			if err != nil {
+				log.WithError(err).Error("failed to generate video section images")
+			}
+		}()
+	}
 
 	c.Status(http.StatusAccepted)
 }
--- a/api/stream.go
+++ b/api/stream.go
@@ -583,7 +583,15 @@ func (r streamRoutes) deleteVideoSection(c *gin.Context) {
 	}
 
 	file, err := r.FileDao.GetFileById(fmt.Sprintf("%d", old.FileID))
-	if err == nil {
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+		log.WithError(err).Error("can not get video section thumbnail file")
+		_ = c.Error(tools.RequestError{
+			Status:        http.StatusInternalServerError,
+			CustomMessage: "can not get video section thumbnail file",
+			Err:           err,
+		})
+		return
+	} else {
 		go func() {
 			err := DeleteVideoSectionImage(r.DaoWrapper.WorkerDao, file.Path)
 			if err != nil {
--- a/web/router.go
+++ b/web/router.go
@@ -33,7 +33,6 @@ var templatePaths = []string{
 	"template/partial/*.gohtml",
 	"template/partial/stream/*.gohtml",
 	"template/partial/course/manage/*.gohtml",
-	"template/partial/stream/chat/*.gohtml",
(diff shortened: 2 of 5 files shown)


api/stream.go
```diff
@@ -593,12 +582,15 @@ func (r streamRoutes) deleteVideoSection(c *gin.Context) {
 		return
 	}
 
-	go func() {
-		err := DeleteVideoSectionImage(r.DaoWrapper.WorkerDao, file.Path)
-		if err != nil {
-			log.WithError(err).Error("failed to generate video section images")
-		}
-	}()
+	file, err := r.FileDao.GetFileById(fmt.Sprintf("%d", old.FileID))
```
alexanderstephan, line 585: IMO first, the error returned by `GetFileById` should be checked (and handled, e.g., with `log.WithError`). Then, `DeleteVideoSectionImage` should be executed when there is `file != nil`.
Maybe we could also use a boolean, and then have an `ok` check.
Overall, I could see this being a bit error prone and possibly ignoring other errors that `GetFileById` could return.

api/stream.go
```diff
@@ -593,12 +582,15 @@ func (r streamRoutes) deleteVideoSection(c *gin.Context) {
...
+	if err == nil {
+		go func() {
```
alexanderstephan, line 595: Just curious: Why is `DeleteVideoSectionImage` non-blocking BTW? Is it too slow otherwise?

web/template/partial/course/manage/edit-video-sections.gohtml
```diff
@@ -36,21 +35,24 @@
                 <span class="font-light">Unsaved Changes</span>
             </div>
         </header>
-        <form @submit.prevent="videoSectionController.publishNewSections()">
+        <form id = "new-section-form" @submit.prevent="videoSectionController.publishNewSections()">
             <div class="flex align-middle">
                 <input id="startHours"
+                       onkeyup="this.value=this.value.replace(/[^\d]/,'')"
```
alexanderstephan, line 41: Maybe we could also filter out values not in the range. E.g., valid hours are 0 <= x <= 23. I noticed that I can still put in any number. This would need to be done for the specific ranges of `startHours`,  `startMinutes` and `startSeconds`. Could be something for a follow-up as well, but I guess it also kinda fits the PR scope.

web/router.go
```diff
@@ -33,7 +33,6 @@ var templatePaths = []string{
 	"template/partial/*.gohtml",
 	"template/partial/stream/*.gohtml",
 	"template/partial/course/manage/*.gohtml",
-	"template/partial/stream/chat/*.gohtml",
```
alexanderstephan, line 36: I guess this a follow up to https://github.com/joschahenningsen/TUM-Live/pull/1071?

alexanderstephan (changes requested):
Thanks for getting to these fixes so quickly! 👍  Firefox really seems to be a diva lately. 😅
I have added some suggestions and remarks.

web/template/partial/course/manage/edit-video-sections.gohtml
alexanderstephan: If we really want to overengineer this, we could check whether the timestamp is actually within the duration of the video. But this is future work.

web/router.go
MatthiasReumann: Yes. 👍🏻

api/stream.go
MatthiasReumann: `DeleteVideoSectionImage` is a request to a worker, therefore yes.

api/stream.go
MatthiasReumann: Great input. Added a `!gorm.ErrNotFound` check.

web/template/partial/course/manage/edit-video-sections.gohtml
MatthiasReumann: 👍🏻  

Added 
```javascript
this.current.startMinutes < 32 && // I think 32h hours will be sufficient.
this.current.startMinutes < 60 &&
this.current.startSeconds < 60
```

Regarding the second comment: I think a follow-up PR would make sense.

api/stream.go
alexanderstephan: Ah, yes, that's the perfect fit here.

alexanderstephan (approved):

PR #1132, opened 2023-08-22
fix duration on video thumbnails

joschahenningsen:
### Motivation and Context

The current duration in only approximated based on the planned times of the lectures. Instead, we should use the duration stored in the database to accurately display the video duration.
Additionally, the current frontend renders the time based on the locale of the browser, which leads to weird behavior like `03:00:00 AM`  (see screenshot)

#### Before 
image

#### After 
 image


MatthiasReumann (approved):
Looks great, thanks! 🚀

PR #1135, opened 2023-08-22
Prolong maximum chat message length to 1000 characters

### LINKED ISSUE #778: Increase Chat character limit
I've run into the character limit of 300 characters per chat message fairly frequently. It allows for maybe two to three sentences, which is simply too short to answer any question in depth. I'd consider a limit of at least 1000 characters the bare minimum.

Can we increase it? Can we at least exclude admins from it? Or could we even allow the course admins to set the limit for each course individually?

JosefSchoenberger:
### Motivation and Context
We want longer chat messages! I can't answer to the questions of our students if I'm not able to write a long prose with at least 800 characters! :-P
Resolves #778.

### Description
I replaced the chat input box with a multiline text area and extended the maximum message length from 200 to 1000 characters.

### Steps for Testing
1. Log in
2. Find a stream or VOD you can comment on
3. Write a long text message with more then 200 characters. May include multiple lines with Shift+Enter.

### DIFF
--- a/web/ts/components/chat-prompt.ts
+++ b/web/ts/components/chat-prompt.ts
@@ -21,7 +21,7 @@ export function chatPromptContext(streamId: number): AlpineComponent {
 
         ws: new ChatWebsocketConnection(SocketConnections.ws),
 
-        inputEl: document.getElementById("chat-input") as HTMLInputElement,
+        inputEl: document.getElementById("chat-input") as HTMLTextAreaElement,
         usersEl: document.getElementById("chat-user-list") as HTMLInputElement,
         emojiEl: document.getElementById("chat-emoji-list") as HTMLInputElement,
 
@@ -41,7 +41,10 @@ export function chatPromptContext(streamId: number): AlpineComponent {
             this.users.reset();
         },
 
-        send() {
+        send(event: KeyboardEvent) {
+            if (event.shiftKey) {
+                return;
+            }
             console.log("🌑 send message '", this.message, "'");
             this.ws.sendMessage({
                 msg: this.message,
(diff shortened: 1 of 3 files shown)


web/ts/components/chat-prompt.ts
```diff
@@ -84,6 +87,10 @@ export function chatPromptContext(streamId: number): AlpineComponent {
             } else if (this.emojis.hasSuggestions()) {
                 this.emojiEl.focus();
             }
+
+            // https://stackoverflow.com/a/21079335
+            this.inputEl.style.height = "0px";
+            this.inputEl.style.height = (this.inputEl.scrollHeight) + "px";
```
joschahenningsen:
```suggestion
            this.inputEl.style.height = this.inputEl.scrollHeight + "px";
```

joschahenningsen (changes requested):
Looks reasonable, just a quick remark found by eslint:

joschahenningsen:
BTW, there also is server side validation: https://github.com/joschahenningsen/TUM-Live/blob/dev/model/chat.go#L40

JosefSchoenberger:
Alright, thanks – I was already wondering, but the server accepted messages with ~220 characters and did not find any other occurence of "200" related to chat messages, so I assumed there was no server-side check.
Btw., was there a reason for why the server-side limit was higher?

joschahenningsen:
> Btw., was there a reason for why the server-side limit was higher?

I can't think of any reason why we would have done this on purpose. I assume it was just an oversight.

joschahenningsen (approved):
Thanks!

PR #1136, opened 2023-08-23
load stream progresses in one query

joschahenningsen:
### Motivation and Context

Currently, we load stream progresses in multiple queries for each request. Especially for admins visiting the start page, this is very inefficient. 

### Description
Unify the db query to accept multiple ids and return all progresses. I simulated a slow database connection by connecting to a remote database via vpn and observed the timings of the endpoint `/api/progress/streams`:
- before: 1959.984845 ms 
- after: 48.86526 ms

### DIFF
--- a/api/progress.go
+++ b/api/progress.go
@@ -1,9 +1,11 @@
 package api
 
 import (
+	"errors"
 	"github.com/joschahenningsen/TUM-Live/dao"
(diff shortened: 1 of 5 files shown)


MatthiasReumann (approved):
LGTM 👍🏻 
good idea btw!

PR #1137, opened 2023-08-23
course management data-providers refactor

mono424:
### Motivation and Context
Currently, the course management relies on Go generated template data. This means we need page reloads
and also it is not good practice to have backend and frontend entangled in that way. Further, much very specific code is used to handle the course management tasks, that could be abstracted in cleaner ways. 

### Description
This PR untangles backend and frontend by using data-providers that are introduced earlier and are already used for data-sections and other parts of the app. This gives a standardized abstraction to handle the lecture's client side.
Also, this will prevent useless page reloads and give a better standardized and app-global API to create new lectures, fetch or update them.

### Steps for Testing
- Go to course management Page
- Open a course
- Edit a lecture
- a) Check uploading media
- b) Check changing meta data
- c) Check uploading attachments (Drag and drop Description field)
- Check JS-Console and Network Tab for errors
- Reload and check if data is persistent

### Screenshots

Way shorter UI specific code (edit-course.ts):

difference


web/ts/api/admin-lecture-list.ts
```diff
@@ -0,0 +1,296 @@
...
+    },
+
+    /**
+     * Updates metadata of a lecture.
+     * @param courseId
+     * @param lectureId
+     * @param request
+     */
+    updateMetadata: async function (courseId: number, lectureId: number, request: UpdateLectureMetaRequest) {
+        console.log({
```
MatthiasReumann:
```suggestion
        console.log({
```

web/ts/api/admin-lecture-list.ts
```diff
@@ -0,0 +1,296 @@
...
+        }
+    },
+
+    /**
+     * Distributes the lecture metadata of given lecture to all lectures in its series.
+     * @param courseId
+     * @param lectureId
+     */
+    saveSeriesMetadata: async (courseId: number, lectureId: number) => {
+        return await postData(`/api/course/${courseId}/updateLectureSeries/${lectureId}`);
```
MatthiasReumann: Any reason why not to use `put`, `post`, etc. from `"../utilities/fetch-wrappers"` in this class?

web/ts/api/admin-lecture-list.ts
```diff
@@ -0,0 +1,296 @@
...
+     * @param file
+     * @param listener
+     */
+    uploadAttachmentFile: async (
+        courseId: number,
+        lectureId: number,
+        file: File,
+        listener: UploadFileListener = {},
+    ) => {
+        return await uploadFile(`/api/stream/${lectureId}/files?type=file`, file, listener);
```
MatthiasReumann, line 253: Add-on to previous comment: The `uploadFile` function could then also be moved to `"../utilities/fetch-wrappers"`.

MatthiasReumann (changes requested):
Great. 🚀 Still amazed how you can write such clean TS code. The comments are always great too!

Tested in the docker-compose setup, everything works fine. 

I've added some small remarks.

web/ts/api/admin-lecture-list.ts
mono424: No actually not, I missed that `../utilities/fetch-wrappers` exists. Thanks for the feedback makes the code even shorter because error handling is already in `../utilities/fetch-wrappers` 🚀

web/ts/api/admin-lecture-list.ts
mono424: Good point! Also moved `postFormData` and renamed `UploadFileListener` to `PostFormDataListener`, because previous name was misleading :)

MatthiasReumann (changes requested):
Changes look good. 

Somehow the "Delete Lecture" in the dropdown menu doesn't work anymore. (Console: `ReferenceError: Can't find variable: deleteLecture`) 😅

mono424:
> Changes look good.
> 
> Somehow the "Delete Lecture" in the dropdown menu doesn't work anymore. (Console: `ReferenceError: Can't find variable: deleteLecture`) 😅

thanks for the ping! Should have tested better, `delete` and `deleteSeries` should be fixed now ;)

MatthiasReumann (approved):
Great. Thanks!

mono424:
@joschahenningsen regarding the size of the pr do you want to take a look first before I merge or can I just go ahead and merge it?

joschahenningsen (approved):
Thanks a ton, this is great!

PR #1142, opened 2023-08-24
Move to TUM-Dev: fix go modules, run configurations and workflows

joschahenningsen:


MatthiasReumann (approved):
Changes LGTM. Plus, since it works on the test server I assume everything is fine. 👍🏻

PR #1146, opened 2023-08-28
Subscribe to WS in video-information

MatthiasReumann:
### Motivation and Context
Currently the `viewers` counter and dynamic description update doesn't work since the websocket is not initialized properly. 

### Description
Move `.subscribe` to parent context `videoInformationContext`.

### Steps for Testing
- 1 Account
- 2 Livestream (1 `+Chat` and 1 `-Chat`)

1. Log in
2. Navigate to a livestream  (`+Chat`)
3. Send messages. Should work. 
4. Navigation to another livestream (`-Chat`)
5. Viewers count should increase without reload if you open the livestream in multiple browser windows.
6. 🕺

### DIFF
--- a/web/ts/components/video-information.ts
+++ b/web/ts/components/video-information.ts
@@ -32,7 +32,7 @@ export function videoInformationContext(streamId: number): AlpineComponent {
                     this.handleDescriptionUpdate(data);
                 }
             };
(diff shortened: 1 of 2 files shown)


joschahenningsen (approved):
Thanks for the fix! :beers:

PR #1145, opened 2023-08-28
Fix false order of videos on course-page

MatthiasReumann:
### Motivation and Context
Resolves #1143 

### Description
Negates predicate which determines order of groups, i.e. months.

### Steps for Testing
- 1 Course

1. Log in
2. Navigate to a Course
3. Re-order Elements
    1. `Oldest first`: Oldest month first; Oldest stream first; 
    2. `Newest first`: Newest month first; Newest stream first; 
4. 🌟

PR #1147, opened 2023-08-28
Fix chat

MatthiasReumann:
### Motivation and Context
In #1135 the `textarea` tag hasn't been closed. As a result the chat UI doesn't work properly. Furthermore, it broke the polls. 
(For examples, see testserver of #1145)

### Description
Close tag. 

### Steps for Testing

Prerequisites:
- 1 Account
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. Chat `send` button should say `send` and not `New Poll`. 
4. Polls work. 
5. 🤠


alexanderstephan (approved):
Works again, thanks! I just noticed though that new lines are stripped server side, so the multi-line textbox is not even that useful right now. 🙈

PR #1150, opened 2023-09-05
Fix group order

### LINKED ISSUE #1149: Ordering Problem
Wenn oldest first ausgewählt wird, dann ist die Auflistung July, Juni, Mai, April und wenn newest first dann ist sie umgekehrt. Die Auflistung der Videos passt zu der gewählten Ansicht.

MatthiasReumann:
### Motivation and Context
- Resolves #1149 

### Description
Dictionaries are not really great for ordering (i.e. they don't have a order). Rewrite the `group` function to return a 2-dimensional array (a array of stream groups starting in the same month and year). 

### Steps for Testing
- 1 Course

1. Log in
2. Navigate to the course page
3. Streams should now also be grouped correctly if multiple years occur. (e.g. 2022 December, 2023 January)

### DIFF
--- a/web/template/home.gohtml
+++ b/web/template/home.gohtml
@@ -431,18 +431,18 @@
                             </header>
                             <section>
                                 <template x-if="courseStreams.hasElements()">
-                                    <article class="flex flex-col" :class="{'flex-col-reverse' : !isOldestFirst() }">
+                                    <article class="grid">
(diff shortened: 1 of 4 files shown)


alexanderstephan (approved):
Nice work, thanks! 🎉

PR #1152, opened 2023-09-07
Settings Page UI

MatthiasReumann:
### Motivation and Context
Update settings page UI for consistency with other pages.

### Description
- Update CSS/HTML. 
- Decrease header footprint, i.e. only import files which are absolutely necessary.

### Steps for Testing
- 1 Account

1. Log in
2. Navigate to settings page
3. Update name, playback speed, etc.
4. Should work
5. 🚀

### Screenshots
Screenshot 2023-09-07 at 07 45 38

### DIFF
--- a/web/assets/css/home.css
+++ b/web/assets/css/home.css
@@ -95,6 +95,10 @@
     @apply bg-gray-800 text-white dark:bg-gray-600
 }
 
+.tum-live-button-muted {
+    @apply bg-gray-200 dark:bg-gray-600 hover:bg-gray-300 dark:hover:bg-gray-700
+}
+
 .tum-live-icon-button {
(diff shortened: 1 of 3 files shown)


alexanderstephan:
The UI itself looks really good. However, I noticed that the menu disappears in the upper right corner. Is that intended?

**Reference**:
image

MatthiasReumann:
> The UI itself looks really good. However, I noticed that the menu disappears in the upper right corner. Is that intended?
> 
> **Reference**: image

Yup. Same for the info pages. I thought that it makes sense to minimize the footprint of the go-template (i.e. include `.User` object) in order to make it more easy to migrate to a REST-only approach.

alexanderstephan:
Okay, I see 👍 In any case, people can just use the back button in their browser.

joschahenningsen (approved):
:heart:

PR #1198, opened 2023-10-18
1175 course import dont ask to leave page all the time

DawinYurtseven:
# What was the problem:
The problem was that each time something happened during a Course Import, the browser would ask the user if they wanted to leave the page, which would lead to a refreshed site and no changes saved.

# What caused it:
there where two reasons why this happened.
- The body of the import was a <form> without a submit type function. this makes the browser think that the user didn't save any change while the site would update itself. 
- The addEventListener in the course-import.ts file wasn't removed after step 2, a successful import/ an error.  

# My Fixes:
- turned the <form> body in course-import.gohtml into a <div>
- added the function OnBeforeUnloadHandler which is just the event itself. but I added removeEventListener in the "notify2" event so that when the second step has been done, the user has no more inputs to do, so asking them would be meaningless


SebiWrn:
Maybe you can add back the two comments to preventDefault() and returnValue that have been there before so we know why these two commands are there

SebiWrn (approved):
Well done :+1:

PR #1221, opened 2023-10-26
1185-recent-VODs-section-doesnt-show-pinned-public-courses

### LINKED ISSUE #1185: Recent VODs section doesnt show pinned (public) courses
**Describe the bug**
The Recent VODs section seems to only show private courses associated to the user when the user is logged in and any public courses when the user is not logged in. When logged in that does not include public courses that the user pinned

**To Reproduce**
1. Log in
2. Pin public course
3. Go to homescreen
4. Not see streams from the pinned course in the recent VODs section

**Expected behavior**
Show public pinned courses as well

**Screenshots**
n/a

 - OS: Win 11
 - Browser: Firefox 118.0.1

not tested

**Additional context**
list of recent streams is set here: 
[https://github.com/TUM-Dev/gocast/blob/bad8f2c194b05d039111f0c45de53b2da460eff9/web/ts/components/main.ts#L36](https://github.com/TUM-Dev/gocast/blob/bad8f2c194b05d039111f0c45de53b2da460eff9/web/ts/components/main.ts#L36)

barisgul15:
### Motivation and Context
Issue 1185: Recent VODs section doesnt show pinned (public) courses

### Description
Before the changes, only user courses were listed for the recently VOD section. Now there is a new array for pinned Courses and if it is detected that user has one or more pinned courses, pinned and user courses are merged into a new array (with eliminating duplicates) and this array is used for the recently VOD section. However a user without any course will still be seeing all public courses in recently VOD section whether he/she has pinned any course or not. That can also be changed to see only pinned courses if desired, but since the probability of a user without any course pinning a course is so low, that feature is not implemented. 

### Steps for Testing
1. Log in
2. Pin a public course that is not also listed in my courses. 
3. Navigate to recently VOD section.

### DIFF
--- a/web/ts/components/main.ts
+++ b/web/ts/components/main.ts
@@ -36,8 +36,9 @@ export function mainContext(year: number, term: string) {
                     if (this.userCourses.length > 0) {
                         if (this.pinnedCourses.length > 0) {
                             //
-                            let pinnedOrUserCourses : Course[] = this.userCourses.concat(
-                                this.pinnedCourses.filter((c : Course) => this.userCourses.indexOf(c) < 0));
+                            const pinnedOrUserCourses : Course[] = this.userCourses.concat(
+                                this.pinnedCourses.filter((c : Course) => this.userCourses.indexOf(c) < 0)
+                                );
                             this.recently.set(this.getRecently(pinnedOrUserCourses));
                             this.loadProgresses(pinnedOrUserCourses.map((c) => c.LastRecording.ID));
                         } else {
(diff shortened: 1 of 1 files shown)


web/ts/components/main.ts
```diff
@@ -26,15 +27,25 @@ export function mainContext(year: number, term: string) {
...
                 .catch((err) => {
                     console.error(err);
                 })
                 .then(() => {
                     this.liveToday = this.getLiveToday();
                     if (this.userCourses.length > 0) {
-                        this.recently.set(this.getRecently(this.userCourses));
-                        this.loadProgresses(this.userCourses.map((c) => c.LastRecording.ID));
+                        if (this.pinnedCourses.length > 0) {
+                            //
```
alexanderstephan, line 38:
```suggestion
```

alexanderstephan (approved):
First, I was a little bit confused, since I intuitively would have checked which of the pinned courses is a public course. However, your solution is more elegant, as it would work with additional course types in the future. Nice job!

PR #1224, opened 2023-10-27
1214 playback at t 0

### LINKED ISSUE #1214: Playback using link with t=0 does not work
**Describe the bug**
Video URLs can include a time parameter from which the video will start, e.g. https://live.rbg.tum.de/w/foo/123456?t=42  
Setting `t=0` does not work but instead resumes the video from the previous playback position.

Not sure if this is intended though.

**To Reproduce**
1. Open a video
2. Use a time code link for accessing this video.
    * t>0 works
    * t=0 continues the video from the last playback position instead of playing the video from the beginning.

**Expected behavior**
Start the video at t=0

 - OS: Linux
 - Browser: Firefox
 - Version: 118.0.2

YiranDuan721:
### Motivation and Context
fixed #1214 

### Description
Distinguish the behavior of the watchProgress function when the parameter t is set to 0 and when it is not set, so that it matches the description [in the comment](https://github.com/TUM-Dev/gocast/blob/109be9a6e6dd6c115ae9fffdc37992a7af1f0236/web/ts/TUMLiveVjs.ts#L314):
> If query parameter 't' is specified, the timestamp given by 't' will be used.

### Steps for Testing

1. Clear website cache
2. Log in
3. Navigate to a recording video
4. Jump to anywhere in the video other than the start
5. Refresh the page, note that the watch progress has been recorded and can be restored, so that the video resumes at the last progress
6. Add parameter "t=0" to current url and visit it
7. See the video plays from the start


alexanderstephan (changes requested):
The fix itself looks good! However, I would not like to check in the `go.work.sum` file here. I did not inspect the `go.work.sum`  file in detail, but it seems odd to me that so many lines are added, although there should be no dependency updates.

YiranDuan721:
Thanks for the review! I can't find a way to revert the change to "go.work.sum" file elegantly -- I'll just create a new branch and do the things again. Sorry for the bother, I'll be more careful with git next time.
:pingubonk:

PR #1229, opened 2023-11-06
No hardcoded file paths

### LINKED ISSUE #539: Eliminate hardcoded file paths

Mjaethers:
### Motivation and Context

Closes #539
Throughout the worker there are multiple instances of hardcoded file paths. By using a PathProvider code duplications is mitigated and all the hardcoded endpoints can be easily found as they are all in once place.

### Description

All the hardcoded file paths I found in the worker module have been moved to the pathprovider. The path providing functions follow the OS agnostic style laid out by [filepath](https://pkg.go.dev/path/filepath)
The overall code has not been changed much, mostly just moved about, so the expected behaviour should also not have changed.

### Steps for Testing

Prerequisites:
- A machine running a Unix-based OS

1. Run go test github.com/TUM-Dev/gocast/tools/pathprovider
2. Ultimately these changes should incur no change in the behaviour of the overall project

### Screenshots

none


Mjaethers:
There was a misunderstanding on my end regarding the difference between the new runner and the worker, so I don't think this PR makes sense any more. I'll close it, but I can re-do this for the actual runner if need be (but I don't think that will be the case).

PR #1233, opened 2023-11-12
Replace Worker with Runner

### LINKED ISSUE #533: Ending and restarting a self stream with different resolutions causes transcoding to fail.
I had a lecturer stream in `2560x1440` for an hour, take a break and continue the lecture in `1920x1080`. This messed up the .ts file and broke transcoding. 

Ideally multiple parts of a stream should be saved in multiple files which are normalized before concatenating and transcoding them.

### LINKED ISSUE #733: Pick up old jobs after reboot
A worker that is rebooted should restart the stream jobs it has been working on.

joschahenningsen:
### Motivation and Context

The current worker implementation is hard to maintain, we therefore want to replace it with the new runner. The goal is to have a clean code base that replaces everything the worker does.

### todos:

- All API endpoints the current worker offers should be available in the runner
  - [x] Request Stream
  - [x] Request Stream End 
  - [ ] Generate live preview #1435
  - [ ] Generate Section images #1435
  - [ ] Delete Section images (low prio) #1435
  - [ ] Combine Thumbnails #1435
- All API endpoints TUM-Live currently offers for the runner
  - [x] Register
  - [x] Heartbeat (nice to have)
  - [ ] request selfstream #1437
  - [ ] notify vodUpload finished #1436
  - [ ] Notify Silence results #1436
  - [x] Notify Stream Started #1436
  - [ ] Notify Thumbnails finished #1436
  - [ ] Get stream info for upload #1437
  - [ ] notify Transcoding Failure #1436

Functionality wise, the runner should:
- When request stream is called: 
  - grab the url with ffmpeg and create a hls stream on disk and notify tum-live that the stream is available
  - After the stream is done, convert the files, append them together and upload them to the vod service. 
  - create a thumbnail and the thumnails for the seekbar
  - detect silences in the video and send them to tum-live (not high prio for now)
- when a vod is uploaded, everything that happens after a livestream should also happen to the uploaded file
- when a selfstreamer starts streaming
  - ask tum-live for the details of the token and continue just like with a normal stream


runner/actions/stream.go
```diff
@@ -0,0 +1,149 @@
...
+				err = c.Start()
+				if err != nil {
+					log.Warn("streamAction: ", err)
+					time.Sleep(5 * time.Second) // little backoff to prevent dossing source
+					continue
+				}
+
+				localhls := fmt.Sprintf("%v/%d/%d/%s/%s/playlist.m3u8", hostname, courseID, streamID, version, end.Format("15-04-05"))
+
+				edgeURL, err := url.Parse(fmt.Sprintf("http://%v:8089/%v",
```
SebiWrn: Edge Server should be accessed via https if not localhost

SebiWrn: Edge Server isn't hosted on Port 8089 when not in development (port needs to be a config option)

SebiWrn (commented):
The whole generating part of the URL should be done on TUM-Live rather than on the runner

runner/actions/stream.go
DawinYurtseven: the config option has been added. mind that the port is rn only one port and if needed to have multiple edge servers, we need to think of a selection option to not swarm one server full

DawinYurtseven: Domain name has been added to the config for variability, with the default being "http://localhost".

DawinYurtseven:
Rn worked on the upload part and looked at the vod service of the worker, trying to see how it works and orientate myself after that.

DawinYurtseven:
Immediate TODO before we can merge this PR:
- [ ] Beta test FFMPEG fix
- [ ] databank merge, cleanup the migrations
- [ ] feature flag to switch between runner and worker for a lecture hall

PR #1236, opened 2023-11-17
1177 course import deduplicate emails

### LINKED ISSUE #1177: [Course-Import]: deduplicate emails
When importing courses, users responsible for a course are listed as notification recipients. Sometimes, a lecturer is listed multiple times with multiple roles in TUMonline (e.g. Verantwortlicher and Vortragender). They should only be listed/contacted once.

barisgul15:
### Motivation and Context
Issue 1177: https://github.com/orgs/TUM-Dev/projects/10/views/1?pane=issue&itemId=41336294
The array coming from TUMonline has some duplicates in contacts and these need to be eliminated.

### Description

The data coming from TUMonline sometimes duplicates contacts of some courses. course.contacts has 5 attributes:
first_name, last_name, email, main_contact, role.
All cases considered, the most distinctive among them is e-mail. So a script is added to modify the array before it is used for course contacts while importing courses.

### Steps for Testing

Prerequisites:
- Having admin rights

1. Log in
2. Click admin in the menu
3. Click import courses
4. Find a course that used to have duplicate contacts: e.g.  Einführung in die Games Eng(IN0031), IN2064, ...
5. Confirm that there are not duplicate contacts anymore.

### Screenshots

Before there was two Vortragende*r, Leiter*in, but the course has only one actually:
twoTimesListedGames
After the changes, the duplicate is eliminated:
Games_duplicates_eliminated

### DIFF
--- a/web/template/admin/admin_tabs/course-import.gohtml
+++ b/web/template/admin/admin_tabs/course-import.gohtml
@@ -78,11 +78,11 @@
                             <tbody>
                             <script>
                                 /*
-                                The data coming from TUMonline sometimes duplicates contacts of a course.
-                                The elements of course.contacts has 5 attributes:
-                                first_name, last_name, email, main_contact, and role.
+                                The data coming from TUMonline sometimes duplicates contacts of some courses.
+                                course.contacts has 5 attributes:
+                                first_name, last_name, email, main_contact, role.
                                 All cases considered, the most distinctive among them is e-mail.
-                                and that's why eliminating the duplicates are down accordingly.
+                                and that's why duplicates are eliminated accordingly.
                                 */
                                 function filterUniqueContacts(contacts) {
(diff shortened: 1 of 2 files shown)


SebiWrn (changes requested):
Please move the script tag to the correct typescript files and use it from there

barisgul15:
The location of the used filtering function is changed from course-import.gohtml to course-import.ts.

SebiWrn (changes requested):
I think there is still an issue with the checkboxes, because you just change the value in the contacts from your method, but the send mail method still uses the whole contact list without the updated values.
With that I think the checkboxes won't work and everybody will get an email even if you don't select them.

SebiWrn:
If you still want to merge this PR please address the issues and reopen the PR

PR #1242, opened 2023-11-20
Only one streaming key for each course

### LINKED ISSUE #124: Require only one streaming key for each course
Currently a stream can be ingested with a streaming key (secret) which is unique for each lecture. 

It would be more convenient if a lecturer wouldn‘t need to reconfigure their software for every stream. A single key for the course would be sufficient. 

**Additional context**
Users need to be able to regenerate the key if it gets leaked.

Mjaethers:
### Motivation and Context

Closes #124 
Currently, every time a lecturer starts a new self-stream, they have to go into their streaming software and change the stream server and key. This is a bit cumbersome as most lecturers only need to stream to one course, and therefore the credentials should be course-specific, instead of stream-specific.

### Description

Each course now has it's own stream key, which is generated when the course is created. Every time a new stream is created, it's stream key will default to the course stream key.

The stream key of a course can be regenerated in the course -> settings tab. This might be used if the stream key is leaked, or to re-sync all the stream keys should they have been changed individually.

The stream key of a stream can be regenerated in the Edit Lecture tab on the course page, this might be used for hosting concurrent streams of a course. Here the stream key can also be restored to the course stream key.

Since the worker can no longer identify streams by their key, it now identifies a stream by key and time. This will cause problems if stream times overlap within 30 minutes.

### Steps for Testing

Prerequisites:
- 1 Lecturer
- 1 Self-Stream setup

1. Log in 
2. Navigate to a course or create a new one
3. Create some Live-streams, notice they will all have the same key
4. Press Edit Lecture -> Regenerate Key of a Stream, notice it will no longer use the same key as the other streams in that course
5. Navigate to the course settings tab and press Regenerate course wide stream keys, notice that all the stream keys will change and will once again all be the same
6.  Regenerate Key of a Stream again
7. Press Edit Lecture -> Restore Key, notice it will once again use the same key as the other streams in that course

### Screenshots
Screenshot 2023-11-29 at 16-49-51 TUM-Live Administration
Changing the key of a specific stream

Screenshot 2023-11-20 at 13-47-00 MUT-Live Administration
Changing the key of all stream in course


Mjaethers:
Apologies regarding the state of the git history, I learned my lesson about syncing forks.

PR #1245, opened 2023-11-26
Fixed and optimized several issues in the front end of the course settings page

YiranDuan721:
## 1. Reset error state upon update of the "New Lecture" panel
**Before:** If anything failed when creating a new lecture, the alert banner "Something went wrong." will always be shown in the form, which is ambiguous.

<p align="center">
  A picture showing the wrong behaviour of the alert banner, which doesn't disappear after already left the page where the error has occurred.
  <br>
  <small>Example: After a user fails to upload a VoD, they are returned to the first step of the settings, but the error message does not disappear, causing confusion.</small>
</p>

**After:** The error banner disappears upon each update.

## 2. Correct the style of the buttons in the create-lecture-form
**Before:** In light mode, the button looks the same whether it is selected or not, making it difficult to notice changes after altering the options for creating a new lecture.

<p align="center">
  Button \
  Button \
  Button \
</p>

**After:** The change of style when selected can be correctly displayed in light mode.

<p align="center">
  Button \
  Button \
  Button \
</p>

## 3. On the edit-course page:
Change the style of the "continue" button when disabled to make it more obvious that it is disabled; Provide reasons for why the 'continue' button is disabled.

**Before:** Hard to distinct whether the "continue" button is disabled or not, in both light and dark mode.

<p align="center">
  The continue button is disabled, but not obvious - in light mode
<br>
  The continue button is disabled, but not obvious - in dark mode
</p>

**After:** When the button is disabled, it is clearly visible; at the same time, display the reason why the button cannot be clicked above it.

<p align="center">
  The continue button is disabled, reasons shown - in light mode, 1/3
  The continue button is disabled, reasons shown - in light mode, 2/3
  The continue button is disabled, reasons shown - in light mode, 3/3
</p>

<p align="center">
  The continue button is disabled, reasons shown - in dark mode, 1/3
  The continue button is disabled, reasons shown - in dark mode, 2/3
  The continue button is disabled, reasons shown - in dark mode, 3/3
</p>

## 4. On edit-course page:
When copying a course, ensure the correct display of yearW for winter semesters.

**Before:**

<p align="center">
  
</p>

**After:**

<p align="center">
  
</p>

## 5. After course copy:
Add a tooltip to explain that the current page is the setting page of the newly copied course, not the original one.

**Before:**

<p align="center">
  
</p>

**After:**

<p align="center">
  
</p>

## 6. Search course on the create-course page:
Hide the search result list if the input box is then emptied, and when the input box loses focus.

## 7. Reset yearW to empty string when year < 2000

**Before:**
The yearW won't be reset until another valid year (>2000) is input. The following picture shows the case when the user first inputs 2077 and then changed the value to 0.
<p align="center">
  
</p>

**After:**
Problem solved. When the year is invalid, yearW will be emptied. Also, the placeholder "25" will only be shown if "year" is displaying the placeholder "2024".

## 8. Placeholder of "year" for a semester-selection:

When the input field "year" gets focused when it's empty, set the placeholder as the input. This also allows the value to change from what the placeholder shows instead of 0, when the value is directly adjusted using the number increase/decrease button in the number-input field.

## 9. On create-course page, mark necessary field with red border if empty

This was not functioning because the line-width of the border had been incorrectly set to 0.

## 10. Ensure the slug of new courses is url-safe

On create-course page, restrict the characters input to the field "slug" to [a-zA-Z0-9\-_.+()~] (Numbers, English characters and 7 special characters). This could prevent the course from having an invalid url.

### DIFF
--- a/web/template/admin/admin_tabs/create-course.gohtml
+++ b/web/template/admin/admin_tabs/create-course.gohtml
@@ -3,7 +3,7 @@
 {{define "create-course"}}
     <div class="text-1 container mx-auto"
          x-data="{ tumonlineid: '', slug: '', title: '', year: '', yearW: '', semester: 'Wintersemester', numberAttendees: null, searchQuery: '', searchResults: [] , isSearchResultVisible: false}"
-        x-init="$watch('year',(value)=>{if(value<2000)return; yearW = (value%1000) + 1})">
+        x-init="$watch('year',(value)=>{ yearW = (value<2000) ? '' : (value%1000) + 1 })">
         <div class="min-w-screen flex items-center justify-center">
             <div class="w-full lg:w-5/6 p-3 bg-gray-100 dark:bg-secondary rounded dark:border dark:border-gray-500 shadow">
                 <h2 class="mb-0">Find your course from TUMOnline</h2>
--- a/web/template/admin/admin_tabs/create-course.gohtml
+++ b/web/template/admin/admin_tabs/create-course.gohtml
@@ -35,14 +35,14 @@
                     <h2>Manually enter Infos:</h2>
                     <div class="flex space-x-1">
                         <label for="name" class="hidden">Course Title</label>
-                        <input class="w-4/5 tl-input !ml-0" id="name" name="name" type="text"
+                        <input class="w-4/5 tl-input border !ml-0" id="name" name="name" type="text"
                                autocomplete="off"
                                placeholder="Einführung in die Informatik (IN0001)"
                                x-model="title"
                                required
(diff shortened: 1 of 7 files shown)


alexanderstephan:
Looks very good so far! 👍  I still have to check the code edge cases tomorrow.

alexanderstephan (approved):
Sorry for the delay. 😓

YiranDuan721:
No worries at all! Thank you for the review, Merry Christmas & have a nice holiday!

PR #1247, opened 2023-11-28
Enh/advanced logging

SebiWrn:
### Motivation and Context
Added custom JSON Logger to every package so logging output can be checked easily

### Description
Every log now prints a JSON String instead of just a readable text string

### Steps for Testing
Run your Instance of gocast


alexanderstephan:
Just so we don't forget it: If you update the Go version, don't forget to update https://github.com/TUM-Dev/gocast/blob/70bc1f36d0afdd53d6cdb8ce2875ff2ebe6a4558/.github/workflows/go-test.yml#L11 as well.

Dockerfile
DawinYurtseven, line 1: please change this back to tumlive.go

DawinYurtseven (commented):
all of it looks good just that one line in the dockerfile should be reverted

DawinYurtseven (approved):
okay hast die dockerfile reverted und sieht aus das du nur die logger replaced hast und die go version auf 1.21 gesetzt hast. sieht gut aus soweit ich sehe

PR #1253, opened 2023-12-06
popupContext ws subscription is added.

### LINKED ISSUE #1199: Popup chat does not update
**Describe the bug**
The popup chat does not receive any updates. Most notably, it does not show any new messages but remains frozen in time.

**To Reproduce**
1. Go to an active chat
2. Open the popup chat
3. Enter a new message on a separate window
4. See that the message has appeared in the chat next to the stream/video window, but not in the popup.

 - OS: Windows 10
 - Browser Firefox 118.0.2 (64-Bit)

**Additional context**
Tested during the livestream of GBS (https://live.rbg.tum.de/w/GBS/36950).

barisgul15:
### Motivation and Context

Fixes the issue Popup chat does not update  #1199 

### Description

When pop-up chat is opened, a new web socket(ws) is created so that the new chat can work properly. When examined, the only difference between normal-chat(chat near video) and popup chat is that, normal chat subscribes a void handler when created, but popup chat does not do that. As a solution, a void handler is created on popup.ts and the web socket subscribed via this handler and popup chat started working.

### Steps for Testing

Prerequisites:
- 1 Lecturer

1. Log in
2. Create a course and enable chat
3. Create a lecture and open the chats lecture
4. Open the popup chat and test sendin messages in both chats and check whether both checks are updated accordingly.

### DIFF
--- a/web/ts/components/popup.ts
+++ b/web/ts/components/popup.ts
@@ -5,7 +5,11 @@ import { SocketConnections } from "../api/chat-ws";
 export function popupContext(streamId: number): AlpineComponent {
     return {
         init() {
+            // subscription?
             SocketConnections.ws = new RealtimeFacade("chat/" + streamId);
+            // ws needs to subscribe, so that pop-out chat can work
+            const handler = (data) => {};
+            SocketConnections.ws.subscribe(handler);
         },
     } as AlpineComponent;
 }


YiranDuan721 (commented):
I tested locally and see the popup chat works correctly with the change. New posts and reactions from the sidebar chat and from another user in another browser are immediately updated in the popup chat; messages and reactions sent in the popup-chat are also updated in other windows.
However I think the comments are redundant and can be removed.

alexanderstephan (approved):
Awesome! 🎉  A surprisingly simple fix. I would like to merge and deploy this as soon as possible, so we can use it in the lecture on Wednesday.

I don't really mind the comments.

alexanderstephan:
@meandaD If it is okay for you, I would merge it as it is to avoid further delays. Feel free to press merge yourself.

YiranDuan721:
> Awesome! 🎉 A surprisingly simple fix. I would like to merge and deploy this as soon as possible, so we can use it in the lecture on Wednesday.
> 
> I don't really mind the comments.

I gree with this

PR #1259, opened 2023-12-07
probesize and analyzeduration added in transcode

DawinYurtseven:
added the probesize and analyze duration for the transcoding as talked in the meeting


SebiWrn (approved):
Well done :)

PR #1262, opened 2023-12-10
Skip initial silence

### LINKED ISSUE #1188: Setting to automatically "Skip" after opening stream
I would like the option to set a default behavior for streams upon opening, where they automatically skip the initial inactive period and start at the next active segment.

Manually pressing the button in the playback controls each time I open a stream is an alternative. However, an automated setting would provide a more convenient and seamless experience. Also avoids the extra loading time of first loading at 00:00 and then loading again after skipping

Mjaethers:
### Motivation and Context

Closes #1188 
Instead of opening a VOD, waiting for it to load, clicking skip, and waiting for it to load again there should be a user setting which allows the player to immediately skip past the first period of silence.

### Description

A new user setting to allow automatic skipping was added, this can be modified in the users settings page.
When this setting is enabled, the watch page sets the progress of the stream VOD to the maximum of the watch progress in the database and the end of the first period of silence.

This only happens if the stream is not a self-stream (lecture_hall_id != 0). This is because the progress is converted into a timestamp using the JS videoplayer duration, not the duration calculated by subtracting the stream end from stream beginning. I think it's safe to assume these two duration values will be the same when it is not a self-stream, as the recording just runs on in those cases. 

### Steps for Testing

Prerequisites:
* An instance of the new starter database (in the docs)

1. Navigate to the "Einführung Brauereiwesen" course
2. Select one of the VODs
3. Notice it will start at the beginning and blend in the skip ahead button
4. Navigate to the user settings
5. Enable "Automatically Skip First Silence"
6. Return to the "Einführung Brauereiwesen" Course
7. Select one of the VODs
8. Notice it will no longer start at the beginning, but where the skip ahead button would have taken you

### Screenshots

Screenshot_Skip_Silence_Setting


YiranDuan721 (approved):
Works smoothly!

YiranDuan721:
Before merging it to dev, I would like to make sure: we do rebuild the database container using "/docs/static/tum-live-starter.sql" for each deployment, right?

Mjaethers:
I'm afraid I don't quite follow.
My understanding of the file is that it is used to create a local docker container, as outlined in the [docs](https://github.com/TUM-Dev/gocast#setup-database) and that it has no impact on the database used in deployment. As far as I know this is just for testing during development.
The container would only be rebuilt if one were to manually rebuild it then.

But we can always get someone who is more familiar with the system in here if need be.

Mjaethers:
I thought it would make sense to commit these changes as well for future testing in this area, but we can also exclude this file if it's cluttering the db.

YiranDuan721:
Yes, I think changes to the sql file should be pushed anyway, but they need to be handled more carefully when deployed. I'd like to hear from someone more familiar with the project on this issue.

Mjaethers:
@alexanderstephan or @joschahenningsen what are your thoughts on the matter?

web/watch.go
```diff
@@ -13,6 +14,7 @@ import (
 	"github.com/TUM-Dev/gocast/tools"
 	"github.com/getsentry/sentry-go"
 	"github.com/gin-gonic/gin"
+	log "github.com/sirupsen/logrus"
```
joschahenningsen:
```suggestion
```

web/watch.go
```diff
@@ -93,6 +95,17 @@ func (r mainRoutes) WatchPage(c *gin.Context) {
 		} else if len(progress) > 0 {
 			data.Progress = progress[0]
 		}
+
+		// Check if user wants to skip first silence
+		autoSkip, err := tumLiveContext.User.GetAutoSkipEnabled()
+		if err != nil {
+			log.Printf("Couldn't decode user setting: %v\n", err)
```
joschahenningsen:
```suggestion
			logger.Error("Couldn't decode user setting", "err", err)
```

joschahenningsen (changes requested):
`tum-live-starter.sql` Is just for local testing and can be modified without worries :)

joschahenningsen (approved):
🍻

PR #1270, opened 2023-12-19
1255 share dialog should be quittable with esc

### LINKED ISSUE #1255: Share dialog should be quittable with 'esc'
only quittable by clicking the 'x' or outside of the popup. This makes mouse action neccesary and feels unnatural (for me. I basically click 'esc', realise it doesnt work and then use my mouse)

also make 'esc' close the popup

barisgul15:
### Motivation and Context

Fixes the issue: #1255

### Description

The eventlistener function to close the popup chat when esc is pressed is added to popup.ts and called in popup chat's html file.

### Steps for Testing

Prerequisites:
- 1 Lecturer

1. Log in
2. Create a course with chat enabled or navigate to a course for which chat is enabled.
3. Open popup chat and press esc to quit the popup chat tab.

### DIFF
--- a/web/template/popup-chat.gohtml
+++ b/web/template/popup-chat.gohtml
@@ -1,6 +1,7 @@
 {{- /*gotype: github.com/TUM-Dev/gocast/web.ChatData*/ -}}
 {{$course := .IndexData.TUMLiveContext.Course}}
 {{$stream := .IndexData.TUMLiveContext.Stream}}
+{{- $closeChatFunction := `closeChatOnEscapePressed();` -}}
 <!DOCTYPE html>
 <html lang="en" class="h-full dark">
 <head>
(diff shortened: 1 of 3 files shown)


SebiWrn:
Your code is working, but I think it doesn't solve the Issue mentioned in your PR. I think the Issue meant, that the share button at the bottom of a livestream, where you can share the current stream URL is not quittable with ESC.

barisgul15:
> Your code is working, but I think it doesn't solve the Issue mentioned in your PR. I think the Issue meant, that the share button at the bottom of a livestream, where you can share the current stream URL is not quittable with ESC.

Thanks for the feedback. When I saw popup, my mind immediately went to the chat but yeap, it was about something else.

SebiWrn (approved):
Works as expected

PR #1271, opened 2023-12-19
Set GIN to only log above level 400

SebiWrn:
# Updated Gin to only log status codes above 400 so we don't have too much logs and can read them automatically.


DawinYurtseven (approved):
looks fine

PR #1277, opened 2023-12-21
IPraktikum progress fetching and marking as watched

### LINKED ISSUE #38: [API] Implement progress endpoints
Implement `getProgress` and `postProgress` to keep track of where a user stopped watching a stream.

mo-cmyk:
### Motivation and Context
Solves https://github.com/TUM-Dev/gocast_mobile/issues/38

This adds possibility for the user to mark streams as watched as well as keeping track of the progress a user has made by watching a stream and according endpoints to fetch this data


api_v2/services/stream_service.go
```diff
@@ -2,59 +2,138 @@
...
+
+	result := db.Where("stream_id = ? AND user_id = ?", streamID, userID).First(p)
+
+	if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+		p.StreamID = streamID
+		p.UserID = userID
+		p.Progress = progress
+		p.Watched = progress == 1
+	} else if result.Error != nil {
+		return nil, result.Error
```
carlobortolan: Return `nil, e.WithStatus(http.StatusInternalServerError, err)` instead

api_v2/services/stream_service.go
```diff
@@ -2,59 +2,138 @@
...
+	} else {
+		p.Progress = progress
+		p.Watched = progress == 1
+	}
+
+	if err := db.Save(p).Error; err != nil {
+		return nil, err
```
carlobortolan: See comment above

api_v2/services/stream_service.go
```diff
@@ -2,59 +2,138 @@
...
+
+	result := db.Where("stream_id = ? AND user_id = ?", streamID, userID).First(p)
+
+	if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+		p.StreamID = streamID
+		p.UserID = userID
+		p.Progress = 1
+		p.Watched = true
+	} else if result.Error != nil {
+		return nil, result.Error
```
carlobortolan: See comment above

api_v2/services/stream_service.go
```diff
@@ -2,59 +2,138 @@
...
+	} else {
+		p.Progress = 1
+		p.Watched = true
+	}
+
+	if err := db.Save(p).Error; err != nil {
+		return nil, err
```
carlobortolan: See comment above

api_v2/services/stream_service.go
```diff
@@ -2,59 +2,138 @@
...
+}
+
+// GetProgress retrieves the progress of a stream for a user.
+func GetProgress(db *gorm.DB, streamID uint, userID uint) (*model.StreamProgress, error) {
+	p := &model.StreamProgress{}
+
+	err := db.Where("stream_id = ? AND user_id = ?", streamID, userID).First(p).Error
+
+	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+		return nil, err
```
carlobortolan: See comment above

api_v2/stream.go
```diff
@@ -2,135 +2,187 @@
...
+	}
+
+	return &protobuf.GetThumbsLiveResponse{Path: path}, nil
+}
+
+func (a *API) GetProgress(ctx context.Context, req *protobuf.GetProgressRequest) (*protobuf.GetProgressResponse, error) {
+	a.log.Info("GetStreamProgress")
+
+	uID, err := a.getCurrentID(ctx)
+	if err != nil && err.Error() != "missing cookie header" {
```
carlobortolan: `err != nil && err.Error() != "missing cookie header"` is used to allow  authenticated users (have valid jwt) as well as non-logged in users (missing jwt / cookie header) access to a certain method (e.g., GetPublicCourses or GetStream). In this case, GetProgress should only be accessed by authenticated users, so replace this with
```go
if err != nil {
  return nil, e.WithStatus(http.StatusUnauthorized, err)
}
```

api_v2/stream.go
```diff
@@ -2,135 +2,187 @@
...
+
+	return &protobuf.GetProgressResponse{Progress: progress}, nil
+}
+
+func (a *API) PutProgress(ctx context.Context, req *protobuf.PutProgressRequest) (*protobuf.PutProgressResponse, error) {
+	a.log.Info("SetStreamProgress")
+
+	uID, err := a.getCurrentID(ctx)
+
+	if err != nil && err.Error() != "missing cookie header" {
```
carlobortolan: Replace with
```go
if err != nil {
  return nil, e.WithStatus(http.StatusUnauthorized, err)
}
```
(see comment above)

api_v2/stream.go
```diff
@@ -2,135 +2,187 @@
...
+
+	progress := h.ParseProgressToProto(p)
+	return &protobuf.PutProgressResponse{Progress: progress}, nil
+}
+
+func (a *API) MarkAsWatched(ctx context.Context, req *protobuf.MarkAsWatchedRequest) (*protobuf.MarkAsWatchedResponse, error) {
+	a.log.Info("MarkAsWatched")
+
+	uID, err := a.getCurrentID(ctx)
+	if err != nil && err.Error() != "missing cookie header" {
```
carlobortolan: Replace with
```go
if err != nil {
  return nil, e.WithStatus(http.StatusUnauthorized, err)
}
```
(see comment above)

carlobortolan (changes requested):
Looks good to me :+1
Make sure to return e.g., `e.WithStatus(http.StatusInternalServerError, err)` instead of just `err` when executing db queries and quickly fix method access in progress endpoints (only authenticated users should access those)
***
***
Note that something like:
```go
_, err := GetStreamByID(db, streamID)
if err != nil {
  return nil, err
}
``` 
is allowed as `GetStreamByID()` already returns only e.WithStatus(...) errors and therefore we can simply return `err`.

***

This:
```go
if err := db.Save(p).Error; err != nil {
  return nil, err
}
```
however could result in undefined behavior, as we should try to always return `e.WithStatus(...)` and in this case it could return a different error.

carlobortolan (approved):

PR #1278, opened 2023-12-21
Added Filter by Role to user search

SebiWrn:
### Motivation and Context
You can now filter users by their role, improves administration and overview over admins.

### Steps for Testing
Prerequisites:
- 1 Admin

1. Log in
2. Navigate to User management page
3. View the new Dropdown

### Screenshots
image

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -129,7 +129,7 @@ func (r usersRoutes) prepareUserSearch(c *gin.Context) (users []model.User, err 
 	reg, _ := regexp.Compile("[^a-zA-Z0-9 ]+")
 	q = reg.ReplaceAllString(q, "")
 	// Removed in order to make the search work with empty query but selected role
-	if len(q) < 3 && rQ == "-1" {
+	if len(q) < 3 && (rQ == "-1" || rQ == "") {
 		_ = c.Error(tools.RequestError{
 			Status:        http.StatusBadRequest,
 			CustomMessage: "query too short (minimum length is 3)",
--- a/api/users.go
+++ b/api/users.go
@@ -125,9 +125,11 @@ func (r usersRoutes) updateUser(c *gin.Context) {
 
 func (r usersRoutes) prepareUserSearch(c *gin.Context) (users []model.User, err error) {
 	q := c.Query("q")
+	rQ := c.Query("r")
 	reg, _ := regexp.Compile("[^a-zA-Z0-9 ]+")
 	q = reg.ReplaceAllString(q, "")
-	if len(q) < 3 {
+	// Removed in order to make the search work with empty query but selected role
+	if len(q) < 3 && rQ == "-1" {
 		_ = c.Error(tools.RequestError{
 			Status:        http.StatusBadRequest,
 			CustomMessage: "query too short (minimum length is 3)",
--- a/api/users.go
+++ b/api/users.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


YiranDuan721:
There's a small problem: after filtering, the user's email doesn't seem to be displayed

karjo24:
Filtering locally results in no show of potential results. Simplest case of this is filtering for students without query string. Change the API instead.

karjo24:
> There's a small problem: after filtering, the user's email doesn't seem to be displayed

I cannot reproduce this problem in Mozilla Firefox for Ubuntu 132.0.2 (64-Bit) or Google Chrome Version 130.0.6723.58 (Official Build) (64-bit).

api/users.go
```diff
@@ -125,16 +126,23 @@ func (r usersRoutes) updateUser(c *gin.Context) {
 
 func (r usersRoutes) prepareUserSearch(c *gin.Context) (users []model.User, err error) {
 	q := c.Query("q")
-	reg, _ := regexp.Compile("[^a-zA-Z0-9 ]+")
+	rQ := c.Query("r")
+	reg := regexp.MustCompile("[^a-zA-Z0-9 ]+")
 	q = reg.ReplaceAllString(q, "")
-	if len(q) < 3 {
+	// Removed in order to make the search work with empty query but selected role
```
joschahenningsen:
```suggestion
	// make the search work with empty query but selected role
```

api/users.go
```diff
@@ -125,16 +126,23 @@ func (r usersRoutes) updateUser(c *gin.Context) {
...
+	if len(q) < 3 && (rQ == "-1" || rQ == "") {
 		_ = c.Error(tools.RequestError{
 			Status:        http.StatusBadRequest,
 			CustomMessage: "query too short (minimum length is 3)",
 		})
 		return nil, errors.New("query too short (minimum length is 3)")
 	}
-	users, err = r.UsersDao.SearchUser(q)
+	role, err := strconv.ParseUint(rQ, 10, 64)
```
joschahenningsen: catch the error separately and return bad request if parsing failed.

joschahenningsen (dismissed):
Looks great, just added two remarks

api/users.go
```diff
@@ -125,16 +126,27 @@ func (r usersRoutes) updateUser(c *gin.Context) {
...
+	if err != nil && rQ != "" && rQ != "-1" {
+		tools.RenderErrorPage(c, http.StatusBadRequest, "invalid role")
```
joschahenningsen: API endpoints should not render HTML pages on error but rather use 
https://github.com/TUM-Dev/gocast/blob/95804eb169b5a717f715c55fe6858478c098ccec/web/course.go#L31

api/users.go
```diff
@@ -125,16 +126,31 @@ func (r usersRoutes) updateUser(c *gin.Context) {
 
 func (r usersRoutes) prepareUserSearch(c *gin.Context) (users []model.User, err error) {
 	q := c.Query("q")
-	reg, _ := regexp.Compile("[^a-zA-Z0-9 ]+")
+	rQ := c.Query("r")
+	reg := regexp.MustCompile("[^a-zA-Z0-9 ]+")
 	q = reg.ReplaceAllString(q, "")
```
kordianbruck: all of these variable names are hard to understand without context. Also why are we only enforcing the regex on one of the two paramters? Does rQ not need to be checked for validity?

kordianbruck (approved):

api/users.go
kordianbruck: Edited the PR, PTAL if you think thats better (otherwise revert my commit please).

karjo24: > Also why are we only enforcing the regex on one of the two paramters? Does rQ not need to be checked for validity?

Probably hard to immediately understand and therefore bad code, but it's checked before it's used: https://github.com/TUM-Dev/gocast/blob/6d0c3cc740995f7205a6d312bb9cab42d7305be0/api/users.go#L143-L144

PR #1281, opened 2023-12-25
popup chat close on esc added.

### LINKED ISSUE #1280: Close popup chat by clicking esc

The popup chat opens another tab, so that students/lecturers can watch the lessons comfortably while seeing the chat, but while testing other issues on popup chat, I generally did not use the popup chat in full-screen mode. I used it as a smaller tab near some other tabs, so I would rather close the popup chat by simply clicking esc, rather then clicking the browsers quit tab button.

popupchat

Add an event listener so that when esc is clicked, the popup chat tab closes.

It is not the most necessary feature, but adding this one can make using popup chat easier.

barisgul15:
### Motivation and Context

closes #1280

### Description

add an event listener so that popup chat is closed when esc is clicked

### Steps for Testing

Prerequisites:

1. Log in
2. Navigate to a lecture recording with chat enabled or create one
3. Open popup chat and click esc and check whether it closes the popup chat browser or not

### DIFF
--- a/web/template/popup-chat.gohtml
+++ b/web/template/popup-chat.gohtml
@@ -17,7 +17,6 @@
     <link href="/static/assets/css-dist/home.css?v={{if .IndexData.VersionTag}}{{.IndexData.VersionTag}}{{else}}development{{end}}"
           rel="stylesheet">
 </head>
-<body class="bg-white dark:bg-secondary h-screen" x-data="interaction.popupContext({{$stream.ID}})">
+<body class="bg-white dark:bg-secondary h-screen" x-data="interaction.popupContext({{$stream.ID}})" x-init="interaction.closeChatOnEscapePressed()">
     {{template "chat-component" .}}
-    <script> interaction.closeChatOnEscapePressed(); </script>
 </body>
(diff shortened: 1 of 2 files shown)


alexanderstephan:
I think that this is a convenient feature. 👍 But we should document that shortcut somewhere, right? Maybe we could have a couple other shortcuts here which are then explained in a small dialog box. I imagine a small icon with a keyboard logo. It could be a component that we could reuse on other parts of the website where we use shortcuts.

alexanderstephan:
Maybe one other note. Commit messages should not end with a dot (`.`) and use sentence case.

barisgul15:
> I think that this is a convenient feature. 👍 But we should document that shortcut somewhere, right? Maybe we could have a couple other shortcuts here which are then explained in a small dialog box. I imagine a small icon with a keyboard logo. It could be a component that we could reuse on other parts of the website where we use shortcuts.

While implementing that I was also asking myself how will user gonna know that popup chat can be closed by pressing esc but could not think of a convenient idea. I think your suggestion makes sense. If its ok for you, I can create a new issue about it and work on it?

SebiWrn (approved):
Works as expected

PR #1282, opened 2023-12-27
1182 more replayplayback speed options

### LINKED ISSUE #1182: More replay/playback speed options
I personally would like to have a slider instead of the current options for speed regarding replay as sometimes 1 is to little and 1.25 to great. In such cases (and probably also other cases) it would be beneficial to be able to adjust the speed more precisely maybe with an increment of 0.05 / 5% or even 0.01 / 1% (1% admittedly might probably be a bit too small, but that would result in complete control over the playback speed). Alternatively I would suggest that the increment could be set somewhere by the user or that you could simply type a value between maybe 0.01 and 3 or maybe even 4 into a input field changing the playback speed this way. In conclusion I would still opt for the first option namely the slider as it appears to me as the best option for this.

SebiWrn:
### Motivation and Context
Add up to 3 custom playback speeds for students to match the lecture speed.

### Steps for Testing
Prerequisites:
- 1 Account

1. Log in
2. Navigate to settings
3. Add a custom playback speed
4. Navigate to a livestream and use your setting

### Screenshots
image


DawinYurtseven (commented):
please add back the comments that were there before.

YiranDuan721 (approved):
Works perfect as far as I can think of.
This is what I personally really need as a user, thanks for the change ;)

PR #1285, opened 2023-12-28
Show worker version on worker dashboard

SebiWrn:
### Motivation and Context

Show worker version instead of version hash


DawinYurtseven (approved):
delayed approval

PR #1286, opened 2023-12-29
1094 list view in new UI

### LINKED ISSUE #1094: List View in new UI
Hey,

it would be nice to have a toggle for switching between a "Thumbnail View" and "List View".
A Thumbnail View is useful for finding out the topic of a lecture (assuming the lecture has only a timestamp), while a
List View (similar to the old UI) is much more convenient for finding specific lectures without clicking through pages of thumbnails (less Pagination)

SebiWrn:
### Motivation and Context
Adds a List View to the course view.

### Description
You can now switch between the thumbnail and list view in the course page.

### Steps for Testing
Prerequisites:
- 1 Account
- 1 course with multiple lectures

1. Log in
2. Navigate to a Course
3. Switch to list view

### Screenshots
Default View:
image

List View:
image

### DIFF
--- a/.github/workflows/deploy-workers.yml
+++ b/.github/workflows/deploy-workers.yml
@@ -54,6 +54,6 @@ jobs:
           context: ./${{ matrix.module }}
           pull: true
           push: true
-          build-args: version=${{ steps.vars.outputs.sha_short }}
+          build-args: version=${{ github.ref_name }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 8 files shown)


SebiWrn (commented):
Somehow previous commits are also in this PR, but only really changed files are home.gohtml and course.ts, everything else is already merged into dev

YiranDuan721:
Currently, the status of whether the list view is displayed is reset when the page is accessed again. I would prefer it to be preserved like the other options. I can see that the state "listView" is indeed preserved in the browser's local storage, but it doesn't look like it's being restored properly when the page loads or reloads.

SebiWrn:
Thank you for your feedback, problem should be fixed now :)

DawinYurtseven (approved):
looks good!

PR #1304, opened 2024-01-10
Fixed bug with long course titles

### LINKED ISSUE #1301: Long words in lecture titles spoil the display of web pages
### Case 2. Playlist in the watch page
A long word in the title keeps the playlist away from where it should be (it's too wide to fit in) XD

SebiWrn:
### Motivation and Context
Fixes UI bugs with long titles.

### Description
Text will now be cut off if it's too long and only be shown on mouse hover.

### Screenshots
image
image

### DIFF
--- a/web/template/home.gohtml
+++ b/web/template/home.gohtml
@@ -476,11 +476,14 @@
                                                                 </div>
                                                             </a>
                                                             <div class="relative flex justify-between px-1 min-h-[32px]">
-                                                                <div>
+                                                                <div class="w-full">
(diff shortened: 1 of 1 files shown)


YiranDuan721:
Looks good to me. Would it be a good idea to add "overflow-hidden" to all elements? Because this problem occurs anywhere...

SebiWrn:
Maybe we can do this in an enh branch, but as a quick fix for this issue, this works for the lecture title. What do you think about this?

YiranDuan721 (approved):

PR #1305, opened 2024-01-10
Enh/update dependencies

SebiWrn:
# Update of all go and npm dependencies


DawinYurtseven (approved):
brain says yes! 

but jokes aside, thanks for updating the dependencies. it should work the same as before so no problems there! ^^

PR #1307, opened 2024-01-11
Show watch progress in playlist view

### LINKED ISSUE #1248: show watch progress in playlist view
In the normal course view (`&view=3`) a progress bar is shown at the bottom of a thumbnail. This makes for a good indicator of watch progress. However when opening the playlist in the normal view of a stream (`/w/<course>/<id>`) those progress bars are no longer shown.
This makes it hard to tell which recordings have been watched (and to what degree)

Show the same progress bars there as well.
alternatively some sort of indicator whether the recording has been watched would solve my issue as well. (would create the problem if deciding what 'watched' specifically means and probably isn't much easier to implement anyways)

open streams in a new tab and hope progress is "restored". This is very cumbersome and doesn't work reliably

**Additional context**
may also be a bug?
Firefox 119 on Ubuntu 22

Mjaethers:
### Motivation and Context

Closes #1248 
Much how the watch progress of a recording is displayed on the thumbnail of a stream on the course page, the watch progress of a recording is displayed in the playlist view.

### Description

The stream API now also passes the stream progress in the returned StreamPlaylistEntry object. This data is passed to the Typescript Progress class constructor allowing the html to call functions on the progress so that it can be rendered in the same manner as on the course page.

### Steps for Testing

1. Log in
2. Navigate to a recording
3. Blend in the playlist
4. Notice the the thumbnails show a progress bar

### Screenshots

Screenshot 2024-01-11 at 17-38-19 TUM-Live Einführung Brauereiwesen VL 2 Wie mache ich Bier

### DIFF
--- a/api/stream.go
+++ b/api/stream.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/TUM-Dev/gocast/tools/pathprovider"
+	log "github.com/sirupsen/logrus"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


YiranDuan721:
I think this solution would work, but unfortunately at the moment it doesn't work correctly for me. The watch progress displayed in the playlist does not match the one on the course page (or the progress saved in database):
test_progress_in_playlist

Another small idea: I would like the width of the play progress not to exceed the width of the playlist, so that the playlist can be embedded perfectly in the page as it was, without the scrollbar at the bottom.

Mjaethers:
Interesting, this is a bug I thought I had fixed, but I suppose I wasn't rigorous enough with different browsers and windows sizes. Thanks for looking into this, I'll try to fix it as soon as I have the time to.

YiranDuan721:
Ooops, it turns out that was due to me forgetting to clear the browser cache before testing. After that it works perfectly.

 Sorry, my bad!

YiranDuan721 (approved):
Works as expected for me locally :)

YiranDuan721 (approved):
No problem ;)

YiranDuan721:
~~I'm not sure about the failing check. It seems to be concerning some problem with user authentification? To test about that, also to trigger the test server deploying test again, I will push one commit here.~~
I'm not sure, but I think you can directly work on this repository, instead of the forked one, to avoid this problem

Mjaethers:
Okay thanks. I'll try without a fork next time, that might be it.

PR #1310, opened 2024-01-14
Add "remember me" option for log in

### LINKED ISSUE #1302: UX: Longer sessions
Currently, when I logon to TUMLive, those sessions expire after a week.
On Desktop, this isn't a big of a deal, but on mobile it is really frustrating that I have to relogin every week.

Make sessions (much) longer in general (or not expire at all) and/or refresh them when the user uses the site. 
I know of no other streaming sites where I have ever had the problem of being logged out.
Frankly, I don't think an applications such as TUMLive warrants such special security measures.

- Only make sessions longer for students (so admins/lecturers don't lose any security benefits)
- Make it a setting
- Make them longer only for mobile useragents
- Have a TUMLive app (but I guess inside of that app you would still have persistent sessions, so why not do that for the website)

**Additional context**
I am not sure if there is something you can do about this given that this is probably mostly handled by Shibboleth, but I hope so.

YiranDuan721:
### Motivation and Context

Provide users with the option to make the login status valid for a longer period of time (6 months instead of 1 week).
Related: #1302

### Description

Passes on whether the user has selected the "remember me" option by setting a cookie, and if so, the jwt cookie set at login is set to be valid for 6 months instead of one week.

### Steps for Testing

1. Logout if necessary, and access the login page
2. Login via saml or internal login, without selecting the "remember me" option
3. Check the jwt cookie, it expires in 1 week; the decoded payload also shows an expire date in 1 week
4. Logout and login again, with "remember me" option checked
5. Check the jwt cookie, it expires in 6 months; the decoded payload also shows an expire date in 6 months

### Screenshots

Change of UI on login page:

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -86,7 +86,7 @@ func (r usersRoutes) impersonateUser(c *gin.Context) {
 		})
 		return
 	}
-	tools.StartSession(c, &tools.SessionData{Userid: u.ID})
+	tools.StartSession(c, &tools.SessionData{Userid: u.ID}, false)
 }
 
 func (r usersRoutes) updateUser(c *gin.Context) {
--- a/tools/session.go
+++ b/tools/session.go
@@ -12,21 +12,26 @@ type SessionData struct {
 	SamlSubjectID *string
 }
 
-func StartSession(c *gin.Context, data *SessionData) {
-	token, err := createToken(data.Userid, data.SamlSubjectID)
+func StartSession(c *gin.Context, data *SessionData, rememberMe bool) {
+	maxAgeInDays := 7 // by default, log-in status expires in one week
+	if rememberMe {
+		maxAgeInDays = 30 * 6 // if user chooses "remember me", let log-in status be valid for 6 months
+	}
+
+	token, err := createToken(data.Userid, data.SamlSubjectID, maxAgeInDays)
 	if err != nil {
 		logger.Error("Could not create token", "err", err)
 		return
 	}
-	c.SetCookie("jwt", token, 60*60*24*7, "/", "", CookieSecure, true)
+	c.SetCookie("jwt", token, 60*60*24*maxAgeInDays, "/", "", CookieSecure, true)
 }
 
-func createToken(user uint, samlSubjectID *string) (string, error) {
+func createToken(user uint, samlSubjectID *string, maxAgeInDays int) (string, error) {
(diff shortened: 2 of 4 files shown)


YiranDuan721:
Help wanted: This feature has not been tested under saml login, how can I do it locally or in some other way?

joschahenningsen:
> how can I do it locally or in some other way

This would involve copying the saml configuration part from the production config.yaml as well as the certificates to your local computer (I have done this before but it's not a great & secure way of doing this...)

I'd argue that the saml provider (login.tum.de) also has a remember me button that should be respected by gocast when it's checked and this feature therefore only should apply to the internal login. Then we additionally should implement handling the saml login preference.

meandaD:
> > how can I do it locally or in some other way
> 
> This would involve copying the saml configuration part from the production config.yaml as well as the certificates to your local computer (I have done this before but it's not a great & secure way of doing this...)
> 
> I'd argue that the saml provider (login.tum.de) also has a remember me button that should be respected by gocast when it's checked and this feature therefore only should apply to the internal login. Then we additionally should implement handling the saml login preference.

Thank you! I will adjust the code with this idea.

YiranDuan721:
Thank you!

> I'd argue that the saml provider (login.tum.de) also has a remember me button that should be respected by gocast when it's checked and this feature therefore only should apply to the internal login. Then we additionally should implement handling the saml login preference.


But the "remember me" option provided by TUM saml is supposed to be valid only "for the time your browser is opened".
image

I'd say this seems to go against the requirement mentioned in #1302, since the valid period of the login status is even shorter. (That would come out like Moodle where one needs to login a hundred times per day... Or is that instead the correct configuration for saml login?)

joschahenningsen:
I don't know if Moodle does this correct. GoCast implements its own session management and I'm all for going against implementation references for SAML if that provides a smoother UX. But maybe we shouldn't issue jwts that are valid for 60 days but instead implement a way of refreshing sessions for logged in users who regularly visit the site. Maybe you can do some research on best practices of managing such sessions?

I imagine there must be a way of having an indefinite session that
- is valid for a period of time, e.g. 2 weeks
- gets refreshed for another 2 weeks if you visit the website and it will expire within the next week
while still being able to invalidate all sessions associated with ones account.

YiranDuan721:
Thanks very much for the hint! I'll do further research on this in the next couple of days.

PR #1316, opened 2024-01-21
"Remember me" feature

### LINKED ISSUE #1302: UX: Longer sessions

YiranDuan721:
### Motivation and Context
Related issue: #1302 

### Description
1. Added a checkbox on the login page that allows users to select "Remember Me" before logging in
2. The state whether or not "Remember Me" is checked, is saved in a temporary cookie during the login process, and stored as a jwt claim after successfully logged in
3. In the middleware, check whether the jwt should be refreshed after validation, if so, extend the validity period in both jwt and the cookie's age
4. A `MaxTokenLifetimeWithRememberMeInDays` is set, to prevent the token from being valid forever
5. A `MinUpdateIntervalInHours` is set, so that (if "Remember Me" checked upon logging in) the token won't get refreshed too frequently

### Steps for Testing

1. To be able to see the effect quickly, first change `MinUpdateIntervalInHours` in "\tools\middlewares.go" to a smaller value, e.g. 0.01, which corresponds to 36 seconds
2. Log in with "Remember Me" selected
3. Refresh the website within the period of `MinUpdateIntervalInHours`, see the cookie remain unchanged
4. Refresh the website after `MinUpdateIntervalInHours`, see both the value and the Max-Age of the cookie "jwt" is changed, and in the payload of jwt, the value of "exp" corresponds to the Max-Age of the cookie

### Screenshots

Change of UI:

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -87,7 +87,7 @@ func (r usersRoutes) impersonateUser(c *gin.Context) {
 		})
 		return
 	}
-	tools.StartSession(c, &tools.SessionData{Userid: u.ID})
+	tools.StartSession(c, &tools.SessionData{Userid: u.ID}, false)
 }
 
 func (r usersRoutes) updateUser(c *gin.Context) {
--- a/tools/middlewares.go
+++ b/tools/middlewares.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/TUM-Dev/gocast/dao"
 	"github.com/TUM-Dev/gocast/model"
@@ -13,6 +14,11 @@ import (
 	"github.com/getsentry/sentry-go"
 	"github.com/gin-gonic/gin"
 	"github.com/golang-jwt/jwt/v4"
+)
+
+const (
+	MaxTokenLifetimeWithRememberMeInDays = 180
+	MinUpdateIntervalInHours             = 1
 )
 
 var templateExecutor TemplateExecutor
@@ -25,6 +31,7 @@ func SetTemplateExecutor(e TemplateExecutor) {
 // JWTClaims are the claims contained in a session
 type JWTClaims struct {
 	*jwt.RegisteredClaims
+	UpdatedAt     *jwt.NumericDate
 	UserID        uint
 	SamlSubjectID *string // identifier of the SAML session (if any)
 	RememberMe    bool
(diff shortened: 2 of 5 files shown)


hmelder:
A 180-day session lifetime is *way* to long. I think one to two weeks is perfectly fine for a VoD platform.

We are already storing the cookie, so naming it "Remember Me" is a bit misleading. It is typically used when you want to use a session cookie, and a persistent cookie (with an expiration date).

A few examples:
Spotify has a "Remember Me" toggle sets the expiration date of the token to one month. When not toggled, only a session cookie is stored.

TIDAL does not have a "Remember Me" toggle and stores a sid cookie with a lifetime of 7 days.

alexmo1997:
> A 180-day session lifetime is _way_ to long. I think one to two weeks is perfectly fine for a VoD platform.

Why do you think a 180-day session is too long?
Just now, I inspected my YouTube and Netflix cookie.
Netflix has 6 months cookie expiration dates, while YouTube seems to have 2 years.
Furthermore, I don't think I have ever been logged out of either of those before without changing anything on my end. They probably even do session renewals to extend that period.

YiranDuan721:
Thank you for the comments! As it comes to security, I do think we should be more careful.

But some thoughts on how a "remember me" feature, realised by refreshing the token for logged-in users from time to time, would actually improve security: With this feature, we can then shorten the default duration of validity of the token, or, provide an option to shorten it (lecturers and admins might want this behaviour).

And we are using jwt instead of a session cookie, which makes a great difference, since invalidating a jwt before its expiration is more tricky.

SebiWrn (commented):
I think this feature is no longer necessary as we are currently switching to another login system

PR #1319, opened 2024-01-23
Date is now clickable, if lecture doesn't have a name

### LINKED ISSUE #1314: When lectures are not named, you can not click them on list view
**Describe the bug**
For the courses which's lectures are not named, there is no link to click so that user can watch the lecture. Normal view still shows the lectures, but when list view is selected, there is only writing on the screen.

**To Reproduce**
Click any course which's lectures are not named and click list view. For example; Advanced Topics of Software Engineering (IN2309, IN2126), Approximation Algorithms [CIT4100003], Automaten und formale Sprachen (IN2041), etc.

**Expected behavior**
When list view is selected, and when the lecture is clicked, the video player should open to that users can watch the lecture. 

**Screenshots**
A course with named lectures in list view:

named_lectures

A course without named lectures in list view:

not_named_lectures

A course without named lectures in normal view:

not_named_lectures_normal_view

**Additional context**

SebiWrn:
### Motivation and Context
Fixes #1314

### Description
If a lecture doesn't have a name, the date is now clickable so you can select them in list view as well.

### DIFF
--- a/web/template/home.gohtml
+++ b/web/template/home.gohtml
@@ -491,8 +491,18 @@
 
                                                                         </a>
                                                                     </template>
+                                                                    <template x-if="vod.HasName()">
+
                                                                     <span class="date text-sm"
                                                                           x-text="vod.FriendlyDateStart()"></span>
+                                                                    </template>
+
+                                                                    <template x-if="!vod.HasName()">
+                                                                        <a :href="course.WatchURL(vod.ID)">
+                                                                        <span class="date text-sm"
+                                                                              x-text="vod.FriendlyDateStart()"></span>
+                                                                        </a>
+                                                                    </template>
                                                                 </div>
                                                                 <button type="button" @click="vod.Dropdown.toggle()"
                                                                         class="md:group-hover:opacity-100 md:opacity-0 px-2">
--- a/web/template/home.gohtml
+++ b/web/template/home.gohtml
@@ -492,14 +492,13 @@
                                                                         </a>
                                                                     </template>
                                                                     <template x-if="vod.HasName()">
-
                                                                     <span class="date text-sm"
                                                                           x-text="vod.FriendlyDateStart()"></span>
                                                                     </template>
 
                                                                     <template x-if="!vod.HasName()">
-                                                                        <a :href="course.WatchURL(vod.ID)">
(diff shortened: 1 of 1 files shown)


barisgul15:
Yeap, functionally works perfectly, but maybe the date which is also the link can look better. What I mean is that for example the font of the named lecture link is bold and when mouse is on it, a straight line appears under the text which indicates that it is a link. It is also the case for the lecture name links on homepage. 
Named lecture link: 
manually_namex_lecture_link
List view link: 
automatically_named_lecture_link
Home course link:
homepage_lecture_link

SebiWrn:
Text now shows up in full size if List View is toggled

web/template/home.gohtml
```diff
@@ -491,8 +491,17 @@
...
+                                                                    </template>
+
+                                                                    <template x-if="!vod.HasName()">
+                                                                        <a :class="isListView() ? 'title overflow-hidden' : ''" :href="course.WatchURL(vod.ID)">
+                                                                        <span class="date" :class="isListView() ? '' : 'text-sm'"   
+                                                                              x-text="vod.FriendlyDateStart()"></span>
+                                                                        </a>
+                                                                    </template>
                                                                 </div>
                                                                 <button type="button" @click="vod.Dropdown.toggle()"
```
YiranDuan721, line 506: This button is currently too far away from the title... This makes it hard for me to notice that there's a button. Do you think it would be better to place it more to the left?
image

YiranDuan721 (changes requested):
Works great in total!

web/template/home.gohtml
SebiWrn: Where would you put it?

YiranDuan721: I would put it directly to the right of the date, with appropriate padding

SebiWrn: But this would cause the button to be at a different position for every entry in the list

YiranDuan721: That's true... And I don't think this button is very frequently used (those who click on them a lot will probably find them anyway). I'll approve the change and maybe you can deploy this fix along with (after) the test tonight.

YiranDuan721 (approved):

PR #1324, opened 2024-02-01
melody config max message size increased

### LINKED ISSUE #1207: Chat cannot send messages longer than ~385 characters: Websocket message too big
**Describe the bug**
When sending a message with > ~385 characters, the message is sent to the backend via the websocket, but the websocket is closed with the close code 1009, [which indicates that the websocket message is too large](https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1).

**To Reproduce**
1. Go to stream/VOD with an active chat
2. Enter a message with ~400 characters
3. Send message
4. Notice that the message never appeared

barisgul15:
### Motivation and Context
fixes #1207 

### Description
melodyInstance := melody.New() creates a melody instance, which has a default config which has max message size is fixed to 512 bytes. That was increased to 1200. The text area, the user write their message has a max size 1000 and all the messages written there can be sent. That means limit is increased to 1000 for now. We decided to be 2000 chars at the meeting, the reason I let it stays as 1000 is that, different problems occur for messages longer than 1000 chars. I also wanted to add some warnings for longer messages and do some frontend changes because longer messages sometimes cause some visual problems. I want to open another issue for that and work on them here. This PR is just a short term solution for sending longer messages.

### Steps for Testing

Prerequisites:
Admin or Lecturer rights

1. Log in
2. Create a new course with chat enabled and create a livestream to test the chat
3. type a message longer than 400 chars, if you reach 1000 chars, the text are will not allow you to write anymore.
Example 100 char random message: 

xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxfskjdfkjsdkcmsdkcmksdmcksdmcksdnvjbsdjvbsdjbvjsdbvjbsdjcvbnjsdkncksdaggckmsdkcmsdkmcksdxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxkkkkkkkkkkjjjjjjjjjjpppppppppp12345sdkj

### Screenshots

The message above is sent and accepted by the channel:
long_text_msg


SebiWrn:
Maybe we can also add some number at the bottom of the chatbox with the current char count and the max char count?

SebiWrn (approved):

PR #1325, opened 2024-02-01
All methods on models as pointer receivers

### LINKED ISSUE #1272: All methods on models should use pointer receivers
Goland has been harassing me about the fact that some structs in the package model mix value and pointer receivers. This is not recommended by the Go documentation.

~~It would make sense to convert the methods with pointer receivers to value receiving methods as the mutability provided by using pointers is achieved through the dao design pattern.~~
It would make sense to convert the methods with value receivers to pointer receiving methods as this would be more efficient for large structs.

~~It would also be possible to convert the methods with value receivers to methods with pointer receivers, the docs suggest this would be more efficient but since most methods use value receivers and the commitment to the dao design pattern was made, I think using value receivers makes more sense.~~
It could either be left as is or the dao pattern could be extended for the model package.

**Additional context**
The Go docs talk about this in their FAQ [here](https://go.dev/doc/faq#methods_on_values_or_pointers)

Mjaethers:
### Motivation and Context
Fixes #1272.
In the model package many structs have methods on both value and pointer receivers. Having mixed receivers is for consistencies sake not ideal, but since many of these structs are also quite large, and using value receivers implies copying data, using pointer receivers would be slightly more performant.

### Description
Any value receivers were changed to pointer receivers, usually this was as simple as changing the signature. Any superfluous receivers were also removed.

### Steps for Testing

Nothing should have changed.


alexanderstephan (approved):

alexanderstephan:
@Mjaethers thank you very much for the contribution! Sadly, it took some time, but better late than never ...

PR #1328, opened 2024-02-09
1032 option to make beta default mode

### LINKED ISSUE #1032: Option to make "beta" mode default when opening a stream
I personally like the beta mode more than the actual live mode and always have to switch into it normally

A toggle in the settings to make the beta mode the default when opening a stream (potentially with a warning or smth if the beta mode is actually less stable)

n/a

**Additional context**
n/a

Mjaethers:
### Motivation and Context

Closes #1032.
Many users prefer the beta stream mode, so it makes sense for there to be a user option that makes this the default.

### Description

Toggling the "Always use beta Stream when Available" option in the user settings will change the json value in the db for that user setting.
When the WatchPageData is being prepared this user setting is checked and if the default stream mode is set to beta, then the dvr value is set (in the same way it is if the URL contains the dvr query)

### Steps for Testing

Prerequisites:
- 1 Student

1. Log in
2. Navigate to Settings
3. Enable the "Always use beta Stream when Available" option, notice the value in the db has updated
4. Disable the option and the db will also have been updated

As for testing whether beta stream mode is actually used: that is something I cannot do locally.

### Screenshots

image

### DIFF
--- a/api/users.go
+++ b/api/users.go
@@ -28,6 +28,7 @@ func configGinUsersRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 	router.POST("/api/users/settings/seekingTime", routes.updateSeekingTime)
 	router.POST("/api/users/settings/customSpeeds", routes.updateCustomSpeeds)
 	router.POST("/api/users/settings/autoSkip", routes.updateAutoSkip)
+	router.POST("/api/users/settings/defaultMode", routes.updateDefaultMode)
 
 	router.POST("/api/users/resetPassword", routes.resetPassword)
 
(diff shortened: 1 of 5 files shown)


api/users.go
```diff
@@ -725,15 +726,33 @@ func (r usersRoutes) updateSeekingTime(c *gin.Context) {
 }
 
 func (r usersRoutes) updateAutoSkip(c *gin.Context) {
-	u := c.MustGet("TUMLiveContext").(tools.TUMLiveContext).User
-	if u == nil {
+	u := getUserFromContext(c)
+	var req struct{ Value model.AutoSkipSetting }
+	if err := c.BindJSON(&req); err != nil {
```
SebiWrn, line 739: I think we should still check if the user is null because else the method should already be returned

SebiWrn (changes requested):
Apart from than that minor changes it looks really good

api/users.go
Mjaethers: Yeah makes sense. This is the case for a lot of functions in this file so they should probably also be changed.

SebiWrn (approved):

PR #1348, opened 2024-04-22
Added headers to config for fixing ase.in.tum

SebiWrn:
### Motivation and Context
Fixed ticket of ase in tum

### Description
Added headers to config

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -7,3 +7,5 @@
 /dataSources.local.xml
 # Editor-based HTTP Client requests
 /httpRequests/
+# GitHub Copilot persisted chat sessions
+/copilot/chatSessions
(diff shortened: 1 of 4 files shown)


joschahenningsen:
Which headers do they need?

SebiWrn:
Access-Control-Allow-Origin with value ase.in.tum.de, because they told us else they can't edit the jframes on their website with javascript

KemalKrKX (approved):
Looks good.

joschahenningsen (commented):
This could also be handled using traefik https://doc.traefik.io/traefik/middlewares/http/headers/

SebiWrn:
Ok so we shouldn't do it with gin directly?

PR #1350, opened 2024-04-25
Watched state determined automatically

### LINKED ISSUE #1312: "watched" state is not being determined and set automatically
**Describe the bug**
in `view=3` the sorting option "Hide watched" does nothing for me. It neither hides 100% watched vods (according to the progress bar), nor does it hide almost fully watched vods (the actual lecture normally ends a few minutes before the vod ends)

**To Reproduce**
1. Go to any lecture with vods
2. Watch one of the vods
3. click "Hide watched"
4. see nothing

**Screenshots**
screencast
[gbs 23w](https://live.rbg.tum.de/?year=2023&term=W&slug=GBS&view=3)

 - OS: Ubuntu 23.10
 - Browser: Firefox 121

SebiWrn:
### Motivation and Context
Fixes #1312 

### Description
If progress > 90%, the vod is marked as watched

### DIFF
--- a/api/progress.go
+++ b/api/progress.go
@@ -117,31 +117,35 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
 		return
 	}
 
-	stream := tumLiveContext.Stream
+	stream, err := r.DaoWrapper.StreamsDao.GetStreamByID(c, strconv.FormatUint(uint64(request.StreamID), 10))
+	if err != nil {
+		return
+	}
 
 	watchedToLastSilence := false
 
-	if stream != nil {
-		duration := stream.Duration.Int32
-		if duration == 0 {
-			dur := stream.End.Sub(stream.Start)
-			duration += int32(dur.Seconds()) + int32(dur.Minutes())*60 + int32(dur.Minutes())*60*60
-		}
-		if duration != 0 {
-			lastSilence := slices.MaxFunc(stream.Silences, func(silence model.Silence, other model.Silence) int {
-				return int(silence.End) - int(other.End)
-			})
+	logger.Debug("Save progress")
+	duration := stream.Duration.Int32
+	duration = 0
+	if duration == 0 {
+		dur := stream.End.Sub(stream.Start)
+		duration += int32(dur.Seconds()) + int32(dur.Minutes())*60 + int32(dur.Minutes())*60*60
+	}
+	logger.Debug("Duration", "duration", duration)
+	if duration != 0 {
+		lastSilence := slices.MaxFunc(stream.Silences, func(silence model.Silence, other model.Silence) int {
+			return int(silence.End) - int(other.End)
+		})
 
-			// Add a little wiggle time to the end if ffmpeg didn't detect the silence till the end
-			if math.Abs(float64(lastSilence.End-uint(stream.Duration.Int32))) < 10 {
-				lastSilencePercent := float64(lastSilence.Start) / float64(stream.Duration.Int32)
-				if request.Progress >= lastSilencePercent {
-					watchedToLastSilence = true
-				}
+		// Add a little wiggle time to the end if ffmpeg didn't detect the silence till the end
+		if math.Abs(float64(lastSilence.End-uint(duration))) < 10 {
+			lastSilencePercent := float64(lastSilence.Start) / float64(duration)
+			if request.Progress >= lastSilencePercent {
+				watchedToLastSilence = true
 			}
 		}
-
 	}
+
 	progressBuff.add(model.StreamProgress{
 		Progress: request.Progress,
 		StreamID: request.StreamID,
--- a/api/progress.go
+++ b/api/progress.go
@@ -126,7 +126,6 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
 
 	logger.Debug("Save progress")
 	duration := stream.Duration.Int32
-	duration = 0
 	if duration == 0 {
 		dur := stream.End.Sub(stream.Start)
 		duration += int32(dur.Seconds()) + int32(dur.Minutes())*60 + int32(dur.Minutes())*60*60
--- a/api/progress.go
+++ b/api/progress.go
@@ -114,11 +114,21 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
 		})
 		return
 	}
-	progressBuff.add(model.StreamProgress{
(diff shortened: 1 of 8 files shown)


api/progress.go
```diff
@@ -114,11 +114,20 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
...
+			UserID:   tumLiveContext.User.ID,
+			Watched:  true,
+		})
+	} else {
+		progressBuff.add(model.StreamProgress{
+			Progress: request.Progress,
+			StreamID: request.StreamID,
+			UserID:   tumLiveContext.User.ID,
+		})
+	}
```
joschahenningsen: Nit:
```suggestion
	progressBuff.add(model.StreamProgress{
		Progress: request.Progress,
		StreamID: request.StreamID,
		UserID:   tumLiveContext.User.ID,
		Watched:  request.Progress > 0.9,
	})
```

joschahenningsen (approved):
Nice! This is long overdue 🥳 

If you want, you can follow up on this by also setting `watched: true` if the progress<90% but rest of the video is a silence segment

SebiWrn:
I fixed another bug, that the watched state isn't set in the database and I added, that watched is set if only silence is after the current timestamp.

api/progress.go
```diff
@@ -114,10 +116,37 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
 		})
 		return
 	}
+
+	streamIDString := strconv.FormatUint(uint64(request.StreamID), 10)
+	if r.StreamsDao == nil {
+		logger.Error("StreamsDao is nil")
```
joschahenningsen: This should not be nil, if it is, there is an issue. Did you encounter that?

api/progress.go
```diff
@@ -114,10 +116,37 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
...
+		return
+	}
+	stream, err := r.StreamsDao.GetStreamByID(c, streamIDString)
```
joschahenningsen: Is that stream not part of `tumLiveContext` already?

api/progress.go
```diff
@@ -114,10 +116,37 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
...
+
+	if err == nil {
+		lastSilence := slices.MaxFunc(stream.Silences, func(silence model.Silence, other model.Silence) int {
+			return int(silence.End) - int(other.End)
+		})
+
+		if math.Abs(float64(lastSilence.End-uint(stream.Duration.Int32))) < 10 {
```
joschahenningsen: I like the idea of having a little wiggle room here, please add a comment so people don't get confused about what that does

api/progress.go
```diff
@@ -114,10 +116,37 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
...
+			lastSilencePercent := float64(lastSilence.Start) / float64(stream.Duration.Int32)
+			if request.Progress >= lastSilencePercent {
+				progressBuff.add(model.StreamProgress{
+					Progress: request.Progress,
+					StreamID: request.StreamID,
+					UserID:   tumLiveContext.User.ID,
+					Watched:  true,
+				})
+				return
```
joschahenningsen: rather don't return but, assign to a var `watchedToLastSilence` or something here and write to buffer `Watched: request.Progress > .9 || watchedToLastSilence` later to avoid duplication

api/progress.go
```diff
@@ -114,10 +116,37 @@ func (r progressRoutes) saveProgress(c *gin.Context) {
...
+	}
+	stream, err := r.StreamsDao.GetStreamByID(c, streamIDString)
+
+	if err == nil {
+		lastSilence := slices.MaxFunc(stream.Silences, func(silence model.Silence, other model.Silence) int {
+			return int(silence.End) - int(other.End)
+		})
+
+		if math.Abs(float64(lastSilence.End-uint(stream.Duration.Int32))) < 10 {
+			lastSilencePercent := float64(lastSilence.Start) / float64(stream.Duration.Int32)
```
joschahenningsen: can we be extra careful of stream.Duration not being 0? I can imagine ffmpeg fucked this up for some streams

dao/progress.go
```diff
@@ -30,12 +30,40 @@ func (d progressDao) GetProgressesForUser(userID uint) (r []model.StreamProgress
...
+	return result
+}
+
 // SaveProgresses saves a slice of stream progresses. If a progress already exists, it will be updated.
 func (d progressDao) SaveProgresses(progresses []model.StreamProgress) error {
-	return DB.Clauses(clause.OnConflict{
-		Columns:   []clause.Column{{Name: "stream_id"}, {Name: "user_id"}}, // key column
-		DoUpdates: clause.AssignmentColumns([]string{"progress"}),          // column needed to be updated
-	}).Create(progresses).Error
+	noWatched := filterProgress(progresses, false)
```
joschahenningsen, line 46: Please help me understand, why is this change required? Can't we just add `watched` to the `assignmentColumns`?

joschahenningsen (dismissed):

api/progress.go
SebiWrn: I tested in while running gocast without any issues, but the progress test failed, because it was nil during the test

dao/progress.go
SebiWrn: If I just add it to the columns it will overwrite the existing value. If the user already watched the whole VoD, but then skips to the middle, to rewatch something, the whole VoD will be marked as unwatched

api/progress.go
SebiWrn: We should keep that in mind but with the new fixes this isn't an issue anymore

SebiWrn:
Nvm, I have to have a look into it, in the tumLiveContext, the stream is null

dao/progress.go
joschahenningsen: makes sense, thanks! can you add that as a comment as well?

joschahenningsen:
> Nvm, I have to have a look into it, in the tumLiveContext, the stream is null

Probably because there is no :streamID in the url of that endpoint, right?

SebiWrn:
I fixed it by adding a MockDao to the test. Now everything should be working fine

DawinYurtseven (approved):
looks good to me just change the return type of the two errors to one of them being goroutine so we can get both errors if they exist.

PR #1361, opened 2024-06-12
Fix Golangci-lint errors

SebiWrn:
### Motivation and Context
This PR should fix the Issues with the golangci


SebiWrn:
golangci-lint now working on github. Issue was that there was no node_modules folder, but the go router referenced it with a //go:embed link. Golangci-lint now first installs all node packages, which creates the node_modules folder and then tests with the linters.

joschahenningsen (approved):
Thanks! 🍻

Issue #1193, opened 2023-10-17
Add location to "Scheduled" lectures on course page


MatthiasReumann:
**Describe the solution you'd like**
Add a clickable link to the location (e.g. HS2) of "Scheduled" lectures on the course page referring to NavigaTUM. 

![Screenshot 2023-10-17 at 09 51 09](https://github.com/TUM-Dev/gocast/assets/29633518/b8de3042-f0c5-47f3-94ff-1bdea8de7520)

mohanadkandil:
Do normally classes provide the location with the stream ? cause I have some classes with no info about location (just time). 
Indeed, I can start working on this

joschahenningsen:
Classes without locations associated to them are usually self streams. (using OBS, etc.)

KemalKrKX:
I would like to work on this issue, if no one actively is.

PR #1376, opened 2024-09-05
feat: add location to "Scheduled" lectures on course page

KemalKrKX:
### Motivation and Context
Under scheduled lectures a new link to lecture hall webpage on NavigaTUM is added.
The issue: https://github.com/TUM-Dev/gocast/issues/1193

### Description
The lecture hall information of scheduled lectures is used to query NavigaTUM.

### Steps for Testing
Prerequisites:
- 1 Module

1. Log in
2. Create a new scheduled lecture for the module
3. Go to lecture homepage on TUMLive

Local copies of TUMLive don't update the lecture hall information, so no link is displayed there. However a reviewer can manually set a RoomCode for the lecture in order to observe the change.


web/template/home.gohtml
```diff
@@ -567,7 +567,9 @@
                         </section>
                     </template>
                     <template x-if="plannedStreams.hasElements()">
-                        <section class="tum-live-course-view-item px-3">
+                        <section x-data="{isLectureHallValid(lectureHall) {const regex = /^\d{4}\.[A-Z0-9]{2}\.[A-Z0-9]{3,4}$/;
```
SebiWrn: Please move this function into the corresponding typescript file. Especially functions shouldn't be in the x-data property as this should only contain data.

web/template/home.gohtml
```diff
@@ -365,7 +365,7 @@
                                                 <template x-if="livestream.InLectureHall()">
                                                     <a class="tum-live-badge bg-black"
                                                        target="_blank"
-                                                       :href="livestream.LectureHall.ExternalURL">
+                                                       :href="livestream.LectureHall.ExternalURL">s.
```
SebiWrn: What  is this "s." at the end of the line for?

KemalKrKX: Oops, it seems i overlooked it. I will correct this asap.

web/template/home.gohtml
KemalKrKX: I created a lecture-hall-validator.ts in the /web/ts/utilities, but I am having problems determining where Alpinejs is initialized so that i can import and use this validator. Do you happen to know where it is?

KemalKrKX:
There were many conflicts in go.mod and web/package-lock.json, so i kept the values in my branch and deleted the old values from dev. Now it fails the tests but works as intended on my end.

web/ts/utilities/lecture-hall-validator.ts
```diff
@@ -0,0 +1,12 @@
+export function isLectureHallValid(lectureHall: string): boolean {
+    const regex = /^\d{4}\.[A-Z0-9]{2}\.[A-Z0-9]{3,4}$/;
+    return regex.test(lectureHall);
+}
+
+declare global {
```
SebiWrn: Why do we need this declare global here? Wouldn't it be enough to just export the function and use it?

KemalKrKX: Theoretically it should suffice, but my local copy would not use this function unless I made it global. I can try making it work without the global declaration if it is an issue.

SebiWrn: We shouldn't have unnecessary global definitions so it would be better to change it

KemalKrKX:
Couldn't get the validator from a separate utility file to work with the website template even after testing various methods. Moved the script inside the home.gohtml instead. Hope this is alright.

SebiWrn:
I'm sorry but we just have to put that code into the .ts files.
Why couldn't you get it to these files, what's your error?

KemalKrKX:
I am not entirely sure what the exact cause is, but when I export the function to a seperate file the check breaks. A global definition solves this, so i believe the file/function is out of scope, however every bypass i could think until now caused further problems.

web/template/home.gohtml
```diff
@@ -566,6 +567,7 @@
                             </section>
                         </section>
                     </template>
+                    <div x-data="{ isLectureHallValid }">
```
SebiWrn: Why is this data Tag necessary? I think you don't need it.

SebiWrn (changes requested):

web/template/home.gohtml
KemalKrKX: Yeah, it is part of another change that i scrapped and it seems i forgot to delete it later on.

SebiWrn:
Is the update of the package-lock.json really needed? If not, we should do this in a different PR so we can locate errors in the git log more easily.
If this didn't happen on purpose, please use ```npm ci``` instead of ```npm i``` because then the package-lock isn't changed.

KemalKrKX:
This is a byproduct of me using `npm i` to recompile tailwind. After checking what changed, I saw they were mostly dependency changes to newer versions (saying mostly because to be honest I have not checked each and every one) and gocast was working so I left them as is. However as far as I understand, after this PR is merged, a new installation would change this file again anyway since `npm i` is used during the setup process. Would you rather prefer these changes to be rolled back entirely or transferred to a new PR?

SebiWrn:
I think we can rollback these changes and if we need we can do a full upgrade of every dependency in our project in another PR sometime.
Btw ```npm ci``` basically does the same as ```npm i``` but doesn't change the dependencies.

KemalKrKX:
The `package-lock.json` file is replaced with its counterpart from `dev` branch. However I also noticed some small files have also been changed indirectly, so I pushed them too to be safe. The only commands I have used that may have caused this are `go get ./...`,  `npm audit fix` and `npm fund`. I can also reset them to their `dev` counterparts if need be.

web/template/home.gohtml
SebiWrn: But now you again included these lines.

SebiWrn (changes requested):

web/template/home.gohtml
KemalKrKX: Sorry, I forgot I changed these lines on Github directly and when I pushed my local it must have overridden the change.

SebiWrn (approved):

PR #1380, opened 2024-09-20
Search Box to search for courses, streams and subtitles

### LINKED ISSUE #1337: Search box in "My Courses" and "Public Courses"
One should not rely on the browsers search functionality to filter the courses list. A search box would be useful.

SebiWrn:
### Motivation and Context
This Change solves the problem of finding lectures in a large list of lectures. It also enhances the search for subtitles or streams in courses.
Closes  #1337.

### Description
The change adds a search bar at the top of every page where you can search for courses or contents of a course, depending if you're on a course page or on the home page.

### Steps for Testing
You can't test this change on the Testserver system.

You have to set up your own local instance of TUM-Live for testing and also have to setup your own meilisearch service.
To start the meilisearch service, you can run the following command:
```bash
docker run -it --rm \
    -p 7700:7700 \
    getmeili/meilisearch:v1.9 \
```

If the data isn't automatically exported to meilisearch, you can run this job from the ```Maintenance``` Tab in the Admin panel.

You can enter your search query in the search bar on every page or you can press 'Enter' while the search bar is focused to see the full search page with filters etc.


model/semester.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	if semesters == nil {
+		if firstSemester.Year == lastSemester.Year && firstSemester.TeachingTerm == lastSemester.TeachingTerm {
+			return s.Year == firstSemester.Year && s.TeachingTerm == firstSemester.TeachingTerm
+		}
+		return s.GreaterEqualThan(firstSemester) && lastSemester.GreaterEqualThan(*s)
+		/* Alternative:
+		return !(s.Year < firstSemester.Year || s.Year > lastSemester.Year ||
+		(s.Year == firstSemester.Year && s.TeachingTerm == "S" && firstSemester.TeachingTerm == "W") ||
+		(s.Year == lastSemester.Year && s.TeachingTerm == "W" && lastSemester.TeachingTerm == "S"))*/
```
karjo24: Delete

model/semester.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	if len(semesters) == 1 {
+		return s.Year == semesters[0].Year && s.TeachingTerm == semesters[0].TeachingTerm
+	}
```
karjo24: delete

model/semester.go
SebiWrn: Done

model/semester.go
SebiWrn: Done

api/search.go
```diff
@@ -1,28 +1,618 @@
...
 
 	searchGroup := router.Group("/api/search")
+	searchGroup.GET("", routes.search)
 	withStream := searchGroup.Group("/stream/:streamID")
 	withStream.Use(tools.InitStream(daoWrapper))
 	withStream.GET("/subtitles", routes.searchSubtitles)
+
+	/*withCourse := searchGroup.Group("/course/:courseID")
+	withCourse.Use(tools.InitCourse(daoWrapper))
+	//withCourse.GET("/streams", routes.searchStreams)*/
```
kordianbruck: Please don't commit commented code.

api/search.go
```diff
@@ -1,28 +1,618 @@
...
+	}
+	return fmt.Sprintf("streamID IN %s", uintSliceToString(streamIDs))
+}
+
+func meiliStreamFilter(c *gin.Context, user *model.User, semester model.Semester, courses []model.Course) string {
+	if courses != nil {
+		return fmt.Sprintf("courseID IN %s", courseSliceToString(courses))
+	}
+
+	semesterFilter := fmt.Sprintf("(year = %d AND semester = \"%s\")", semester.Year, semester.TeachingTerm)
```
kordianbruck, line 366: This looks awfully a lot like SQL injection would be possible here? Please correct me if I'm wrong, but ugh.

api/search.go
```diff
@@ -1,28 +1,618 @@
...
+	}
+	idsStringSlice := make([]string, len(ids))
+	for i, id := range ids {
+		idsStringSlice[i] = strconv.FormatUint(uint64(id), 10)
+	}
+	filter := "[" + strings.Join(idsStringSlice, ",") + "]"
+	return filter
+}
+
+func ToSearchCourseDTO(cs ...model.Course) []SearchCourseDTO {
```
kordianbruck, line 593: Exported functions should have a comment.

api/search.go
karjo24: I'm currently on vacation and will address this in detail either tomorrow or next week. Short answer: User/client input is checked, so Injections are not possible, if these checks work as intended.

karjo24: This function expects all parameters to be "friendly". Under these circumstances this means that the slice `courses` may only consist of courses that the user is allowed to search through and all semesters consisting of a base-10 four digit `year` and the `teachingTerm` being either "W" or "S" and nothing else. I added comments to clarify that these checks are the caller's responsibility.

When looking at the usages, the slice `courses` is either `nil` or the return value of `parseCourses`, which retrieves the returned courses from the database. If there's an SQL injection possible up to this point, there'd be another problem. The returned courses are then checked here:
https://github.com/TUM-Dev/gocast/blob/94dae0434a527817865a9942acd120b2cece3348/api/search.go#L100-L105

Similarly the `semester` is either of zero value or the return value of `parseSemesters`. `parseSemesters` ensures that the semesters from the URL parameter conform to the format mentioned at the beginning using this regex
https://github.com/TUM-Dev/gocast/blob/94dae0434a527817865a9942acd120b2cece3348/api/search.go#L481

Am I missing something or misunderstanding you?

karjo24: However, there's another mistake here:
https://github.com/TUM-Dev/gocast/blob/94dae0434a527817865a9942acd120b2cece3348/api/search.go#L352-L354 This ignores the case, where the user is not allowed to see private streams. However, this mistake does not lead to data leaks, which is why it did not come up during testing. Even if Meilisearch returns streams the user is not allowed to see, this results will get thrown out by `checkAndFillResponse`.
I'll fix it though.

api/search.go
```diff
@@ -1,28 +1,632 @@
...
+	return fmt.Sprintf("streamID IN %s", uintSliceToString(streamIDs))
+}
+
+// meiliStreamFilter returns a filter conforming to MeiliSearch filter format that can be used for filtering streams
+//
+// Checking eligibility to search for courses and validation of model.Semester format is the caller's responsibility
+func meiliStreamFilter(c *gin.Context, user *model.User, semester model.Semester, courses []model.Course) string {
+	if courses != nil {
+		return fmt.Sprintf("courseID IN %s", courseSliceToString(courses))
+	}
```
karjo24: This ignores the case, where the user is not allowed to see private streams. Even if this does not lead to data leaks because of `checkAndFillResponse`, this should be dealt with.

karjo24: Commit 8295e423d7108f8e8b3393aa92c70c102ba26524 solves that issue.

api/search.go
karjo24: Regarding the latest comment: See https://github.com/TUM-Dev/gocast/pull/1380#discussion_r1781055089

tools/meiliSearch.go
```diff
@@ -21,3 +34,89 @@ func SearchSubtitles(q string, streamID uint) *meilisearch.SearchResponse {
...
+		Filter:               courseFilter,
+		AttributesToRetrieve: []string{"name", "slug", "year", "semester"},
+	}
+	return req
+}
+
+// Search passes search requests on to MeiliSearch instance and returns the results
+//
+// searchType specifies bit-wise which indexes should be searched (lowest bit set to 1: Index SUBTITLES | second-lowest bit set to 1: Index STREAMS | third-lowest bit set to 1: Index COURSES)
+func (d *meiliSearchFunctions) Search(q string, limit int64, searchType int, courseFilter string, streamFilter string, subtitleFilter string) *meilisearch.MultiSearchResponse {
```
CommanderStorm, line 74: FYI (This should definitively not be done in this PR, if at all):
what might simplify the codebase in the frontend is the new federation support in `meilisearch 1.10.0`.
- https://github.com/meilisearch/meilisearch-go/issues/573
Federations reorder the search results amongst each all search results, not just in the `MultiSearchResponse`.

Currently the solution is to split this into different boxes in the frontend.

tools/meiliExporter.go
```diff
@@ -106,6 +123,32 @@ func (m *MeiliExporter) Export() {
...
+	_, err = coursesIndex.DeleteAllDocuments()
+	if err != nil {
+		logger.Warn("could not delete all old courses", "err", err)
+	}
+
+	m.d.CoursesDao.ExecAllCourses(func(courses []dao.Course) {
+		meilicourses := make([]MeiliCourse, len(courses))
+		courseIDs := make([]uint, len(courses))
+		for i, course := range courses {
+			courseIDs[i] = course.ID
```
CommanderStorm: This seems to be dead code or at least debugging code

```suggestion
		for i, course := range courses {
```

tools/meiliExporter.go
```diff
@@ -130,4 +180,61 @@ func (m *MeiliExporter) SetIndexSettings() {
...
+	res := make([]MeiliStream, len(streams))
+	for i, s := range streams {
+		c, err := daoWrapper.GetCourseById(context.Background(), s.CourseID)
+		if err != nil {
+			return nil, err
+		}
+		courseName := c.Name
+		year := c.Year
+		teachingTerm := c.TeachingTerm
+		visibility := c.Visibility
```
CommanderStorm: nit: This seems like testing code => please inline

api/search.go
hmelder: > if these checks work as intended.

Yea... I really don't like seeing query construction using string formatting regardless of prior checks, but it seems like (please correct me if I'm wrong) mailisearch-go does not have prepared statements, which is utterly rediculous.

web/ts/search.ts
```diff
@@ -0,0 +1,264 @@
...
+                this.open = false;
+            }
+        },
+    };
+}
+
+export function initPopstateSearchBarListener() {
+    console.log("Initialized popstate listener");
+    document.body.addEventListener("click", (event) => {
+        setTimeout(() => {}, 50);
```
CommanderStorm: This does nothing. Is this intentional?

```suggestion
```

api/search.go
CommanderStorm: > which is utterly rediculous

There is the option to use recursive arrays for filters which offers a bit more breakout preventions: https://www.meilisearch.com/docs/learn/filtering_and_sorting/filter_expression_reference#creating-filter-expressions-with-arrays

Thing is that the sdk needs to support this alternative and MS-go does not.

api/search.go
```diff
@@ -1,28 +1,644 @@
...
+	for i, res := range response.Results {
+		switch res.IndexUID {
+		case "STREAMS":
+			hits := res.Hits
+			res.Hits = []any{}
+			response.Results[i] = meilisearch.SearchResponse{}
+
+			var meiliStreams []SearchStreamDTO
+			temp, err := json.Marshal(hits)
+			if err != nil { // shouldn't happen
```
CommanderStorm, line 217: There is an awfull lot of ignoring errors here and in other parts.
Logging these errors might be helpfull..

api/search.go
karjo24: Makes sense.
> but it seems like (please correct me if I'm wrong) mailisearch-go does not have prepared statements

To my knowledge that is correct.

dao/courses.go
```diff
@@ -281,6 +282,32 @@ func (d coursesDao) GetCourseAdmins(courseID uint) ([]model.User, error) {
...
 }
 
+type Course struct {
+	Name, Slug, TeachingTerm, Visibility string
+	ID                                   uint
+	Year                                 int
+}
+
+// ExecAllCourses executes f on all courses.
+func (d coursesDao) ExecAllCourses(f func([]Course)) {
```
CommanderStorm, line 295: This function has the assumption that the courses never change during iteration (as using offset + limit for pagination).
This needs to be documented.

You can (and if you want to stick with offset/limit should) reduce this assumption via using [Keyset Pagination](https://www.dashbase.ai/blog/pagination-sql-limit-offset).

Also the name and description should include that this is batched to avoid a footgun for future devs which could expect that this is executed on all courses ^^

dao/courses.go
```diff
@@ -281,6 +282,32 @@ func (d coursesDao) GetCourseAdmins(courseID uint) ([]model.User, error) {
...
+	var res []Course
+	batchNum := 0
+	batchSize := 100
+	var numCourses int64
+	DB.Model(&model.Course{}).Count(&numCourses)
+	for batchSize*batchNum < int(numCourses) {
+		err := DB.Raw(`SELECT id, name, slug, year, teaching_term, visibility
+							FROM courses 
+							WHERE deleted_at IS NULL
+							LIMIT ? OFFSET ?`, batchSize, batchNum*batchSize).Scan(&res).Error
```
CommanderStorm: Wondering: are the courses really many as to require such small pagination and pagination in the first place?
I would assume, that either
- just querying all and then use meilisearches `AddDocumentsInBatches`
- or using [gorms batching](https://gorm.io/docs/advanced_query.html#FindInBatches) would be better

All of the data should be short strings, which means that they can be loaded into memory without the need for this => `AddDocumentsInBatches` seems like a much simpler solution, right?

CommanderStorm: Please add a `ORDER BY ID` to avoid any undefined behaviour with limit/offset based pagination.

model/semester.go
```diff
@@ -0,0 +1,33 @@
...
+	TeachingTerm string
+	Year         int
+}
+
+// InRangeOfSemesters checks if s is between firstSemester (inclusive) and lastSemester (inclusive) or is element of semesters slice
+func (s *Semester) InRangeOfSemesters(firstSemester Semester, lastSemester Semester, semesters []Semester) bool {
+	if s == nil {
+		return false
+	}
+	if semesters == nil {
```
CommanderStorm: given that this quite subtely splits the controlflow (`firstSemester`/`lastSemester` is only checked if no `semesters` exist, which would not be obvious from the docsstring), using two functions for the different functionality where previously `InRangeOfSemesters` is used would make sense.

model/semester.go
```diff
@@ -0,0 +1,33 @@
+package model
+
+type Semester struct {
+	TeachingTerm string
+	Year         int
+}
+
+// InRangeOfSemesters checks if s is between firstSemester (inclusive) and lastSemester (inclusive) or is element of semesters slice
+func (s *Semester) InRangeOfSemesters(firstSemester Semester, lastSemester Semester, semesters []Semester) bool {
+	if s == nil {
```
CommanderStorm: nit: these emptyness (`s` can never be actualy `nil`) checks inside the function seem a bit strange in terms of API design. Are they intentional?

api/search.go
```diff
@@ -1,28 +1,644 @@
...
+type MeiliSearchMap map[string]any
+
 func (r searchRoutes) searchSubtitles(c *gin.Context) {
 	s := c.MustGet("TUMLiveContext").(tools.TUMLiveContext).Stream
 	q := c.Query("q")
-	c.JSON(http.StatusOK, tools.SearchSubtitles(q, s.ID))
+	c.JSON(http.StatusOK, r.m.SearchSubtitles(q, s.ID))
+}
+
+/*
```
CommanderStorm, line 52: nit: This comment seems sort of random and out of place.
- it's written in german and
- does not follow the style of the other doc-comments

=> is this intentional?

CommanderStorm (dismissed):
Hi, you were likely not expecting a review from me, but @joschahenningsen told me that meilisearch is happening and I thought that looking over the usage might be benefitial.

This is an incredibly large PR.
Reviewing this is impossible, anybody who tells you that they reviewed it is likely lying on at least some part.

=> Please in the future split your PRs way before the +2,513 lines mark.

web/ts/search.ts
SebiWrn: The following code should be in the method body.

model/semester.go
SebiWrn: Actually s can be nil so this check is necessary.

tools/meiliSearch.go
SebiWrn: What do you mean with "split it into different boxes in the frontend"?

CommanderStorm: In this code here

https://github.com/TUM-Dev/gocast/blob/cae868bde52013dd403d349c0a359804f9512380/web/template/search-page.gohtml#L266

you have different boxes (bit too lazy to get tum.live online to show visually)

image

Ideally, these boxes should not be there, as relevancy of subtitles can be higher than courses.
Have also not done that refactoring for https://nav.tum.de ^^

model/semester.go
karjo24: I assume you picked this as an example (you talk about "these checks" (plural)) and you are referring to the multiple times, this is done throughout this PR.

They are/were intentional. Generally, I added these to every pointer method in this PR to ensure that anyone who uses any method added in this PR does not need to worry about whether the pointer is nil or not.

In some parts, they are even crucial to this PR working as expected, e.g. here, because I don't ever check if the user is nil (not loggedin) or not:
https://github.com/TUM-Dev/gocast/blob/cae868bde52013dd403d349c0a359804f9512380/model/user.go#L279-L281

However, in the specific code lines you commented on and a few others, it may be better to just let it panic, as this does not handle any meaningful case.

Any more thoughts on this? If not, I would differentiate between methods where checking for nil pointer is necessary/adds a meaningful use case and methods where this was done purely for consistency reasons and delete these checks in the latter.

model/semester.go
karjo24: The straightforward solution in other programming languages would be to use method overloading. Since this is not possible in Go, I brought this up in our weekly meeting a while ago and iirc we agreed on this solution. 
Also consider that a change in this method would propagate upwards through the usage chain and require changes not only in this method + adding more complexity to more methods than just this one. 

From this background, is adding clarifying comments enough? Is there any best practice in Go I'm not aware of for legitimate (imo) use cases of method overloading?

tools/meiliExporter.go
karjo24: Yes, this is a relic of me copying old code without thinking about it too much. I won't commit your suggestion and instead delete this here too ;)
https://github.com/TUM-Dev/gocast/blob/af18af51725468c0b118c252e4927990ac0163c4/tools/meiliExporter.go#L76-L78

dao/courses.go
karjo24: I removed the pagination from this function as suggested in another comment of yours. This stems from me copying old code over without giving it much of a thought. I'll change this and the other issues in the following old function though:
https://github.com/TUM-Dev/gocast/blob/1bc9f8ececf5e6f463d2ae5ace6a29b025dad3d8/dao/streams.go#L220-L221

tools/meiliExporter.go
```diff
@@ -106,6 +121,30 @@ func (m *MeiliExporter) Export() {
...
+			meilicourses[i] = MeiliCourse{
+				ID:           course.ID,
+				Name:         course.Name,
+				Slug:         course.Slug,
+				Year:         course.Year,
+				TeachingTerm: course.TeachingTerm,
+				Visibility:   course.Visibility,
+			}
+		}
+		_, err := coursesIndex.AddDocumentsInBatches(&meilicourses, 500, "ID")
```
karjo24:
```suggestion
		_, err := coursesIndex.AddDocumentsInBatches(meilicourses, 500, "ID")
```

model/semester.go
CommanderStorm: > whether the pointer is nil or not

I hate that go does make discussing this so (unnessesarily) difficult..

The pointer can never be `nil`. The value it is pointing to can be default and thus equal to `nil`.

Having invalid input this deep into the application seems.. a bit like a smell..
But maybe that is just the rust person talking 😓

That was what I was commenting on ^^

dao/courses.go
karjo24: 2fdb45a41fea7b7cf87a9e4a73535e972e431872
Better?

dao/streams.go
```diff
@@ -212,39 +212,46 @@ func (d streamsDao) GetAllStreams() ([]model.Stream, error) {
...
+				ORDER BY streams.id ASC
+				LIMIT ? 
              	)
-			SELECT *, GROUP_CONCAT(subtitles, '\n') AS subtitles FROM sws GROUP BY sid;`, batchSize, batchNum*batchSize).Scan(&res).Error
+			SELECT *, GROUP_CONCAT(subtitles, '\n') AS subtitles FROM sws GROUP BY sid ORDER BY sid;`, lastSeenId, batchSize).Scan(&res).Error
 		if err != nil {
 			fmt.Println(err)
 		}
+		if err == nil {
+			lastSeenId = res[len(res)-1].ID + 1
```
CommanderStorm: I don't think `+1` is correct, as it is `WHERE streams.id > ?`

```suggestion
			lastSeenId = res[len(res)-1].ID
```

karjo24: Whoops, obviously, I really need to look more carefully before committing, thanks. It used to be >= before I checked whether primary keys can/should start at 0...

model/semester.go
karjo24: Suggestion is applied fc14c036963437eaefc5aaf0cef52b4313258585

dao/courses.go
CommanderStorm: A better fit for what this function currently does would be `ListCourses() ([]Course,error)`[^1]..

I think keeping it as `ExecAllCourses(f func([]Course))` is a bit like a smell / far-fetched..

Given that this now does not have the (idk how real they were) paging bugs / risks it had before, I would be "fine" with merging it.

[^1]: in case you don't know, google has published a bit of guidance on this: https://cloud.google.com/apis/design/naming_convention#method_names

model/semester.go
CommanderStorm: I am going to give you a bit of pushback on mushing 2 methods together as good style just because java somewhat does that.
Using overloading like hides the controllflow a lot.

Especially that the `firstSemester`/`lastSemester` parameter is unused for `InRangeOfSemesters` if `semesters` is set is likely not obvious..
I don't think the usage of this function understand this nuance.

Having them be one function, does loose the quite subtle difference

This is essentially these two functions (python to better reason about this):

```py
@dataclasses.dataclass # to get equality and constructors
class Semester:
  term: Literal["WS","SS"]
  term: int

  def in_range_of_semesters(self, semesters:  List[Semester]):
    return any(s == self for s in semesters)

  def in_range_of(first_inclusive: Semester, end_inclusive: Semester):
    direct_match = self == first_inclusive or self == end_inclusive
    semester_is_between_limits = self.is_after(first_inclusive) and not self.is_after(end_inclusive)
    return direct_matchs or semester_is_between_limits

  def is_after(self, other: Semester):
    same_year_but_term_after = other.year == self.year and self.term == "WS" and  self.term == "SS"
    return self.year > other.year or same_year_but_term_after
```

> [!TIP]
> in the end, my word does not count in this project.
> I have not put in the work
> => feel free to ignore 😉

CommanderStorm:
From my point of view, there is nothing left to review.
There are a few threads (like https://github.com/TUM-Dev/gocast/pull/1380#discussion_r1804770695) which are not quite resolved, but from my point nothing blocks merging.

model/semester.go
karjo24: I generally agree with what you see as problematic.

I did split the function into two (in the calling method as well). There are still parts, where this still is kind of mushed together, but only in non-exported functions in api/search.go, where I added comments. => Still far from optimal, but I guess, it's better when that is only the case in those parts of the code, where one probably has to look and understand the code anyway before doing something with it.

karjo24:
> From my point of view, there is nothing left to review.

@CommanderStorm Thanks for your review and suggestions, it was much appreciated! :)

joschahenningsen (approved):

PR #1388, opened 2024-10-20
Fixed non working admin download

### LINKED ISSUE #1192: Issues with lecture downloads
Both the course page and the watch page don't add (the right?) jwt to the url required for downloads.

### LINKED ISSUE #1381: Download as admin not working
If you're logged in as an admin and want to download a video, you can't.
If you try from the admin panel, you get the missing JWT error.
If you try from the "normal" user page, you get the "Forbidden, download not allowed" error.

SebiWrn:
### Motivation and Context
Admins should be able to download VODs from the admin panel.

### Description
Now every VOD that is displayed on the admin panel is already signed with a corresponding jwt token which allows downloading of the vods.


SebiWrn:
Some more changes than expected because worker was also reverted to last working version of worker so the latest version of the worker is also functional. Everything in the worker/ directory is just reverting changes.

joschahenningsen (changes requested):
There are a lot of unrelated changes, can you clean that up a bit?

SebiWrn:
Sorry for that, now it should be cleaned up

joschahenningsen (approved):

PR #1390, opened 2024-10-22
Enh/add runner workflow

SebiWrn:
Adds a new workflow the the github workflows so runners from the beta branch can be deployed

### DIFF
--- a/.github/workflows/deploy-runners.yml
+++ b/.github/workflows/deploy-runners.yml
@@ -1,0 +1,59 @@
+name: Build & Deploy Runners
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - '.*beta.*'
+
+permissions:
+  packages: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        module: ["runner"]
+
+    steps:
+      - name: get container name
+        run: |
+          echo "containername=${{ matrix.module }}" | sed 's/\//-/g' > $GITHUB_ENV
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Docker meta (${{ matrix.module }})
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ghcr.io/tum-dev/gocast/${{ env.containername }}
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+            type=sha
+          flavor: |
+            latest=true
+            prefix=
+            suffix=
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-runners.yml
```diff
@@ -0,0 +1,59 @@
...
+      matrix:
+        module: ["runner"]
+
+    steps:
+      - name: get container name
+        run: |
+          echo "containername=${{ matrix.module }}" | sed 's/\//-/g' > $GITHUB_ENV
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
CommanderStorm, line 25: - Let's use `checkout@v4` and
- add a comment wat this ref is for (I'd assume this or something equivalent should be the default, right?)

.github/workflows/deploy-runners.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set outputs
+        id: vars
+        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+      - name: Build and push Docker images (${{ matrix.module }})
+        uses: docker/build-push-action@v4
```
CommanderStorm, line 52: You might want to configure sbom and build provence here.
Example how I did this:
https://github.com/TUM-Dev/NavigaTUM/blob/main/.github/workflows/_docker-build.yml

(This does not have an immediate impact but allows more forensics for some attack scenarios for little to no maintenance-effort as far as I read up on)

.github/workflows/deploy-runners.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          context: ./${{ matrix.module }}
+          pull: true
+          push: true
+          build-args: version=${{ github.ref_name }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
CommanderStorm, line 59: You might want to configure `cache-from` and related parameters.
Here is an example how to do this: https://docs.docker.com/build/ci/github-actions/cache/#cache-mounts

Likely the builds are already really fast (looking at you, rust), but a few seconds for a few lines...

CommanderStorm (approved):
LGTM, have provided some enhancement proposals but no bugs/.. ^^

SebiWrn:
Thank you for the feedback, I think I'll deploy it anyway so we can deploy the beta system as fast as possible, but I'll keep your feedback in mind and open another PR for these changes so I can also change this on the current workers and tumlive deployment actions.

PR #1393, opened 2024-10-24
1291 split view progress bar in wrong position after entering fullscreen mode if the width ratio of two views adjusted

### LINKED ISSUE #1291: Split view: Progress bar in wrong position after entering fullscreen mode, if the width ratio of two views adjusted
**To Reproduce**
1. Go to any VoD (or live stream, in beta-stream mode) with both camera and presentation views
2. Switch to splitview mode
3. Adjust the width ratio by dragging the horizontal gutter
4. Enter fullscreen mode
5. See error
6. Exit fullscreen mode
7. The position of progress bar is still incorrect

**Screenshots**
<p align="center">

</p>

 - OS: Windows 10
 - Browser: Microsoft Edge
 - Version: 120.0.2210.91

karjo24:
### Motivation and Context
See Issue #1291.

### Description
Adjusting size of players in splitview when in fullscreen mode (or vice versa) doesn't result in the control bar being in the wrong position anymore. This applies to subtitles as well.
Currently, the control bar of the right video player is used as the control bar for both video players, the one on the left is hidden. This results in said weird issues as the control bar has to have (potentially high) negative left margins. As I couldn't figure out what exactly is the issue with updating the left margin, this is fixed by switching roles of the two control bars (left is "global" control bar, right is hidden). This may have side effects! Only side effect known to me is that the subtitle settings menu is now an overlay of the left video player (instead of the right player).

### Steps for Testing
Prerequisites:
- 1 Livestream / VOD (preferably with subtitles)

1. Navigate to a Livestream
2. Play around by adjusting sizes and toggling fullscreen mode. 
3. See no errors.

### DIFF
--- a/web/ts/splitview.ts
+++ b/web/ts/splitview.ts
@@ -1,6 +1,8 @@
 import { getPlayers } from "./TUMLiveVjs";
 import Split from "split.js";
 import { cloneEvents } from "./global";
+import videojs, {VideoJsPlayer} from "video.js";
+import PlayerOptions = videojs.PlayerOptions;
 
 const mouseMovingTimeout = 2200;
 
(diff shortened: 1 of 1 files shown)


karjo24:
> This may have side effects! Only side effect known to me is that the subtitle settings menu is now an overlay of the left video player (instead of the right player).

This one and only side effect is irrelevant imo and as far as I can tell, but I my have missed something. I'd guess, that there was a reason for choosing the non-straight-forward solution of making the right control bar the main control bar, but I haven't found anything plausible yet. This may be on me though. If anyone knows more, let me know.

carlobortolan (approved):
Looks good to me :+1:

PR #1398, opened 2024-11-04
Rollback to 1.4.15 which as every later version has errors during streaming

SebiWrn:
This PR rolls back everything in the worker to 1.4.15 as this is the latest working version of the worker


joschahenningsen:
What happened here? Was wip code committed to dev?

SebiWrn:
No but I think the worker wasn't really tested before pushing to dev because this should be mostly linter fixes.

joschahenningsen (approved):

PR #1414, opened 2024-11-24
Check audio codec on uploaded videos

### LINKED ISSUE #1336: Be more format agnostic
In order to be successful, this website should be more accessible to people who are not computer scientists. Therefore, it should not matter which kind of video someone Uploads. gocast should always try its best to convert input to a format that is recognised by popular browsers.

KemalKrKX:
### Motivation and Context

closes #1336

### Description

Audio codec is now also checked during filavigate to a Livestreame upload and if it is not in aac format it will be transcoded.

### Steps for Testing

1. Log in as prof or admin
2. Create a lecture by choosing upload video option
3. Upload a video with an audio codec that is not aac
4. After the video is available to watch, locate the file and ffprobe it to see the audio codec is changed


worker/worker/ffmegtools.go
```diff
@@ -14,12 +16,42 @@ func getDuration(file string) (float64, error) {
...
 }
 
-func getCodec(file string) (string, error) {
+func getVideoCodec(file string) (string, error) {
 	probe, err := probe(file)
 	if err != nil {
 		return "", err
 	}
-	return gjson.Get(probe, "streams.0.codec_name").String(), nil
+	codecNumber := gjson.Get(probe, "streams.#").Int()
```
joschahenningsen: I think a more descriptive name would be nice, something like 
```suggestion
	nStreams := gjson.Get(probe, "streams.#").Int()
```

worker/worker/ffmegtools.go
```diff
@@ -14,12 +16,42 @@ func getDuration(file string) (float64, error) {
...
+	videoIndex := -1
+	for i := 0; i < int(codecNumber); i++ {
+		if gjson.Get(probe, fmt.Sprintf("streams.%d.codec_type", i)).String() == "video" {
+			videoIndex = int(gjson.Get(probe, fmt.Sprintf("streams.%d.index", i)).Int())
+			break
```
joschahenningsen: Instead we could just return here

```suggestion
			return gjson.Get(probe, fmt.Sprintf("streams.%d.codec_name", i)).String(), nil
```

worker/worker/ffmegtools.go
```diff
@@ -14,12 +16,42 @@ func getDuration(file string) (float64, error) {
...
+		}
+	}
+	if videoIndex != -1 {
+		return gjson.Get(probe, fmt.Sprintf("streams.%d.codec_name", videoIndex)).String(), nil
+	}
+	return "", errors.New("no video stream found")
+}
+
+func getAudioCodec(file string) (string, error) {
```
joschahenningsen: I wonder if we could combine these methods into one, something like this:
```suggestion
// getCodec returns the codec of the file, codecType=video returns the video codec, codecType=audio the audio codec.
func getCodec(file string, codecType string) (string, error) {
```

joschahenningsen (changes requested):
Very nice!! I added a couple ideas on your code :)

worker/worker/ffmegtools.go
```diff
@@ -14,12 +16,29 @@ func getDuration(file string) (float64, error) {
...
+		}
+		return "", errors.New("no video stream found")
+	}
+	if codecType == "audio" {
+		for i := 0; i < int(nStreams); i++ {
+			if gjson.Get(probe, fmt.Sprintf("streams.%d.codec_type", i)).String() == "audio" {
+				return gjson.Get(probe, fmt.Sprintf("streams.%d.codec_name", i)).String(), nil
+			}
+		}
+		return "", errors.New("no audio stream found")
```
joschahenningsen: This should do the exact same, no?

```suggestion
	for i := 0; i < int(nStreams); i++ {
		if gjson.Get(probe, fmt.Sprintf("streams.%d.codec_type", i)).String() == codecType {
			return gjson.Get(probe, fmt.Sprintf("streams.%d.codec_name", i)).String(), nil
		}
	}
	return "", fmt.Errorf("no %s stream found", streamType)
```

KemalKrKX: Yeah you are right, this is much more compact than my version. I will push this change before today's meeting.

joschahenningsen (approved):
Niceee 🔥 Thanks!!

PR #1425, opened 2024-12-18
Test course feature for lecturers

### LINKED ISSUE #1418: Test courses for lecturers
Internally, TUM-Live already has test courses, but they need to be created by the admins.  
To allow lecturers to test their setup, a "Create Testcourse" feature would be nice to have. While working on this issue, I would also fix a current bug that shows courses on the start page.

A button on the admin page that allows lecturers and admins to create a personal test course.

\-

**Additional context**
\-

carlobortolan:
### Motivation and Context
Closes #1418.
While this approach is not optimal, as we need to "reserve" special values for test courses, it is relatively simple and builds on top of the existing code in comparison to more complex approaches (e.g., updating the current course model & table to include a `isTestCourse` attribute or adding a completely new test course model).

### Description
- Internally, test courses are handled just like regular courses but belong to a special "test-semester" (where the course's year equals `1234`).
- Updated semester API endpoint and course DAO methods to include an optional `includeTestSemester` flag, which defaults to false. This allows the endpoint & method to return the same information as before (all normal courses), and only if the flag is set, also include the test courses.
- Updated all current occurrences of the semester-fetching (e.g., on the start page, search page, and admin page) accordingly so that the test semesters and courses are not included there. 

### Steps for Testing
Prerequisites:
- 1 Lecturer
- 1 Student

1. Log in as a lecturer
2. Go to the admin panel and press on `+ Create Test Course` to create a new test course
3. The test course should now appear after reloading the page
4. Log out and log in as another user
5. Check that the previously created test course is not visible anywhere and that the semesters don't include the test semester

### Screenshots
How to create a test course:
image

How an Admin sees all test courses: 
image


api/courses.go
```diff
@@ -1451,6 +1452,21 @@ func (r coursesRoutes) createCourse(c *gin.Context) {
 	c.JSON(http.StatusCreated, gin.H{"id": courseWithID.ID})
 }
 
+func (r coursesRoutes) createTestCourse(c *gin.Context) {
+	tumLiveContext := c.MustGet("TUMLiveContext").(tools.TUMLiveContext)
+
+	_, err := r.StreamsDao.CreateOrGetTestCourse(tumLiveContext.User)
```
SebiWrn, line 1458: Should every user be able to create a test course? If every student can create a test course, they could possibly fill up the entire disk just by streaming nonsense

dao/streams.go
```diff
@@ -405,8 +406,8 @@ func (d streamsDao) CreateOrGetTestStreamAndCourse(user *model.User) (model.Stre
 func (d streamsDao) CreateOrGetTestCourse(user *model.User) (model.Course, error) {
 	var course model.Course
 	err := DB.FirstOrCreate(&course, model.Course{
-		Name:         "(" + strconv.Itoa(int(user.ID)) + ") " + user.Name + "'s Test Course",
-		TeachingTerm: "Test",
+		Name:         user.GetPreferredName() + "'s Test Course",
```
SebiWrn: Maybe we should have a fallback if the users preferred name is empty

web/admin.go
```diff
@@ -221,7 +231,7 @@ func (r mainRoutes) CourseStatsPage(c *gin.Context) {
 		logger.Error("couldn't query courses for user.", "err", err)
 		courses = []model.Course{}
 	}
-	semesters := r.CoursesDao.GetAvailableSemesters(c)
+	semesters := r.CoursesDao.GetAvailableSemesters(c, false)
```
SebiWrn: Should we allow to see stats of the test courses? Maybe with that users could also test this stats functionality?

web/admin.go
```diff
@@ -87,8 +87,17 @@ func (r mainRoutes) AdminPage(c *gin.Context) {
 			logger.Warn("could not get all server notifications", "err", err)
 		}
 	}
-	semesters := r.CoursesDao.GetAvailableSemesters(c)
+	semesters := r.CoursesDao.GetAvailableSemesters(c, true)
 	y, t := tum.GetCurrentSemester()
+
```
SebiWrn: Duplicate with line 284, this should be another function so the test course check can be adjusted if we change the implementation of test courses

SebiWrn (changes requested):

api/courses.go
carlobortolan: Only lecturers should be allowed to create test courses, but this should already be the case here, no?

See:
https://github.com/TUM-Dev/gocast/blob/68e73553a9a80e8d3e353f0e3fea5bfbf8eff0a1/api/courses.go#L51
and:
https://github.com/TUM-Dev/gocast/blob/68e73553a9a80e8d3e353f0e3fea5bfbf8eff0a1/api/courses.go#L54

dao/streams.go
carlobortolan: Thanks for pointing that out; I updated it so that it defaults to `Test Couse` if a user has a blank name. 

~_Edit:_ On a side note, is it an issue for the workers/edge server if all test courses have the same slug? (/should the slug be unique for all courses?) From what I was able to test locally, this is not an issue in the web client.~
_Edit 2:_ Updated it so that the course slug is structured as follows: `TEST-<unique-hash>` (e.g., `TEST-6b86b273ff34fce19d6b804eff5a3f5747ada4eaa22f1d49c01e52ddb7875b4b`)

web/admin.go
carlobortolan: Done :white_check_mark:

web/admin.go
carlobortolan: I don't think there should be any problem if we allow lecturers to see stats for test courses; updated it in my latest commit :+1:

api/courses.go
SebiWrn: Oh sorry I didn't look into the path mapping. Yes it's the case

SebiWrn (approved):
Looks good :)

Issue #1351, opened 2024-04-26
A pinned course that is no longer public cannot be unpinned.


DominikRemo:
**Bug**
A pinned course that is no longer public cannot be unpinned.

**To Reproduce**
1. Pin any course. The course is now listed in pinned courses
2. Delete or make the course private (I'm not sure if the course was deleted or set to private)
3. As a user, you cannot access the course. Therefor, you can also not unpin the course.

**Expected behavior**
Either remove unaccessible courses automatically from the pinned list or enable the user to unpin courses without accessing the course page.

**Desktop:**
 - OS: MacOS 14
 - Browser Chrome

SebiWrn:
Can't replicate this behaviour, can you please describe it a bit further or add screenshots to your problem?

DominikRemo:
The video shows that I cannot access the "Fortgeschrittene Themen des Softwaretests (IN2084)" course. It is still listed under pinned courses, but when I try to access it I am redirected to the home page. I can still access other courses and pin/unpin them.

https://github.com/TUM-Dev/gocast/assets/47261058/736b15a3-5b6d-4019-be8e-d5ec74698c52


EDIT: Unfortunately I don't really know how to reproduce this problem as I don't know what happened to the course.

PR #1428, opened 2025-01-06
1413 vis feedback

### LINKED ISSUE #1413: Deleting lectures doesn't give immediate visual feedback to the user
**Describe the bug**
When deleting lectures on the course admin page, the lectures are not immediately removed from the page. This might suggest to the user that it was not actually deleted and might lead them to retry. Retrying will fail with "An unknown error occured during the deletion process", probably as double deletion cannot occur. The lecture disappears when reloading the page.

**To Reproduce**
1. Go to course admin page
2. Click on 'Delete 1 Lecture'
3. Verify that the lecture has not disappeared
4. Click on 'Delete 1 Lecture' again
5. See error alert
6. Reload page and see that lecture has disappeared

**Expected behavior**
The lecture immediately disappears to avoid confusion.

KemalKrKX:
### Motivation and Context
https://github.com/TUM-Dev/gocast/issues/1413

### Description
The page refreshes after lecture deletion, giving visual feedback of changes

### Steps for Testing
Prerequisites:
- 1 Lecture
- 1 Admin

1. Log in
2. Create and delete a lecture
3. Page refreshes automatically, lecture does not stay on page


dao/courses.go
```diff
@@ -44,6 +44,8 @@ type CoursesDao interface {
 
 	RemoveAdminFromCourse(userID uint, courseID uint) error
 	DeleteCourse(course model.Course)
+
+	IsUserEnrolledInCourse(userID uint, courseID uint) (bool, error)
```
carlobortolan: Is this function necessary? It seems to be dead code, which is not called by other methods.

web/template/home.gohtml
```diff
@@ -13,6 +13,7 @@
 
     {{if and .VersionTag (eq .VersionTag "development")}}
     <script defer src="/static/node_modules/alpinejs/dist/cdn.js"></script>
+    <script defer src="/static/node_modules/alpinejs/dist/cdn.js"></script>
```
carlobortolan: This line is a duplicate of line 15 and can be removed.

web/ts/data-store/admin-lecture-list.ts
```diff
@@ -76,6 +76,7 @@ export class AdminLectureListProvider extends StreamableMapProvider<number, Lect
         await AdminLectureList.delete(courseId, lectureIds);
         this.data[courseId] = (await this.getData(courseId)).filter((s) => !lectureIds.includes(s.lectureId));
         await this.triggerUpdate(courseId);
+        window.location.reload();
```
carlobortolan: I might be missing something, but I think that this line is not necessary and not related to #1413.
If I understand it correctly, the issue refers to the `Delete ... Lecture(s)` button (see screenshot below), meaning that this line should be added not here but in [web/ts/edit-course.ts#L401](https://github.com/TUM-Dev/gocast/blob/1a5939d5fb38f0a3ae75367e63d9ed58c96fe792/web/ts/edit-course.ts#L401).

image

carlobortolan (changes requested):
The code looks good to me, but I think you might have implemented it for the wrong delete button (see review comments)

dao/courses.go
KemalKrKX: I did not change this file, I branched from dev and only modified web/ts/data-store.

web/template/home.gohtml
KemalKrKX: Will remove this line but I don't remember changing this file.

web/ts/data-store/admin-lecture-list.ts
KemalKrKX: I understood the issue as after a lecture is deleted the interface stays the same so you can still click delete lecture again. Clarification may be needed.

dao/courses.go
carlobortolan: Hm that's strange. The current dev branch doesn't have this code (see https://github.com/TUM-Dev/gocast/blob/dev/dao/courses.go)

KemalKrKX: Oh, it is related to another branch I work on. It seems I messed up and merged the files together. Will fix asap.

carlobortolan: ah ok, thanks :+1:

KemalKrKX: > Hm that's strange. The current dev branch doesn't have this code (see https://github.com/TUM-Dev/gocast/blob/dev/dao/courses.go)

Yeah it is from #1351.

web/ts/data-store/admin-lecture-list.ts
carlobortolan: I tried it locally (using the current dev branch) and deleting the lectures already reloads the lecture list when using this button: 
image

What currently doesn't work is using the `Delete ... lecture(s)` button (see previous message), which is what I think the issue refers to.

(@karjo24)

KemalKrKX: Alright, thanks for the heads-up. I will work on this today

carlobortolan (approved):
:+1:

PR #1434, opened 2025-01-10
Remove test deployments that have been broken

kordianbruck:
### Motivation and Context
Its apparently broken and we prefer testing changes locally first.


joschahenningsen (approved):
merci

PR #1476, opened 2025-01-13
Highlight private streams on stream overview

JosefSchoenberger:
### Motivation and Context
On the stream overview, course admins still see private streams. While I like this behavior, I would like some visual indication which streams are private and which are not.

### Description
For private streams, this adds a dashed eye icon next to the lecture name.

### Screenshots

Before:
image
image

After:
image
image


carlobortolan (approved):
:+1:

PR #1479, opened 2025-01-15
Upload Custom Thumbnail for a Lecture

### LINKED ISSUE #1155: Custom Thumbnail Upload
It should be possible to upload a custom thumbnail instead of the automatically generated one. Preferably on the admin-course-page.

feriel97:
### Motivation and Context
resolves #1155

### Description
Currently the thumbnail for the videos is generated automatically. The new Upload Custom Thumbnail button allows the admin users to be able to drag and drop an image file as a custom thumbnail for a lecture. The UI is still a work in progress and the file upload without drag and drop will be supported for better support on other devices such as mobiles. 

**The current to-dos:**
- [x] Visual Feedback for the upload to the user
- [x] Preview the thumbnail 

Optional: 
-  Option to go back to the automatically generated thumbnail 

### Steps for Testing

Prerequisites:
- 1 Lecturer
- 1 Lecture
- 1 Picture 

1. Log in
2. Navigate to the Admin Panel 
3. Select a Course
4. Select a Lecture
5. Edit Lecture 
6. Toggle the button 
7. Drag and drop the file 
8. Select Save Lecture

### Screenshots

UploadCustomThumbnailButton


api/stream.go
```diff
@@ -50,6 +50,8 @@ func configGinStreamRestRouter(router *gin.Engine, daoWrapper dao.DaoWrapper) {
 				thumbs.GET(":fid", routes.getThumbs)
 				thumbs.GET("/live", routes.getLiveThumbs)
 				thumbs.GET("/vod", routes.getVODThumbs)
+				thumbs.POST("/customlive", routes.putCustomLiveThumbnail)
```
joschahenningsen: Why do we need two routes to do the same thing? I'd argue there's no need to distinguish livestreams from vods when uploading thumbnails

joschahenningsen: Imo the logical method and path would be something like `POST /api/stream/:streamID/thumbnails`
```suggestion
				thumbs.POST("/", routes.putCustomLiveThumbnail)
```

api/stream.go
```diff
@@ -894,3 +896,45 @@ func (r streamRoutes) updateChatEnabled(c *gin.Context) {
...
+	tumLiveContext := c.MustGet("TUMLiveContext").(tools.TUMLiveContext)
+	streamID := tumLiveContext.Stream.ID
+	course := tumLiveContext.Course
+	file, err := c.FormFile("file")
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid file"})
+		return
+	}
+
+	filename := file.Filename
```
joschahenningsen, line 921: remove the filename if it's unused

api/stream.go
```diff
@@ -894,3 +896,45 @@ func (r streamRoutes) updateChatEnabled(c *gin.Context) {
...
 	}
 }
+
+func (r streamRoutes) putCustomLiveThumbnail(c *gin.Context) {
+	tumLiveContext := c.MustGet("TUMLiveContext").(tools.TUMLiveContext)
+	streamID := tumLiveContext.Stream.ID
+	course := tumLiveContext.Course
+	file, err := c.FormFile("file")
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid file"})
```
joschahenningsen: you can use our RequestError to abort:

```go
_ = c.AbortWithError(http.StatusForbidden, tools.RequestError{
	Status:        http.StatudBadRequest,
	CustomMessage: "Couldn't read file",
	Err:           err,
})
```

api/stream.go
```diff
@@ -894,3 +896,45 @@ func (r streamRoutes) updateChatEnabled(c *gin.Context) {
...
+
+	filename := file.Filename
+	fileUuid := uuid.NewV1()
+
+	filesFolder := fmt.Sprintf("%s/%s.%d/%s.%s/files",
+		tools.Cfg.Paths.Mass,
+		course.Name, course.Year,
+		course.Name, course.TeachingTerm)
+
+	path := fmt.Sprintf("%s/%s%s", filesFolder, fileUuid, filepath.Ext(filename))
```
joschahenningsen: please use path.Join from the standard library to construct this filepath:
https://pkg.go.dev/path#example-Join

api/stream.go
```diff
@@ -894,3 +896,45 @@ func (r streamRoutes) updateChatEnabled(c *gin.Context) {
...
+
+	//tempFilePath := pathprovider.LiveThumbnail(strconv.Itoa(int(streamID)))
+	if err := c.SaveUploadedFile(file, path); err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to save file"})
```
joschahenningsen: Use RequestError here too

api/stream.go
```diff
@@ -894,3 +896,45 @@ func (r streamRoutes) updateChatEnabled(c *gin.Context) {
...
+		return
+	}
+
+	thumb := model.File{
+		StreamID: streamID,
+		Path:     path,
+		Filename: file.Filename,
+		Type:     model.FILETYPE_THUMB_CAM,
```
joschahenningsen: iirc this filetype is for the scrubbar, not the preview. To be sure, use FILETYPE_THUMB_LG_CAM_PRES, which is always elected default

api/stream.go
```diff
@@ -894,3 +896,45 @@ func (r streamRoutes) updateChatEnabled(c *gin.Context) {
...
+	}
+
+	fileDao := dao.NewFileDao()
```
joschahenningsen: no need to create a new DAO, just use r.DaoWrapper.FileDao

joschahenningsen (changes requested):
Looks like a good start! This is only a draft but I thought I could leave some input already. 
In addition to my thoughts, we need a way during livestreams, to prevent the thumbnail from being overwritten by the automatomatically generated one. My suggestion is, creating a new FileType for costum thumbnails, which is preferred over automatically generated ones.

api/stream.go
feriel97: yes the two routes were originally there in case a distinction was necessary but now of course one route is enough.

api/stream.go
feriel97: filename is needed to create an object of type File as it has the element Filename as part of it.

api/stream.go
feriel97: I will change the type now temporarily but the idea is as you suggested to have a new type for custom thumbnail

kordianbruck:
@joschahenningsen ping please re-review?

joschahenningsen:
We ported your commits to https://github.com/TUM-Dev/gocast/pull/1731. Thanks for your contribution!

PR #1490, opened 2025-02-02
feat(runner): implement runner registration with gocast

joschahenningsen:
### Motivation and Context
As discussed, let's break down further development of the runner into smaller independent PRs. 

### Description
This PR implements the registration of the runner with gocast. This is achieved by:
- Cleaning up the gRPC services in runner.proto (`FromRunner` is now`RunnerManagerService`)
- Creating a new package, `runner_manager` in gocast, which implements `RunnerManagerService` -> gocast is the runners manager 
- Creating a new database model Runner with the fields hostname and port. The hostname is the primary key, on insert conflicts, the port is updated.
- On startup, the runner connects to this service to make itself known to gocast.

### Steps for Testing

1. start gocast
2. start runner
3. observe logs 
4. check database for inserted runner.

### Screenshots
image

grafik


runner/runner.proto
```diff
@@ -31,8 +31,8 @@ message StreamResponse {
   string jobID = 1;
```
carlobortolan: Not directly related to changes in PR, but I think we should try to keep the naming consistent and use snake_case for fields for both `StreamResponse` and `RegisterResponse` (see https://protobuf.dev/programming-guides/style/#field-names)

runner/serverhandler.go
```diff
@@ -29,19 +31,19 @@ func (r *Runner) RegisterWithGocast(retries int) {
 	}
 	_, err = con.Register(context.Background(), &protobuf.RegisterRequest{Hostname: ptr.Take(config.Config.Hostname), Port: ptr.Take(int32(config.Config.Port))})
 	if err != nil {
-		r.log.Warn("error registering with gocast", "error", err, "sleeping(s)", registerRetries-retries)
+		r.log.Error("error registering with gocast", "error", err, "sleeping(s)", registerRetries-retries)
```
carlobortolan, line 34: Currently, the process exits after 5 retries (~10s). Should we increase this or implement some exponential backoff for cases where a runner starts while the API is offline/unreachable? Or is the current approach sufficient?

carlobortolan (approved):
Just two small nits, but everything else looks good to me :+1: 
Tested the Runner registration locally and it works as expected (the deregistration doesn't work yet, but that's intended, right?)

runner/runner.proto
joschahenningsen: Absolutely! I applied the styleguide and also added a linter to the makefile.

runner/serverhandler.go
joschahenningsen: I'm applying a sort of exponential backoff retry pattern with the circut breaker being the service dying (and then being restarted by docker). 
I think that's best we can do right now. I can't come up with a sort of degraded mode the runner could run in when gocast is unavailable.

joschahenningsen:
> the deregistration doesn't work yet, but that's intended, right?

There is no deregistration, when we get to task scheduling on the runner, the last heartbeat should be considered to determine whether a runner is available. Heartbeat is not implemented yet though :D

carlobortolan:
> > the deregistration doesn't work yet, but that's intended, right?
> 
> There is no deregistration, when we get to task scheduling on the runner, the last heartbeat should be considered to determine whether a runner is available. Heartbeat is not implemented yet though :D

Ah ok; thanks for clarifying this :+1:

PR #1494, opened 2025-02-06
feat(runner): implement streaming functionality

joschahenningsen:
### Motivation and Context
This PR enables streaming functionality for the runner by cleaning up my example code 🎉 

### Description
Key changes:
- Add logic to main.go to await a subsequent signal to shutdown even if actions are still running (this is useful when testing locally. This way, a runner can be shut down while a stream is running by `ctrl`+`c`'nig twice)
  - when this happens, all running actions are cancelled via their respective contexts.
- Add `RequestStreamEnd` rpc that allows stopping a running stream.
- Add `StreamEnd` action that notifies gocast when a stream ends
- Fix issues with `AbortingError` so actions can properly indicate their doneness to the `RunAction` loop
- Properly log ffmpeg output
- Add a `WaitDelay` to the exec.Command so it can be killed if it doesn't adhere to the termination request in a timely manner

### Steps for Testing

Prerequisites:
- VPN connection to a network with a testing source, e.g. a SMP
- `grpcurl` for testing: https://github.com/fullstorydev/grpcurl?tab=readme-ov-file#installation

1. run gocast: `go run cmd/tumlive/tumlive.go`
2. run runner: `cd runner && go run cmd/runner/main.go`
3. send grpc call to trigger a stream (verify that `end` is in the future)

        grpcurl -plaintext -use-reflection -d '{"stream_id":1, "version": "STREAM_VERSION_COMBINED", "end": "2025-02-06T22:00:00Z", "ffmpeg_global_options": "-rtsp_transport tcp", "ffmpeg_output_options": "-c:a copy -c:v copy", "input": "rtsp://172.24.17.21/extron2"}' localhost:52735 protobuf.RunnerService/RequestStream

5. With the stream test:
    a. play the stream from the runners HLS server

        mpv http://localhost:8187/1/playlist.m3u8

    b. Stop the stream by calling this rpc (make sure to use the jobId from previous execution):

        grpcurl -plaintext -use-reflection -d '{"jobId": "6e0c320b-54da-43e9-8da5-3e34e0c5b97d"}' localhost:52735 protobuf.RunnerService/RequestStreamEnd

    c. stop the worker by hitting `ctrl`+`c` twice, while the stream is running
    d. kill ffmpeg and verify that the stream restarts and appends to the playlist:

        `killall ffmpeg`

### Screenshots

https://github.com/user-attachments/assets/1508e8fa-5575-42da-b4ea-d7fce8aae28e


DawinYurtseven (approved):
tested it locally and works well!

PR #1496, opened 2025-02-12
feat(runner/metrics): Add metric broker

joschahenningsen:
depends on #1494, merge after.

### Steps for Testing
only code review


runner/pkg/metrics/broker.go
```diff
@@ -0,0 +1,92 @@
...
+func WithPort(port int) Option {
+	return func(broker *Broker) {
+		broker.port = port
+	}
+}
+
+// Run starts an HTTP server that exposes the metrics.
+func (b *Broker) Run() {
+	http.Handle("/metrics", promhttp.Handler())
+	slog.Error("Serving metrics", "err", http.ListenAndServe(fmt.Sprintf(":%d", b.port), nil))
```
CommanderStorm, line 66: Is this really an `Error`?
Seems a bit sus, but would need an IDE to check further

joschahenningsen: yes, listen and serve blocks indefinitely and returns an error if something goes wrong (e.g. port in use)

runner/pkg/metrics/broker.go
```diff
@@ -0,0 +1,92 @@
...
+// LabelBuilder helps construct Prometheus labels dynamically.
+type LabelBuilder prometheus.Labels
+
+// L converts LabelBuilder to a Prometheus Labels map.
+func (b LabelBuilder) L() prometheus.Labels {
+	return prometheus.Labels(b)
+}
+
+// With initializes a new LabelBuilder.
+func (b *Broker) With() LabelBuilder {
```
CommanderStorm, line 78: nit: In rust this would be named

```suggestion
func (b *Broker) Builder() LabelBuilder {
```

CommanderStorm (approved):
LGTM once you have checked if the two nits are something or just false postives from FrankLint2000™️

runner/pkg/metrics/broker.go
joschahenningsen: I went for this because it allows using the api like so `b.With().Stream(123).Input("rtmp://1.2.3.4/src")`, which is shorter and quite naturally readable

Issue #1426, opened 2024-12-28
Saving the favourite view of the student.


rayannus:
**Is your feature request related to a problem? Please describe.**
No just a feature

**Describe the solution you'd like**
If I always watch using the presentation only view, why not save it so that I don't have to change it each time? Would be a really useful and simple feature to add. Thank you!

joschahenningsen:
Hey there, I like the idea! Some questions as to how this could be done:

- should it be a setting, the user can set or should we rather remember the last used preset and select it automatically? (similar to how we do it with the volume)
- Should this setting be different for each course? (e.g. in a maths class my prof uses the blackboard so I want the camera view to be default, in a programming class I need both views, so I choose the combined view as default)

Thanks for your input :)

rayannus:
Hi, nice suggestions!

I think as a first step, start with automatically saving the chosen preset for each lecture. This way, the user only changes the view once for each lecture, which I think is okay.

Once that is done, it can also be improved upon by combining both methods you mentioned. So basically allow the user to set a default view in the settings and that view can be overwritten manually for each lecture (from step 1).

feriel97:
@joschahenningsen can you assign me this issue please?

SebiWrn:
Default setting now available

PR #1498, opened 2025-02-12
Favorite Lecture View

feriel97:
### Motivation and Context
fixes issue #1426
### Description
I have added a user setting "Preferred Lecture View" in the global user settings that allows the user to choose his preferred lecture mode : split, combined, presentation... I have then according to the preference updated the view shown ie the URL redirected to the preferred view of the student.

### Steps for Testing

Prerequisites:
- 1 Student 
- 1 Lecture : lecture should have all 4 views available to see if the switch happens as expected

1. Log in
2. Navigate to Settings 
3. Choose preferred lecture view 
4. Go back to the main page and choose a course
5. Click on a lecture and check if the view that the student chose in the settings is actually being displayed. 

### Screenshots

preferred-view


web/template/partial/stream/actions.gohtml
```diff
@@ -61,7 +61,7 @@
                         {{if $stream.PlaylistUrl}}
                             <a class="text-left flex justify-start items-center w-full text-3 px-4 py-2 hover:cursor-pointer hover:bg-gray-200 dark:hover:bg-gray-600"
                                title="Combined view"
-                               @click="watch.switchView('/w/{{$course.Slug}}/{{$stream.Model.ID}}')">
+                               @click="watch.switchView('/w/{{$course.Slug}}/{{$stream.Model.ID}}/COMB')">
```
SebiWrn: I think we don't need to add the '/COMB' at the end of the URL and thus shouldn't because the URL looks cleaner without.

web/watch.go
```diff
@@ -82,6 +82,21 @@ func (r mainRoutes) WatchPage(c *gin.Context) {
 		}
 	}
 
+	if c.Param("version") == "" {
```
SebiWrn: We're already checking for version != "" so maybe we should combine these with an if/else so code is more readable

SebiWrn (changes requested):

web/watch.go
```diff
@@ -82,6 +82,21 @@ func (r mainRoutes) WatchPage(c *gin.Context) {
...
+		switch tumLiveContext.User.GetPreferredView() {
+		case "Presentation":
+
+			c.Redirect(http.StatusFound, c.Request.RequestURI+"/PRES")
+
```
karjo24: The unnecessary whitespace can be removed.

web/watch.go
```diff
@@ -82,6 +82,21 @@ func (r mainRoutes) WatchPage(c *gin.Context) {
...
+		case "Camera":
+			c.Redirect(http.StatusFound, c.Request.RequestURI+"/CAM")
+		case "Combined":
+			c.Redirect(http.StatusFound, c.Request.RequestURI+"/COMB")
```
karjo24: If we choose to leave out `/COMB` at https://github.com/TUM-Dev/gocast/pull/1498/files#r1971590406, we can leave it out here as well afaik.

web/template/partial/stream/actions.gohtml
carlobortolan: Fixed in https://github.com/TUM-Dev/gocast/pull/1498/commits/fe3d4509367b375070059d3f40caaf29aad80cae

web/watch.go
carlobortolan: Fixed in https://github.com/TUM-Dev/gocast/pull/1498/commits/fe3d4509367b375070059d3f40caaf29aad80cae

web/watch.go
carlobortolan: Fixed in https://github.com/TUM-Dev/gocast/pull/1498/commits/fe3d4509367b375070059d3f40caaf29aad80cae

web/watch.go
carlobortolan: Fixed in https://github.com/TUM-Dev/gocast/pull/1498/commits/fe3d4509367b375070059d3f40caaf29aad80cae

web/watch.go
```diff
@@ -58,7 +58,23 @@ func (r mainRoutes) WatchPage(c *gin.Context) {
...
+				c.Redirect(http.StatusFound, c.Request.RequestURI+"/CAM")
+				return
+			case "Split":
+				c.Redirect(http.StatusFound, c.Request.RequestURI+"/SPLIT")
+				return
+			case "Combined":
+				// do nothing, as version=="" implies combined
+			}
+		}
+	} else if c.Param("version") != "" && tumLiveContext.User != nil {
```
SebiWrn: We can remove `c.Param("version") != "" &&` as we are already in the `else`-block

SebiWrn:
```suggestion
	} else if tumLiveContext.User != nil {
```

SebiWrn (approved):
Just a small suggestion which you can but do not have to change.

Apart from that looks good to me.

web/watch.go
carlobortolan: Yes, thanks for pointing that out  :sweat_smile:

PR #1499, opened 2025-02-12
feat(runner/actions/mkvod): Create vod from livestream

joschahenningsen:
### Steps for Testing
- create a stream as in #1494 
- stop the stream via grpc
- observe stream being copied as vod to mass storage


runner/pkg/actions/mkvod.go
```diff
@@ -0,0 +1,142 @@
...
+	streamVersion, ok := d["streamVersion"].(string)
+	if !ok {
+		return AbortingError(fmt.Errorf("no stream end in context"))
+	}
+	recordingDir, ok := d["recordingDir"].(string)
+	if !ok {
+		return AbortingError(fmt.Errorf("no recordingDir in context"))
+	}
+
+	vodDir := path.Join(config.Config.StoragePath, fmt.Sprintf("%d", streamID), streamVersion)
```
CommanderStorm, line 38: Just to be sure (honestly, this code requires a bit too much context for a proper review):

streamVersion is not user-controlled, right? (unintentional path traversal)

runner/pkg/actions/mkvod.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return nil
+}
+
+// vodFromEventPlst modifies an HLS playlist from EVENT to VOD
+func vodFromEventPlst(playlist string) string {
+	var lines []string
+	hasEndlist := false
+
+	scanner := bufio.NewScanner(strings.NewReader(playlist))
```
CommanderStorm: Okay.. so to read a file line by line one has to read it as binary, convert to string and then scan the string?

Could you not skip the first part?

CommanderStorm (commented):
I don't think that my review is particularly relevant here as I just don't know enough about this part of the codebase

runner/pkg/actions/mkvod.go
joschahenningsen: No, it's a proto enum mapping to either of these values:
``` golang
0: "STREAM_VERSION_UNSPECIFIED",
1: "STREAM_VERSION_COMBINED",
2: "STREAM_VERSION_CAMERA",
3: "STREAM_VERSION_PRESENTATION",
```

runner/pkg/actions/mkvod.go
joschahenningsen: good remark, I changed vodFromEventPlst accept an io.Reader instead 👍

PR #1504, opened 2025-03-03
feat(runner/cleanup): start cleanup and vod sanity check action

joschahenningsen:
### Steps for Testing

Prerequisites:
- 1 Lecturer
- 2 Students
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. ...


runner/runner.go
```diff
@@ -158,28 +162,31 @@ func (r *Runner) RunAction(a []actions.Action, data map[string]any) string {
...
-				err := action(c, log, r.notifications, data, r.Metrics)
-				log.Info("action completed", "duration", time.Since(s).String())
+				if c.Err() != nil {
+					slog.Info("skipping action, parent context done.", "reason", c.Err().Error(), "job", job)
+					return
+				}
+				err := r.runSingularAction(c, action, data, log, job)
 				if err != nil {
 					log.Error("action error", "error", err) // use action specific logger
 					if actions.IsAbortingError(err) {
```
DawinYurtseven, line 185: I'm sorry if I am wrong but could this not lead to an infinite loop if an error is not an AbortingError? like when the ffmpeg command fails for other reasons?

runner/runner.go
```diff
@@ -158,28 +162,31 @@ func (r *Runner) RunAction(a []actions.Action, data map[string]any) string {
...
 						log.Info("action can't continue")
 						break // escape retry loop on unrecoverable error
 					}
 				} else {
-					break // escape retry loop on no error
+					return // Action done without error
```
DawinYurtseven, line 190: and here, if 'a' has multiple actions and the first action has no error, would it not just return the function without finishing the other actions? please let me know if I am wrong though

DawinYurtseven (commented):
everything that is new seems fine but the comments I'd like to at least know if it is intended for now or if they have been oversights (or if I am wrong)

DawinYurtseven:
after testing with grpcurl, it turns out that when the a job is created, the actions in the struct are not assigned to it, leaving to a access to nil error. I want to make sure if I am doing it right so:

 `grpcurl -plaintext -use-reflection -d '{"stream_id":1, "version": "STREAM_VERSION_COMBINED", "end": "2025-03-11T16:00:00Z", "ffmpeg_global_options": "-rtsp_transport tcp", "ffmpeg_output_options": "-c:a copy -c:v copy", "input": "rtsp://172.24.17.21/extron2"}' localhost:52735 protobuf.RunnerService/RequestStream`

my guess is that in the `RunAction` function the jobContext struct has no actions map initiated, leaving to a nil map.

PR #1506, opened 2025-03-03
Revise stats page

karjo24:
### Motivation and Context
The warning in the statistics section doesn't look great and isn't relevant any more. There hasn't been an active course with lectures before and after the mentioned date of 28th of June 2021 for years. Additionally, there is one template defined twice with the sole difference of whitespace and one headline. One of them is never used.

### Description
Go template to check the year of the course and display warning only when course is in a semester of 2022 or earlier. Delete the unused template.


SebiWrn (approved):

SebiWrn (changes requested):
We shouldn't remove it from the server statistics

kordianbruck (approved):

SebiWrn (approved):

PR #1510, opened 2025-03-08
840 short description for each video

### LINKED ISSUE #840: Short Decription for each video
I don't really know the topics of each video, so It takes a lot of time just to click on each of the video to see what they're about. Some lecturers also upload their lectures and exercises together so it is very confusing.

May be an additional button to view and edit a personal note on each video. This can also allow the lecturer to describe the overall topic of that video (e.g. Lecture 10: what is AI) and students can also add some reminders of their own such as "Need rewatch". We can reduce the memory required by limiting the number of allowed characters.

**Additional context**
image

karjo24:
### Motivation and Context
#840. Having the ability to make short notes or add/change the title of the lecture is useful for organising lectures when titles are not provided by the course admins.

### Description
There's an additional option in the dropdown menu for each stream on the course overview page that allows users to input their own (limited to 80 characters) lecture title. There's a new API endpoint, that takes the new title and stores it in a new table in the database.

### Steps for Testing
Prerequisites:
- 1 User
- 1 Course with 1 stream with title, 1 stream without

1. Log in and navigate to course
2. Click on the "More options" menu (three dots next to stream)
3. Click on "Edit personal lecture title"
4. Enter new title in popup menu
5. Verify the following behaviour
  5.1 Pressing "Enter" or clicking outside the input field changes the title
  5.2 Pressing "Escape" disregards any changes
  5.3 Deleting the value of the input field resets the title to the actual title provided by the lecturer
6. Reload to make sure the changes are persistent

### Screenshots

https://github.com/user-attachments/assets/d8adfd72-19e3-43b7-ab91-e7ec0bc55373


api/courses.go
```diff
@@ -344,7 +344,14 @@ func (r coursesRoutes) getCourseBySlug(c *gin.Context) {
...
 
-	course, err := r.CoursesDao.GetCourseBySlugYearAndTerm(c, uri.Slug, query.Term, query.Year)
+	user := tumLiveContext.User
+	var course model.Course
+	var err error
+	if user == nil {
+		course, err = r.CoursesDao.GetCourseBySlugYearAndTerm(c, uri.Slug, query.Term, query.Year)
+	} else {
+		course, err = r.CoursesDao.GetCourseBySlugYearTermUser(c, uri.Slug, query.Term, query.Year, user.ID)
+	}
```
carlobortolan: Maybe consider handling both cases (with and without user) in the GetCourseBySlugYearAndTerm method to avoid duplicate code in `GetCourseBySlugYearAndTerm` and `GetCourseBySlugYearTermUser` (see comment below)

dao/courses.go
```diff
@@ -231,6 +232,25 @@ func (d coursesDao) GetCourseBySlugYearAndTerm(ctx context.Context, slug string,
 	return course, err
 }
 
+// GetCourseBySlugYearTermUser loads courses with streams preloaded. Difference to GetCourseBySlugYearAndTerm: personal lecture titles of user are loaded as well
+func (d coursesDao) GetCourseBySlugYearTermUser(ctx context.Context, slug string, term string, year int, userID uint) (model.Course, error) {
```
carlobortolan: Would it be possible to simplify both `GetCourseBySlug...` methods into one or would it break something?

E.g., something like:
```go
func (d coursesDao) GetCourseBySlugYearAndTerm(slug string, term string, year int, userID uint) (model.Course, error) {
    var cacheKey string
    if userID == 0 {
        cacheKey = fmt.Sprintf("courseBySlugYearAndTerm%v%v%v", slug, term, year)
    } else {
        cacheKey = fmt.Sprintf("courseBySlugYearTermUser%v%v%v%v", slug, term, year, userID)
    }

    cachedCourses, found := Cache.Get(cacheKey)
    if found {
        return cachedCourses.(model.Course), nil
    }

    var course model.Course
    query := DB.Preload("Streams.VideoSections").Preload("Streams.Units", func(db *gorm.DB) *gorm.DB {
        return db.Order("unit_start desc")
    }).Preload("Streams", func(db *gorm.DB) *gorm.DB {
        return db.Order("start desc")
    }).Preload("Admins").Where("teaching_term = ? AND slug = ? AND year = ?", term, slug, year)

    if userID != 0 {
        query = query.Preload("Streams.CustomLectureTitles", "user_id = ?", userID)
    }

    err := query.First(&course).Error
    if err == nil {
        Cache.SetWithTTL(cacheKey, course, 1, time.Minute)
    }
    return course, err
}
```

web/template/home.gohtml
```diff
@@ -511,6 +517,12 @@
                                                                 <template x-if="vod.Dropdown.value">
                                                                     <div class="absolute tum-live-menu py-px bottom-full right-0 h-fit overflow-hidden z-50 w-56">
                                                                         {{if $user}}
+                                                                            <button type="button"
+                                                                                    @click="showLectureTitleInput=true;vod.Dropdown.value=false;"
+                                                                                    class="tum-live-menu-item">
+                                                                                <i class="fa-solid fa-pencil-square mr-4"></i>
+                                                                                <span class="block">Edit personal lecture title</span>
```
carlobortolan: Consider renaming this to "Personalize title" or "Edit personal title" to make the text shorter and similar to the other video actions.
image

carlobortolan (commented):
Tested it locally and it looks good to me :+1: 

Just two more general questions: 
- Should the personal lecture title be displayed only in the course-view, or also in the "Recent VODs" and videoplayer/stream-view?
- Should lecturers/admins of a course be allowed to add personal lecture titles, or would it be better to have the button set the actual lecture title to avoid confusion?

dao/courses.go
karjo24: Yes, that probably makes sense. As this only adds information to the return struct, I doubt that this breaks something (I did have a look at all usages and did a quick test locally). Technically we'd not need the check for userID != 0, because the where clause would not match any entry for == 0, but I'd like to keep it for clarity.

api/courses.go
karjo24: Fixed nearly everything, thanks for the review :)

karjo24:
>     * Should the personal lecture title be displayed only in the course-view, or also in the "Recent VODs" and videoplayer/stream-view?

Yes, thanks for catching that. Somehow thought the title was not displayed there and didn't check. I'll change that.
 
>     * Should lecturers/admins of a course be allowed to add personal lecture titles, or would it be better to have the button set the actual lecture title to avoid confusion?

It probably shouldn't be left like this as I agree with the confusion part. I'm not sure, whether we should hide the menu option for admins or make it change the title globally. After all, that's what the admin page is for. I lean towards the "hide it"-approach, but I'm open for other opinions.

carlobortolan:
> It probably shouldn't be left like this as I agree with the confusion part. I'm not sure, whether we should hide the menu option for admins or make it change the title globally. After all, that's what the admin page is for. I lean towards the "hide it"-approach, but I'm open for other opinions.

I think we can leave it out for now and add it in the future in a separate PR if necessary.

api/courses.go
```diff
@@ -344,7 +344,14 @@ func (r coursesRoutes) getCourseBySlug(c *gin.Context) {
 		Streams []model.StreamDTO
 	}
 
-	course, err := r.CoursesDao.GetCourseBySlugYearAndTerm(c, uri.Slug, query.Term, query.Year)
+	userId := uint(0)
```
joschahenningsen: nit:
```suggestion
	var userId uint
```

dao/user_defined_lecture_titles.go
```diff
@@ -0,0 +1,50 @@
+package dao
```
joschahenningsen, line 1: Nit: I know we do it in some places, but go discurages this style of filename. My suggestion would be `lecture_titles.go`

dao/user_defined_lecture_titles.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Delete a userDefinedLectureTitlesDao by id.
+func (d userDefinedLectureTitlesDao) Delete(userID uint, streamID uint) error {
+	return d.db.Delete(&model.UserDefinedLectureTitle{}, "user_id = ? AND stream_id = ?", userID, streamID).Error
+}
+
+// Upsert updates the entry if it exists, inserts it else
+func (d userDefinedLectureTitlesDao) Upsert(userLectureTitle *model.UserDefinedLectureTitle) error {
+	return d.db.Save(userLectureTitle).Error
+}
```
joschahenningsen: Do we need upserting? With user_id and stream_id being composite primary keys, we can use create with a conflict clause to update a record. See e.g. https://github.com/TUM-Dev/gocast/blob/fe0190d8b3d1c6e536bc4e5cb8f82cc2e5a534b2/dao/progress.go#L58
I would go that way and rename the method to `Save` and drop `Create` and `Upsert`

model/stream.go
```diff
@@ -398,9 +400,14 @@ func (s Stream) ToDTO() StreamDTO {
 	if s.Duration.Valid {
 		duration = s.Duration.Int32
 	}
+	customLectureName := ""
+	if len(s.CustomLectureTitles) > 0 {
+		customLectureName = s.CustomLectureTitles[0].Title
+	}
```
joschahenningsen: What exactly is happening here? We just use the first title? Wouldn't we send a random users title to someone?

model/user_defined_lecture_titles.go
```diff
@@ -0,0 +1,23 @@
...
+type UserDefinedLectureTitle struct {
+	UserID   uint   `gorm:"primaryKey" json:"userId"`
+	StreamID uint   `gorm:"primaryKey" json:"streamId"`
+	Title    string `gorm:"type:varchar(256)" json:"title"`
+}
+
+// BeforeCreate is a GORM hook that is called before a new user is created.
+// UserDefinedLectureTitle will not be saved if the title is too long
+func (u *UserDefinedLectureTitle) BeforeCreate(tx *gorm.DB) (err error) {
+	if len(u.Title) > 256 {
```
joschahenningsen, line 21: Can you trim the whitespace? https://pkg.go.dev/strings#TrimSpace

dao/courses.go
```diff
@@ -214,19 +214,34 @@ func (d coursesDao) GetInvitedUsersForCourse(course *model.Course) error {
...
 
-func (d coursesDao) GetCourseBySlugYearAndTerm(ctx context.Context, slug string, term string, year int) (model.Course, error) {
-	cachedCourses, found := Cache.Get(fmt.Sprintf("courseBySlugYearAndTerm%v%v%v", slug, term, year))
+// GetCourseBySlugYearAndTerm gets course by slug, year, term from database. Preloads personal lecture titles if userID is not 0
+func (d coursesDao) GetCourseBySlugYearAndTerm(ctx context.Context, slug string, term string, year int, userID uint) (model.Course, error) {
+	var cacheKey string
+	if userID == 0 {
+		cacheKey = fmt.Sprintf("courseBySlugYearAndTerm%v%v%v", slug, term, year)
+	} else {
+		cacheKey = fmt.Sprintf("courseBySlugYearTermUser%v%v%v%v", slug, term, year, userID)
```
joschahenningsen: Not going to be an issue soon if we don't have weird years, but this introduces a cache key collision if e.g. year=1, and userId=11 we produce the same key as year=11, userId=1. I suggest adding a `-` in between

joschahenningsen (commented):
Pretty nice! Left just a few comments

dao/user_defined_lecture_titles.go
karjo24: Done.

model/stream.go
karjo24: Although this isn't set to a random users title, this should definitely be changed, yes, Idk what I was thinking. It wasn't set to a random title, because the CustomLectureTitles in this PR only get preloaded in conjunction with a Where clause, that checks the userId.

model/user_defined_lecture_titles.go
karjo24: Is there any particular reason for this? Wouldn't this lead to a title being falsely accepted by `BeforeCreate` and then failing `Save`, if #whitespace + #meaningfulLetters > 256, but #meaningfulLetters < 256? Is this a best practice I'm not aware of?

model/stream.go
karjo24: It is now fixed. However, is there a cleaner way using Gorm to achieve the same to avoid having to add it manually afterwards?

SebiWrn:
Maybe there should also be a button for the user to delete the custom title, as just removing it to reset it to the lecturers title may not be intuitive.

karjo24:
> > * Should the personal lecture title be displayed only in the course-view, or also in the "Recent VODs" and videoplayer/stream-view?
> 
> 
> Yes, thanks for catching that. Somehow thought the title was not displayed there and didn't check. I'll change that.

I'd postpone this change as this would require changing the signature of two methods, which are heavily used throughout the project, which would inflate this PR.

alexanderstephan:
@copilot resolve the merge conflicts in this pull request.

alexanderstephan:
@karjo24 are you still working on this?

karjo24:
> @karjo24 are you still working on this?

No and I unfortunately don't have the time right now to do so :/ Forgot to unassign myself/delegate this.

joschahenningsen:
Sadly closing this as stale

PR #1519, opened 2025-03-14
Bump meilisearch-go

karjo24:
### Motivation and Context
This PR is in preparation for adding the new custom lecture titles (#1510) to Meilisearch. To be able to get a single list for lectures both based on lecturer-given and user-defined titles, we need the [federated search feature](https://www.meilisearch.com/docs/learn/multi_search/performing_federated_search) of Meilisearch, which the current version of meilisearch-go doesn't support.

### Steps for Testing
Prerequisites:
- 1 running instance of Meilisearch, e.g. using 

```
docker run -it --rm \
    -p 7700:7700 \
    getmeili/meilisearch:v1.13.0 \
```
Without any guarantees, that this doesn't break anything: Note that if you had Meilisearch on a lower version running on your device before, this may trigger an incompatible indices error. To avoid having to migrate the databases, just delete the working directory of meilisearch (here `meili_data`), start Meilisearch again and then simply run the Cron job `Export to Meili` from TUMLive's admin->maintenance panel.

Steps for Testing:
Make sure that the search function works just as before. This PR doesn't change our search requests, only updates the code to work with the newer version.


karjo24:
Closed as subset of #1462.

PR #1560, opened 2025-04-15
feat(admin/lecture-edit): Show warning when using self-stream feature

joschahenningsen:
grafik


CommanderStorm:
Offtopic comment, because of the screnshot:
Maybe consider adding a bit more padding. Currently, everything in this image looks quite cramped.

although: that applies to the most of the admin-ui, so maybe don't change that

joschahenningsen:
Agreed, admin ui needs a overhaul. That's out of scope here though

PR #1564, opened 2025-04-17
Enh/live admin view

### LINKED ISSUE #127: Live reactions for lectures
When watching a lecture that uses Tweedback there are things like the panic button, that clearly indicate that the lecture is hard to follow. This can be really useful, since the chat can become hard to track for large amount of students.
That also makes it difficult to actually estimate what percentage of the student shares a certain opinion about the lecture. Therefore there should be an option for easy, instant, grouped feedback, that can also be visualized while being completely anonymous.

There should be buttons that represent the mood of the student. For example there could be buttons that indicate the lecture moved too fast eg. 🏎️ , too slow e.g 🐌, or even mood indications. There could also be a certain threshold of votes e.g 20% of the current viewers within 5 minutes, that trigger an alert, that will be also directly sent to the lecturer.
The buttons should be combined with a heatmap or gauge, that will fade again after some time, as we expect the relevancy of the feedback to decrease over time. 

Of course you could always use Tweedback additionally, but it would be awesome if TUM-live became independent of other platforms and offers an integrated solution for lectures. Quizes can also be used for getting opinions about the lecture, but it is not ideal, as those have to be created manually.

SebiWrn:
### Motivation and Context
This pull request adds a new admin panel to use during livestreams. This panel allows presenters to see the live stream with some information on it. There are also buttons to start and stop the stream and a live chat view.

More features will be added in the future to this panel (e.g. reactions from students).

### Screenshots
image


web/template/admin/lecture-live-management.gohtml
```diff
@@ -0,0 +1,195 @@
...
+                                     x-text="(Math.round(data.bufferSeconds * 100) / 100) + 's'"></div>
+                                <div class="text-right" x-text="data.mediaRequests"></div>
+                                <div class="text-right" x-text="data.mediaRequestsFailed"></div>
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+
+            <!-- TODO: Add live view count and live view history -->
```
carlobortolan: I think it's better to create a new issue instead of putting it as a TODO

web/template/admin/lecture-live-management.gohtml
```diff
@@ -0,0 +1,195 @@
...
+                        px-6"
+                     @click.away="showModal = false">
+                    <!--Title-->
+                    <div class="flex justify-between pb-3">
+                        <p class="text-2 font-bold">Keep the recording after ending the stream?</p>
+                    </div>
+                    <!--Footer-->
+                    <div class="flex justify-start pt-2">
+                        <button class="bg-green-500 inline-block text-center w-24 text-white hover:bg-green-600 dark:hover:bg-green-600 font-bold rounded cursor-pointer p-1 mr-3"
+                                @click="fetch(`/api/stream/${streamID}/end?discard=${false}`).then(showModal=false)">
```
carlobortolan: Should we add a `.catch()` block here?

web/template/admin/lecture-live-management.gohtml
```diff
@@ -0,0 +1,195 @@
...
+                            Yes
+                        </button>
+                        <button class="bg-red-500 inline-block text-center w-24 text-white hover:bg-red-600 dark:hover:bg-red-600 font-bold rounded cursor-pointer p-1 mr-3"
+                                @click="fetch(`/api/stream/${streamID}/end?discard=${true}`).then(showModal=false)">
```
carlobortolan: Should we add a `.catch()` block here?

web/admin.go
```diff
@@ -209,6 +210,46 @@ func (r mainRoutes) LectureStatsPage(c *gin.Context) {
...
+	}
+
+	if !stream.LiveNow {
+		tools.RenderErrorPage(c, http.StatusNotFound, "Lecture is not live")
+		return
+	}
+
+	if c.Query("restart") == "1" {
+		c.Redirect(http.StatusFound, strings.Split(c.Request.RequestURI, "?")[0])
+		// TODO: Add restart functionality
```
carlobortolan: Is this meant to be a placeholder? 😅
(If it’s not implemented yet, I’d suggest leaving it out for now to avoid any confusion.)

carlobortolan (commented):
I tested it locally and it worked when accessing the URL directly (e.g., http://localhost:8081/admin/management/52/98), but I couldn’t find a way to navigate to the new admin view via the UI; am I missing something? 😅

SebiWrn:
> I tested it locally and it worked when accessing the URL directly (e.g., http://localhost:8081/admin/management/52/98), but I couldn’t find a way to navigate to the new admin view via the UI; am I missing something? 😅



Actually I forgot to add this functionality in this PR but it is added in #1565

web/admin.go
SebiWrn: I checked this and it seems like we do not have a restart functionality at all (also the restart button on the watch page does not do anything apart from reloading the webpage). I'll create a new issue to implement a real restart functionality that checks the stream and if needed restarts the stream on the runner

web/template/admin/lecture-live-management.gohtml
SebiWrn: Done

carlobortolan (approved):

PR #1565, opened 2025-04-17
Add Stream Reactions

### LINKED ISSUE #127: Live reactions for lectures

SebiWrn:
### Motivation and Context
This pull request adds previously mentioned reactions by students (look at #1564).
Closes #127.

The percentage below the reactions shows the usage in the last few minutes.

Available reactions can be modified in the config.yaml.
> [!IMPORTANT]
> Do not merge before #1564 !!

### Screenshots
image
image


SebiWrn:
Did not run golangci-lint yet as this would make changes to the runner code etc. which has nothing to do with this PR. Please ignore failing golangci-lint.

Fixing these issues should be done in a separate PR.

api/stream_reactions.go
```diff
@@ -0,0 +1,359 @@
...
+	"github.com/TUM-Dev/gocast/tools/realtime"
+	"github.com/getsentry/sentry-go"
+	"github.com/gin-gonic/gin"
+)
+
+type StreamReactionRoutes struct {
+	dao.DaoWrapper
+}
+
+// TODO: This can be modified to allow different reactions for different streams
```
kordianbruck, line 27: Probably want to just change to a comment. Its not really a TODO?

api/stream_reactions.go
```diff
@@ -0,0 +1,359 @@
...
+func (r StreamReactionRoutes) allowedReactions(c *gin.Context) {
+	c.JSON(http.StatusOK, tools.Cfg.AllowedReactions)
+}
+
+func (r StreamReactionRoutes) addReaction(c *gin.Context) {
+	cooldownSeconds := 10
```
kordianbruck: This variable is only used in a single location, please just put the value there instead of having this variable.

api/stream_reactions.go
```diff
@@ -0,0 +1,359 @@
...
+		return
+	}
+
+	if !user.IsEligibleToWatchCourse(course) {
+		_ = c.Error(tools.RequestError{
+			Status:        http.StatusForbidden,
+			CustomMessage: "user not eligible to watch course",
+		})
+		return
+	}
```
kordianbruck, line 70: optional: these three conditionals seem like they could be a separate function that could be reused?

api/stream_reactions.go
```diff
@@ -0,0 +1,359 @@
...
+		OnSubscribe:   reactionUpdateOnSubscribe,
+		OnUnsubscribe: reactionUpdateOnUnsubscribe,
+		OnMessage:     reactionUpdateSetStream,
+	})
+	daoWrapper = wrapper
+
+	go func() {
+		// Notify admins every 5 seconds
+		logger.Info("Starting periodic notification of reaction percentages")
+		for {
```
kordianbruck, line 233: Isn't this a leaking goroutine? You never end it or cancel it via a context?

config.yaml
```diff
@@ -104,3 +104,8 @@ meili:
 vodURLTemplate: https://stream.lrz.de/vod/_definst_/mp4:tum/RBG/%s.mp4/playlist.m3u8
 canonicalURL: https://tum.live
 rtmpProxyURL: https://proxy.example.com
+allowedReactions:
+  - 😊
+  - 👍
+  - 👎
+  - 😢
```
kordianbruck: Missing empty line at the end.

web/admin.go
```diff
@@ -209,6 +210,46 @@ func (r mainRoutes) LectureStatsPage(c *gin.Context) {
...
+	}
+
+	if !stream.LiveNow {
+		tools.RenderErrorPage(c, http.StatusNotFound, "Lecture is not live")
+		return
+	}
+
+	if c.Query("restart") == "1" {
+		c.Redirect(http.StatusFound, strings.Split(c.Request.RequestURI, "?")[0])
+		// TODO: Add restart functionality
```
kordianbruck: Still planned to be added in this PR?

kordianbruck (dismissed):

api/stream_reactions.go
SebiWrn: This should be ok as this is only started once at the initialize and it will be stopped when the program stops.

web/admin.go
SebiWrn: Already moved into new Issue

api/stream_reactions.go
SebiWrn: I will leave them there for now and they can be extracted if we need them somewhere else

alexanderstephan:
@copilot resolve the merge conflicts in this pull request

api/stream_reactions.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+func NotifyAdminsOnReactionPercentages(context context.Context) {
+	liveReactionListenerMutex.Lock()
+	defer liveReactionListenerMutex.Unlock()
+	streams := make([]uint, 0)
+	for _, session := range liveReactionListener {
+		streams = append(streams, session.stream)
+	}
+	liveReactionListenerMutex.Unlock()
```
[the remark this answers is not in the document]
Mazi02: done. First block uses lock and unlock manually and second one uses defer unlock.

api/stream_reactions.go
```diff
@@ -0,0 +1,377 @@
...
+
+	streamReactionPercentages := map[uint]map[string]float64{}
+
+	for _, stream := range streams {
+		reactionsRaw, err := daoWrapper.StreamReactionDao.GetByStreamWithinMinutes(context, stream, 2) // TODO: Make this variable for the lecturer
```
coolgirl19: Time can be increased to 15 mins as the lecturer might miss the reactions otherwise.
```suggestion
		reactionsRaw, err := daoWrapper.StreamReactionDao.GetByStreamWithinMinutes(context, stream, 15) // TODO: Make this variable for the lecturer
```

api/stream_reactions.go
```diff
@@ -0,0 +1,377 @@
...
+	// This can be modified to allow different reactions for different streams
+	if !slices.Contains(tools.Cfg.AllowedReactions, reaction.Reaction) {
+		_ = c.Error(tools.RequestError{
+			Status:        http.StatusBadRequest,
+			CustomMessage: "reaction not allowed",
+		})
+		return
+	}
+
+	lastReaction, _ := r.DaoWrapper.StreamReactionDao.GetLastReactionOfUser(c, user.ID)
```
coolgirl19, line 95: This should also take the stream id into account, because otherwise users following more than one live stream will not be able to react to multiple streams at once due to a 10-second global cooldown per user.

coolgirl19 (changes requested):
With the help of Copilot, I avoided the lock on the stream reactions object, which otherwise could have blocked other reactions from coming in. @Mazi02, please take a look at whether it makes sense; you can revert it otherwise. Is there any way for a user to know if their reaction has been rejected because of a cool-down period?

Mazi02 (approved):

api/stream_reactions.go
Mazi02: The question is if there is any user who follows two streams at once. I can't imagine somebody being able to be present at two different lecture livestreams and reacting to both. Thus having that feature seems to be redundant. What do you think?

Mazi02:
@coolgirl19 regarding the user notification I am going to make a notification pop up for it.

api/stream_reactions.go
coolgirl19: Yes, I thought so too. We can keep it as it is.

alexanderstephan:
@Mazi02 can this be reviewed again?

Mazi02:
@alexanderstephan yes I am done with the final touches, though I would like to have the @coolgirl19 to look over it too and if she agrees we can merge it.

coolgirl19 (approved):

PR #1574, opened 2025-05-09
Updated protobuf files to latest version and added runner version to DB

SebiWrn:


runner/commons.proto
```diff
@@ -1,4 +1,5 @@
-edition = "2023";
+syntax = "proto3";
```
kordianbruck: Why? The edition syntax is the new hot thing that we should probably also use?

SebiWrn: You are right

joschahenningsen (approved):

PR #1588, opened 2025-05-23
Fix host in APIv2 docs

carlobortolan:
### Overview
Update the base URL from `localhost:8081/api/v2` to `tum.live/api/v2`

### Steps for Testing
- Start the API and go to http://localhost:8081/api/v2/docs
- See the correct base URL

See also https://tum.live/api/v2/docs


apiv2/server/apiv2.proto
```diff
@@ -33,7 +33,7 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
     HTTPS
   ]
   base_path: "/api/v2"
-  host: "localhost:8081"
+  host: "tum.live"
```
joschahenningsen, line 36: swagger v3 allows specifying multiple hosts. Is that possible here too?

https://swagger.io/docs/specification/v3_0/api-host-and-base-path/

joschahenningsen: Might require updating to https://github.com/protoc-gen/protoc-gen-openapiv3

carlobortolan: I tried using [protoc-gen-openapiv3](https://github.com/protoc-gen/protoc-gen-openapiv3), but it didn’t seem to work. According to [grpc-gateway#441](https://github.com/grpc-ecosystem/grpc-gateway/issues/441) support for OpenAPI v3 is still in development in the "standard grpc ecosystem". I suggest we revisit this and update it once official support is available

joschahenningsen: Sounds good.

joschahenningsen (approved):

PR #1592, opened 2025-05-27
fix(web): blinking text on page (re)load

### LINKED ISSUE #1591: Placeholder text flashes when loading homepage
When loading the homepage, the placeholder text ("No streams, VODs or courses to show in this semester.") flashes for a fraction of a second before alpine is fully initialized. 

We should use [the cloak directive](https://alpinejs.dev/directives/cloak) to hide it.

RisGar:
### Motivation and Context

Fixes #1591

### Description

Adds a "loaded" variable to main.ts which is set to true after courses have successfully reloaded.

### Steps for Testing

Quickly reload the home page, the text should not blink any more as mentioned in  #1591

### DIFF
--- a/web/template/home.gohtml
+++ b/web/template/home.gohtml
@@ -814,7 +814,7 @@
                             </template>
                         </article>
                     </template>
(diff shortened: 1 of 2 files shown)


joschahenningsen (approved):
🫶

PR #1604, opened 2025-06-09
Added stream protocol to lecture halls

SebiWrn:
### Motivation and Context
This additional information is needed for the new runners as these are able to capture srt and rtmp streams. To specify this, lecture halls now have a field to store the streaming protocol.

> [!NOTE]
> This is currently not used but will be used in future PRs of the runner 

### Description
Added streaming protocol field to the lecture halls


carlobortolan (approved):
Looks good to me 👍

PR #1605, opened 2025-06-10
Add converting after stream (MKVod action)

SebiWrn:


runner/pkg/actions/mkvod.go
```diff
@@ -35,6 +35,13 @@ func MkVOD(_ context.Context, _ *slog.Logger, notify chan *protobuf.Notification
 		return AbortingError(fmt.Errorf("no recordingDir in context"))
 	}
 
+	logger = logger.With("stream_id", streamID)
```
joschahenningsen: Let's have `RunAction` accept a logger instead and construct one with useful metadata in the request handler: https://github.com/TUM-Dev/gocast/blob/49e3de093ae29ee28ed216a5f1f0353e0f118cd1/runner/handlers.go#L33

joschahenningsen (changes requested):
Just one suggestion, lgtm otherwise 🫶

joschahenningsen (approved):
Nice!

PR #1420, opened 2024-12-03
Delisting unavailable pinned courses

KemalKrKX:
### Motivation and Context
(https://github.com/TUM-Dev/gocast/issues/1351)

### Description
The code preloads the students when preloading the course, so that if a student is not enrolled at said course/doesn't have the view right the pinned lecture is not displayed anymore.

### Steps for Testing
Prerequisites:
- 1 Lecturer
- 1 Students
- 1 Lecture

1. Create a course
2. Enroll a student to said course
3. Pin the course as the student
4. Set the course as enrolled only and remove the student from database
5. Student should not see the course pinned anymore

PR #1614, opened 2025-07-10
fix: set pinned courses based on user eligibility

### LINKED ISSUE #1351: A pinned course that is no longer public cannot be unpinned.

carlobortolan:
This PR supersedes #1420.

Tried implementing it there, but due to large diffs in unrelated files, it was more convenient to open a new separate branch and PR.

Closes #1351.

### DIFF
--- a/web/index.go
+++ b/web/index.go
@@ -253,11 +253,7 @@ func (d *IndexData) LoadPinnedCourses() {
 	if d.TUMLiveContext.User != nil {
 		pinnedCourses = d.TUMLiveContext.User.PinnedCourses
 		for i := range pinnedCourses {
-			if d.TUMLiveContext.User.IsEligibleToSearchForCourse(pinnedCourses[i]) {
-				pinnedCourses[i].Pinned = false
-			} else {
-				pinnedCourses[i].Pinned = true
-			}
+			pinnedCourses[i].Pinned = d.TUMLiveContext.User.IsEligibleToSearchForCourse(pinnedCourses[i])
 		}
 		sortCourses(pinnedCourses)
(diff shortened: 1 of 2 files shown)


web/index.go
```diff
@@ -253,7 +253,11 @@ func (d *IndexData) LoadPinnedCourses() {
 	if d.TUMLiveContext.User != nil {
 		pinnedCourses = d.TUMLiveContext.User.PinnedCourses
 		for i := range pinnedCourses {
-			pinnedCourses[i].Pinned = true
+			if d.TUMLiveContext.User.IsEligibleToSearchForCourse(pinnedCourses[i]) {
+				pinnedCourses[i].Pinned = false
```
SebiWrn: Wouldn't this unpin all courses instead of unpinning the courses a use is not eligible to see

carlobortolan: Yes, you're right; I've fixed this in https://github.com/TUM-Dev/gocast/pull/1614/commits/e98fd0bf05def0bf2903156f0e938e7528bb7fd0. Thanks for noticing!

SebiWrn (approved):

PR #1622, opened 2025-07-24
fix(api): sanitation of course and lecture hall fields

joschahenningsen:
This fixes a security issue, where a malicious lecturer or admin, or attacker who gained their access
can create a course with a slug that is injected into commands, a worker runs.

Executing just `ffmpeg` in the worker has been considered, but the worker currently relies on the shells
capability to redirect its stdout to a file.

Thanks to @oiidmnk and @lcarilla for reporting the issue!

In similar fashion to https://github.com/TUM-Dev/gocast/commit/b995a237b2246e5809e8715e21e2ed0f59882081, we want to sanitize these values because they are used
to execute commands in the worker and can be abused by admins or lecturers to run arbitrary commands

### DIFF
--- a/model/course.go
+++ b/model/course.go
@@ -2,8 +2,10 @@ package model
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
+	"regexp"
 	"time"
 
 	"gorm.io/gorm"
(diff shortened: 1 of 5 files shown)


lcarilla:
thx for fixing, lgtm!

PR #1651, opened 2025-10-05
feat(runner): Implement checkvod action

### LINKED ISSUE #1646: [Runner]: Verify VoD health action
After the MKVod action, the runner should run an action that probes both the livestreamed hls stream and the created vod using `ffprobe` and conduct a sanity check (duration, number of streams, codec, etc.) to verify that the stream has been converted successfully. On failure it should log an error and fire a metric.

joschahenningsen:
### Motivation and Context

Closes #1646

The intention behind this action is to verify a vod created by the `MKVod` action, fire a metric if it is unhealthy and add the health status to the jobs data for descisions about the cleanup approach in a subsequent action (unimplemented).

### DIFF
--- a/pkg/runner_manager/manager.go
+++ b/pkg/runner_manager/manager.go
@@ -69,21 +69,21 @@ func (m *Manager) TriggerDueStreams() error {
 			errs = append(errs, fmt.Errorf("RequestStream COMB: %w", err))
 			continue
 		}
-		log.With("stream", s.ID, "job", resp.JobId, "version", model.COMB).Info("started Stream")
+		log.With("stream", s.ID, "job", resp.GetJobId(), "version", model.COMB).Info("started Stream")
 
 		resp, err = m.requestStreamVersion(ctx, s, client, lh, protobuf.StreamVersion_STREAM_VERSION_PRESENTATION)
 		if err != nil && !errors.Is(err, errNotNoLectureSource) {
 			errs = append(errs, fmt.Errorf("RequestStream PRES: %w", err))
 			continue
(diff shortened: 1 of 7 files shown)


runner/config/config.go
```diff
@@ -14,7 +14,7 @@ var Config struct {
 	SegmentPath  string `env:"SEGMENT_PATH" envDefault:"storage/live"`
 	GocastServer string `env:"GOCAST_SERVER" envDefault:"localhost:50056"`
 	Hostname     string `env:"REALHOST" envDefault:"localhost"`
-	EdgeServer   string `env:"EDGE_SERVER" envDefault:"http://localhost:50057"`
+	EdgeServer   string `env:"EDGE_SERVER" envDefault:"http://localhost:8089"`
```
joschahenningsen, line 17: This was the wrong default

pkg/runner_manager/manager.go
```diff
@@ -69,21 +69,21 @@ func (m *Manager) TriggerDueStreams() error {
 			errs = append(errs, fmt.Errorf("RequestStream COMB: %w", err))
 			continue
 		}
-		log.With("stream", s.ID, "job", resp.JobId, "version", model.COMB).Info("started Stream")
+		log.With("stream", s.ID, "job", resp.GetJobId(), "version", model.COMB).Info("started Stream")
```
joschahenningsen, line 72: Resolves to `string` instead of `*string` -> logs the content, not the memory address now

SebiWrn (approved):

PR #1656, opened 2025-10-15
Added SMP support to runners

SebiWrn:
### Motivation and Context
This PR adds SMP support to the runners (this is only active when in development mode)

PR #1657, opened 2025-10-15
Add selfstreaming feature to the runners

### LINKED ISSUE #1654: [Runner]: Self streams
To be filled out @joschahenningsen

SebiWrn:
### Motivation and Context
This PR adds the selfstreaming to the runners

> [!IMPORTANT]
> Please first merge #1656, as these PRs build on top of each other


SebiWrn:
Fixes #1654

pkg/runner_manager/manager.go
```diff
@@ -24,6 +26,10 @@ import (
 	"github.com/tum-dev/gocast/runner/protobuf"
 )
 
+var logger = log.New(log.NewJSONHandler(os.Stdout, &log.HandlerOptions{
+	Level: log.LevelDebug,
+})).With("service", "runner_manager")
```
joschahenningsen: Ideally, logger should not be package global, but a member of `*Manager` and initialized in `runner_manager.New`[1]. Also, please apply it to all logging in this file if you introduce a new logging approach, e.g. here https://github.com/TUM-Dev/gocast/blob/515502a3c9543ba42195b52537f90fb3bf3e54c3/pkg/runner_manager/manager.go#L61

[1]: https://github.com/TUM-Dev/gocast/blob/515502a3c9543ba42195b52537f90fb3bf3e54c3/pkg/runner_manager/manager.go#L42

runner/serverhandler.go
```diff
@@ -39,12 +39,12 @@ func (r *Runner) RegisterWithGocast(retries int) {
 	}
 }
 
-// dialIn connects to manager instance and returns a client
-func (r *Runner) dialIn() (protobuf.RunnerManagerServiceClient, error) {
+// DialIn connects to manager instance and returns a client
+func DialIn() (protobuf.RunnerManagerServiceClient, error) {
```
joschahenningsen: Why is this change neccessary? I don't like that this method becomes part of the API surface of the `runner` package. Is it to be able to use the method in onPublish [1]?
If so, I'd suggest making onPublish a member function of Runner to avoid having to export this.

[1]: https://github.com/TUM-Dev/gocast/blob/a6a86eb11465b5383ecd4122c054781397452c7e/runner/rest.go#L53

pkg/runner_manager/manager.go
```diff
@@ -240,15 +252,85 @@ func (m *Manager) requestStreamVersion(ctx context.Context, s model.Stream, clie
...
 	})
 }
 
+func (m *Manager) RequestSelfStream(ctx context.Context, request *protobuf.SelfStreamRequest) (*protobuf.SelfStreamResponse, error) {
+	if _, err := m.dao.RunnerDao.Get(ctx, *request.Hostname); err != nil {
+		logger.Warn("Could not find runner for selfstream", "hostname", *request.Hostname)
+		return nil, err
+	}
+
+	if *request.StreamKey == "" {
```
joschahenningsen: This should work, right? Using the protoc generated getters is usually safer than dereferencing a potentially nil string.

```suggestion
	if request.GetStreamKey() == "" {
```

runner/rest.go
```diff
@@ -0,0 +1,121 @@
...
+
+type SafeStreamStruct struct {
+	mutex   sync.Mutex
+	streams map[string]SafeStreamStructStream
+}
+
+var SafeStreams = SafeStreamStruct{
+	mutex:   sync.Mutex{},
+	streams: make(map[string]SafeStreamStructStream),
+}
```
joschahenningsen: This is mostly from the worker right? Tbh I don't quite recall why this was done and maybe we can think about simplifying it or unifying it with the runners state handling of streams that are in progress?

runner/runner.go
```diff
@@ -95,6 +95,7 @@ func (r *Runner) Run() {
 	go r.Metrics.Run()
 	go r.handleNotifications()
 	go r.InitApiGrpc()
+	go r.InitApi(":8060")
```
joschahenningsen: Did you test this? It seems like `InitApi` is blocking, isn't it?

runner/rest.go
```diff
@@ -0,0 +1,121 @@
...
+	Query    string `json:"query"`
+}
+
+// InitApi creates routes for the api consumed by mediamtx and TUM-Live
+func (r *Runner) InitApi(addr string) {
+	//http.HandleFunc("/", defaultHandler)
+	http.HandleFunc("/on_publish", onPublish)
+	// this endpoint should **not** be exposed to the public!
+	//http.HandleFunc("/upload", handleUpload)
+	log.Fatal(http.ListenAndServe(addr, nil))
```
joschahenningsen:
```suggestion
	http.HandleFunc("/on_publish", onPublish)
	// this endpoint should **not** be exposed to the public!
	log.Fatal(http.ListenAndServe(addr, nil))
```

runner/rest.go
```diff
@@ -0,0 +1,121 @@
...
+			Hostname:   &config.Config.Hostname,
+			StreamKey:  &streamKey,
+			CourseSlug: &slug,
+		})
+		if err != nil {
+			log.Error(err)
+			http.Error(w, "Authentication failed for SendSelfStreamRequest", http.StatusForbidden)
+			return
+		}
+	}()
```
joschahenningsen: I see no reason for this to be in a goroutine

joschahenningsen (changes requested):
Sorry for being so verbose reviewing some of the old worker code, I think it's worthwile to re-think how it was done back then now and get rid of some of that tech depth :) 

Overall, I really love the approach of having gocast be the initiator of self-streams as though they were normal lecture-hall streams!

runner/runner.go
joschahenningsen: Just saw, that it was in a goroutine, sorry :D

runner/rest.go
SebiWrn: The Idea of this is from the worker, but the code is not copy paste. This is needed so we don't start multiple ffmpeg instances for the same stream.
Currently the runner does not have a real state for managing streams that are in progress, as this is done through jobs. I don't think it makes sense to add a Job for handling the self-stream request.

However, maybe we should implement such a state management for the runner in the future, but I would like to postpone this to another PR.

pkg/runner_manager/manager.go
```diff
@@ -285,6 +293,47 @@ func (m *Manager) requestStreamVersion(ctx context.Context, s model.Stream, clie
...
+		return nil, errors.New("stream rejected")
+	}
+
+	client, err := m.getClient(ctx)
+	if err != nil {
+		m.logger.Error("Could not get client", "err", err)
+		return nil, err
+	}
+
+	resp, err := m.requestStreamVersion(ctx, stream, client, model.LectureHall{}, protobuf.StreamVersion_STREAM_VERSION_COMBINED)
```
joschahenningsen, line 328: We need to make sure, this is requested at the same runner, that this request originates from, otherwise it will break as soon as we have more than one runner

joschahenningsen (changes requested):
One concern with this, otherwise looks great! 💯

pkg/runner_manager/manager.go
SebiWrn: You're right, I forgot to implement this. An alternative would be to have only one mediaMTX instance at a time and then distribute the load of selfstreams evenly to the runners as every runner could fetch the stream from this instance and we wouldn't rely on only one machine for a stream (except for the mediaMTX). What do you think about this?

joschahenningsen: I'd like all runners to be treated the same, let's design this to work without magic knowledge of what which runner does

joschahenningsen: If we need load balancing to multiple ingest servers, let's use traefik and DNS rr to distribute streams to multiple runners

SebiWrn: This means we only have one mediamtx for all instead of having one for each runner?

SebiWrn:
Whole functionality of requesting a selfstream now moved to main instance (gocast), so that every runner can handle the selfstreaming requests.
Ingest is now own package that could be deployed separately

pkg/runner_manager/manager.go
```diff
@@ -50,16 +57,16 @@ func New(dao dao.DaoWrapper, opts ...Option) *Manager {
 type Option func(m *Manager)
 
 func (m *Manager) TriggerDueStreams() error {
-	log.Info("Triggering due streams")
+	m.logger.Info("Triggering due streams")
 	ctx := context.Background()
 	streams, err := m.dao.GetDueStreamsForRunners()
 	// TODO: Remove this when turning off workers
-	if web.VersionTag == "development" {
+	if false { // TODO: For testing you have to change this
```
joschahenningsen, line 64: Why not use the versionTag?

pkg/runner_manager/manager.go
```diff
@@ -304,6 +312,29 @@ func (m *Manager) requestStreamVersion(ctx context.Context, s model.Stream, clie
 	})
 }
 
+
+func (m *Manager) RequestSelfStream(ctx context.Context, stream model.Stream) error {
```
joschahenningsen, line 315: This method is missing a godoc comment

runner/runner.proto
```diff
@@ -44,6 +44,7 @@ service RunnerManagerService {
   rpc Notify (Notification) returns (NotificationResponse) {}
 }
 
+
```
joschahenningsen:
```suggestion
```

tools/config.go
```diff
@@ -172,10 +172,11 @@ type Config struct {
...
 	} `yaml:"meili"`
-	VodURLTemplate string `yaml:"vodURLTemplate"`
-	CanonicalURL   string `yaml:"canonicalURL"`
-	WikiURL        string `yaml:"wikiURL"`
-	RtmpProxyURL   string `yaml:"rtmpProxyURL"`
+	VodURLTemplate   string `yaml:"vodURLTemplate"`
+	CanonicalURL     string `yaml:"canonicalURL"`
+	WikiURL          string `yaml:"wikiURL"`
+	RtmpProxyURL     string `yaml:"rtmpProxyURL"`
+	RtmpProxyService string `yaml:"rtmpProxyService"`
```
joschahenningsen, line 179: I don't see where this is used, is it inteded to superseed RtmpProxyURL at some point?

pkg/runner_manager/manager.go
```diff
@@ -304,6 +312,29 @@ func (m *Manager) requestStreamVersion(ctx context.Context, s model.Stream, clie
 	})
 }
 
+
```
joschahenningsen:
```suggestion
```

joschahenningsen (commented):
Couple nits, otherwise great 🚀

runner/runner.proto
joschahenningsen: I think this is what the linter dislikes

pkg/runner_manager/manager.go
SebiWrn: Because this would result in a cyclic dependency

tools/config.go
SebiWrn: You're right, I thought I need do add this, but then I apparently found out that it's the same as ingestbase

joschahenningsen: Got it, let's remove it then

joschahenningsen (approved):

PR #1660, opened 2025-10-17
feat(camera): add sony srg support

carlobortolan:
### Motivation and Context
Add initial support for Sony SRG-A40 cameras.

### Description
- Added SonySRG methods (SetPreset, TakeSnapshot, GetPresets)
- Updated model/lecture_hall.go and api/worker_grpc.go to support new camera type


tools/camera/sonysrg.go
```diff
@@ -0,0 +1,68 @@
...
+
+func NewSonySRG(addr string, auth string) SonySRG {
+	return SonySRG{addr: addr, auth: auth}
+}
+
+func (s SonySRG) SetPreset(presetId int) error {
+	auth := &s.auth
+	if s.auth == "" {
+		auth = nil
+	}
```
joschahenningsen: Probably it's easier to have `SonySRG.auth` be a pointer and have `tools/presets.go` dereference the config value.

tools/camera/sonysrg.go
```diff
@@ -0,0 +1,68 @@
...
+type SonySRG struct {
+	addr string
+	auth string
+}
+
+func NewSonySRG(addr string, auth string) SonySRG {
+	return SonySRG{addr: addr, auth: auth}
+}
+
+func (s SonySRG) SetPreset(presetId int) error {
```
joschahenningsen: The other cameras have pointer receivers, right? Can we do it the same for the sony?

tools/camera/sonysrg.go
```diff
@@ -0,0 +1,68 @@
...
+	err = saveResponseBuffer(outDir, filename, resp)
+	if err != nil {
+		return "", err
+	}
+	return filename, nil
+}
+
+func (s SonySRG) GetPresets() ([]model.CameraPreset, error) {
+	// Sony SRG-A40 cameras support up to 256 presets, but only a few are used (see panasonic.go)
+	presets := make([]model.CameraPreset, 16)
```
joschahenningsen: This looks very static, is there no way to fetch how many presets are actually configured on the camera like we do with the axis?

tools/camera/sonysrg.go
carlobortolan: since the initial setup you pushed had it as a non-pointer, I assumed there might have been a specific reason for that; I'll update this later :+1:

tools/camera/sonysrg.go
carlobortolan: ~~yes, this should probably be possible; I'll look into it later today~~

tools/camera/sonysrg.go
joschahenningsen: Yup, that was my oversight 👍

carlobortolan: Updated it in https://github.com/TUM-Dev/gocast/pull/1660/commits/a75c3b0b614752cabf19f56c4efd41351823b887

tools/camera/sonysrg.go
carlobortolan: Should be fixed now - although I found that the other cam implementations seem to be sometimes inconsistent (e.g., Sony and Axis both return the `Cam` interface, while Panasonic doesn't; Axis has some methods as pointer receivers and others as value receivers, while Panasonic has them all as value receivers, ...) Is this intended/am I missing something? If not, I'd look into cleaning them up as well, if that's ok.

tools/camera/sonysrg.go
carlobortolan: I looked into it and I think that the Sony API is more limited than Axis cameras (or at least based on the [docs](https://www.sony.com/electronics/support/res/manuals/5053/f48354480e724f0aef3d45d2e1eeab63/50531551M.pdf) I couldn't find a single endpoint to list all configured presets), so we can either:
- Query each preset position individually (0-255) to check if it's set
- Hardcode preset ranges (e.g., 0-15, as currently implemented)

joschahenningsen:
Need any help here? @carlobortolan

carlobortolan:
> Need any help here? @carlobortolan

@joschahenningsen The requested changes should be implemented (see the comments above). The reason this is still marked as a draft is that I've had difficulties testing this with actual Sony cameras - I currently can't access them from my machine and haven't yet figured out a workaround; if you're able to test it on your side, that would be a big help.

joschahenningsen (approved):

PR #1669, opened 2025-10-19
fix: combined view redirect

Coronon:
### Motivation and Context

Some VODs don't have a pure presentation stream. When switching to the combined view (`COMB`) however, the user is instantly redirected to the presentation view again -> without manually changing the URL the user can't access the combined view.

https://github.com/user-attachments/assets/e7fc245d-7b24-42de-ae7d-5f3782f223c5

### Description
Add the `/COMB` path explicitly to prevent redirect back to faulty stream

### Steps for Testing
Prerequisites:
- 1 VOD without presentation such as: https://tum.live/w/WiSe25_26_CE_NetSec/66507/PRES

1. Log in
2. Open such a VOD
3. Try to change stream to "Combined view"

### Screenshots
No UI change - see video above


joschahenningsen (approved):
LGTM, thanks a lot for your contribution!

PR #1670, opened 2025-10-19
Add livestream cleanup to runners

### LINKED ISSUE #1647: [Runner]: Cleanup livestreams
The local hls livestreams need to be cleaned up after some time.
Proposed solution:

- The verify VoD health action should write the state of the vod to disk. This could be done with a local sqlite database that lives along the livestreaming files in a table like this: 
  - `type StreamConversionState struct { time time.Time, healthy bool, string livestreamFiles }`
- A cron-job like function runs every few minutes and looks for streams that have completed more than two hours ago.
  - If they were transcoded successfully, the livestream files are deleted.
  - If not, they are moved somewhere in the mass storage for manual recovery.

SebiWrn:
#### Cleanup for runner

Adds a periodic task to the runner cleaning up all livestreams from the past


runner/livestream_cleanup.go
```diff
@@ -0,0 +1,97 @@
...
+			if !deletionError && shouldDelete {
+				r.log.Info("Deleting whole livestream dir as VoD healthy", "name", streamDir.Name())
+				err = os.RemoveAll(filepath.Join(livePath, streamDir.Name()))
+				if err != nil {
+					r.log.Error("Error removing livestream cleanup dir", "name", streamDir.Name())
+				}
+			}
+		}
+		// TODO: Set this higher
+		time.Sleep(1 * time.Minute)
```
joschahenningsen: This will leave a stray goroutine when the runner tries to exit. The method should receive a context that is canceled when the application should terminate. Then instead of sleeping, use time.After with select:

```go
select {
case <-ctx.Done():
	return nil, ctx.Err()
case <-time.After(time.Minute):
}
```

runner/livestream_cleanup.go
```diff
@@ -0,0 +1,97 @@
...
+				}
+				for _, streamVersionDir := range streamVersionDirs {
+					if streamVersionDir.IsDir() {
+						files, err := os.ReadDir(filepath.Join(livePath, streamDir.Name(), streamVersionDir.Name()))
+						if err != nil {
+							r.log.Error("Error reading livestream cleanup dir", "name", streamDir.Name())
+							deletionError = true
+							continue
+						}
+						for _, file := range files {
```
joschahenningsen: This is really deeply nested, please refactor into multiple functions

joschahenningsen (commented):
Added some comments, also please look into the copilot comments, I think they are valid

runner/livestream_cleanup.go
joschahenningsen: Something like this should be useful in the future, feel free to cherry-pick https://github.com/TUM-Dev/gocast/commit/5f5158dac5fbf51a88c8c973183ec9ba25055b0a

joschahenningsen:
@SebiWrn Check out 70380fcc1773ffe56aaf2dca01a9493d7df3247a, there I refactored this a bit to use [filepath.WalkDir](https://pkg.go.dev/path/filepath#WalkDir) to make it a bit cleaner.

joschahenningsen (approved):

PR #1680, opened 2025-11-03
Feat/runner thumbnails

joschahenningsen:


runner/pkg/actions/error.go
```diff
@@ -34,6 +38,6 @@ func (e *abortingError) Error() string {
 }
 
 func IsAbortingError(err error) bool {
-	_, ok := err.(isAbortingError)
+	ok := errors.As(err, new(isAbortingError))
```
joschahenningsen, line 41: There was a bug here, this didn't work previously with wrapped errors.

pkg/runner_manager/manager.go
```diff
@@ -285,6 +304,59 @@ func (m *Manager) requestStreamVersion(ctx context.Context, s model.Stream, clie
...
+	fname := fmt.Sprintf("%d_%s.jpeg", stream.ID, req.GetStreamVersion().String())
+	file := model.File{
+		StreamID: stream.ID,
+		// /mass/thumbs/2025/10/500/1024_STREAM_VERSION_COMBINED.jpeg
+		Path:     filepath.Join(m.massStorage, "thumbs", stream.Start.Format("2006/01"), fmt.Sprintf("%d", stream.CourseID), fname),
+		Filename: fname,
+		Type:     fileType,
+	}
+
+	err = os.MkdirAll(file.Path, 0o755)
```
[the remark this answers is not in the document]
joschahenningsen: Not true, file is a domain model that we define, it's not the go stdlib file. Path does not include the filename.

runner/pkg/actions/checkvod.go
```diff
@@ -27,6 +27,8 @@ func CheckVoD(ctx context.Context, logger *slog.Logger, _ chan *protobuf.Notific
 		// Add a keep flag to the folder, add duration of 2h on average to the keep
 		formatString = ".keep-%d"
 		duration = 2 * time.Hour
+	} else {
+		d["duration"] = duration
```
SebiWrn: Why do we need this in the data and why do we only add it if there's no error? Shouldn't we still have the default value here?

pkg/runner_manager/manager.go
```diff
@@ -285,6 +304,59 @@ func (m *Manager) requestStreamVersion(ctx context.Context, s model.Stream, clie
...
+	}()
+
+	_, err = f.Write(req.GetThumbnail())
+	if err != nil {
+		return status.Errorf(codes.Internal, "can't write file: %v", err)
+	}
+
+	err = m.dao.FileDao.NewFile(&file)
+	if err != nil {
+		return status.Errorf(codes.Internal, "can't save thumbnail: %v", err)
```
SebiWrn: We should write something like "can't add file to db", this makes Debugging easier as it does not look like an filesystem error

runner/pkg/actions/mkthumb.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	thumb, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		_ = thumb.Close()
+		_ = os.Remove(path)
+	}()
+	return io.ReadAll(thumb)
```
SebiWrn, line 72:
```suggestion
	return io.ReadAll(thumb), nil
```

joschahenningsen: [ReadAll](https://pkg.go.dev/io#ReadAll) returns `[]byte, error`

runner/pkg/actions/checkvod.go
joschahenningsen: Right, that was something I used before to grab the thumbnail from the middle of the video, which I dropped because the library I use also does this. Removed.

SebiWrn (approved):
Looks good to me

PR #1694, opened 2025-11-22
Enh/update videojs to version 8

SebiWrn:
### Motivation and Context
Videojs moved to version 8 so we should also move to version 8

### Description
Migrate everything to version 8

### DIFF
--- a/web/ts/TUMLiveVjs.ts
+++ b/web/ts/TUMLiveVjs.ts
@@ -1,6 +1,6 @@
 import { getQueryParam, keepQuery, postData, Time } from "./global";
 import { StatusCodes } from "http-status-codes";
-import videojs, { VideoJsPlayer } from "video.js";
+import videojs from "video.js";
 import airplay from "@silvermine/videojs-airplay";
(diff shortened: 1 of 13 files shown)


web/ts/TUMLiveVjs.ts
```diff
@@ -205,8 +206,8 @@ export const initPlayer = function (
     });
 
     // When catching up to live, resume at normal speed
-    player.liveTracker.on("liveedgechange", function (evt) {
-        if (player.liveTracker.atLiveEdge() && player.playbackRate() > 1) {
+    (player as any).liveTracker.on("liveedgechange", function (evt) {
```
joschahenningsen, line 209: Why don't we need `// eslint-disable-next-line @typescript-eslint/no-explicit-any` here?

joschahenningsen (approved):
Super nice! 🙌

web/ts/TUMLiveVjs.ts
SebiWrn: I think this is the case because we do not store it as any but just use it like that

PR #1695, opened 2025-11-22
feat(subtitles): Autogenerate subtitles when VOD is ready

joschahenningsen:
When streams are ready, by both workers and runners, a subtitle is automatically requested.

### DIFF
--- a/api/worker_grpc.go
+++ b/api/worker_grpc.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/reflection"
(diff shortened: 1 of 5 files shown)


api/worker_grpc.go
```diff
@@ -423,20 +428,67 @@ func (s server) NotifyUploadFinished(ctx context.Context, req *pb.UploadFinished
...
+
+	err = s.requestSubtitles(ctx, stream, modelVersion)
+	if err != nil {
+		logger.Error("failed to request subtitles", "stream", stream.ID, "version", modelVersion, "err", err)
+	}
+
 	return &pb.Status{Ok: true}, nil
 }
 
+func (m *server) requestSubtitles(ctx context.Context, stream model.Stream, version model.StreamVersion) error {
```
joschahenningsen: This method is duplicated, but I consider the worker_grpc.go code legacy and want to delete it soon anyway.

SebiWrn (approved):

PR #1717, opened 2025-12-09
feat: add artemis integration

Anishyou:
### Motivation and Context
**Problem:** Artemis (TUM's interactive learning platform) needs to integrate with TUM-Live to allow instructors to upload exercise/tutorial videos directly from Artemis to course-specific TUM-Live streams. Currently, there's no way for external systems to authenticate users and upload videos programmatically.

**Solution:** This PR implements a secure SSO-based authentication flow and video upload API that allows Artemis to:
1. Authenticate TUM users who have already logged in via SAML SSO
2. Retrieve available courses where the user has upload permissions
3. Upload videos to specific courses with proper authorization checks

This integration enables seamless video content creation within Artemis that automatically publishes to TUM-Live for student viewing.

### Description

This PR adds a complete Artemis integration with two main components:

#### 1. SSO Authentication Endpoint (`api/artemis_auth_sso.go`)
- **Endpoint:** `POST /api/artemis/auth/sso`
- Accepts SAML user attributes (LRZ ID, name, email) from Artemis after TUM SSO authentication
- Validates a shared API key (`artemisApiKey`) to ensure requests come from authorized Artemis instances
- **Security:** Only authenticates pre-existing users (no auto-registration) who have lecturer/admin roles
- Returns a long-lived upload token (1 year validity) and list of VOD-enabled courses
- Implements token reuse logic to avoid creating duplicate tokens

#### 2. Video Upload Endpoint (`api/artemis_upload.go`)
- **Endpoint:** `POST /api/course/artemis/:courseID/upload`
- Token-based authentication via Bearer token or query parameter
- Supports creating new streams or uploading to existing streams
- Validates course permissions (user must be course admin, VOD must be enabled)
- Proxies multipart file uploads to available workers with load balancing
- Returns video URLs in response headers for easy integration:
  - `X-Video-URL`: Direct watch URL
  - `X-Embed-URL`: Embeddable iframe URL
  - `X-Stream-ID`: Stream identifier

#### 3. Configuration & Router Updates
- Added `artemisApiKey` configuration parameter in `tools/config.go`
- Registered new routes in `api/router.go`
- Added `courseInfo` type for structured course information responses

Prerequisites:
- 1 Lecturer
- 2 Students
- 1 Livestream

1. Log in
2. Navigate to a Livestream
3. ...

### DIFF
--- a/api/artemis_auth_sso.go
+++ b/api/artemis_auth_sso.go
@@ -32,12 +32,12 @@ type artemisSSORoutes struct {
 
 type artemisSSORequest struct {
 	// User information from SAML assertion
-	LrzID       string `json:"lrzId" binding:"required"`       // TUM LRZ ID (e.g., "ge12abc")
-	MatrNr      string `json:"matrNr"`                         // Matriculation number (optional for staff)
-	FirstName   string `json:"firstName" binding:"required"`   // Given name from SAML
-	LastName    string `json:"lastName"`                       // Surname from SAML (optional)
-	Email       string `json:"email"`                          // Email (optional, will be constructed if missing)
-	
+	LrzID     string `json:"lrzId" binding:"required"`     // TUM LRZ ID (e.g., "ge12abc")
+	MatrNr    string `json:"matrNr"`                       // Matriculation number (optional for staff)
+	FirstName string `json:"firstName" binding:"required"` // Given name from SAML
+	LastName  string `json:"lastName"`                     // Surname from SAML (optional)
+	Email     string `json:"email"`                        // Email (optional, will be constructed if missing)
+
 	// Artemis authentication
 	ArtemisAPIKey string `json:"artemisApiKey" binding:"required"` // Shared secret between Artemis and TUM Live
 }
@@ -265,4 +265,3 @@ func (r artemisSSORoutes) getOrCreateUploadTokenSSO(c *gin.Context, userID uint)
 	logger.Info("Created new token", "userID", userID, "tokenID", newToken.ID, "expires", expires.Time)
 	return &newToken, nil
 }
-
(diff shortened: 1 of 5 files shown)


joschahenningsen:
Let me intervene early here. We have a token concept, it's not used widely but has the potential to allow integrating third parties like Artemis. The current approach of this PR does not fit the project well imo.
Have a look at this: https://github.com/TUM-Dev/gocast/blob/dev/api/token.go

The way I think this could be done is:
- Allow token scopes to be per course
- Add the possibility for lecturers to create a course-scoped token in the course settings
- Adjust the auth middleware to accept that token as a mean of authentication for the existing endpoints. 

If the endpoints don't do what you need, please open an issue or discussion to walk through design of a potential new one. Those should also probably be defined as gRPCs (which can be accessed via REST through our gateway) in `api_v2` .

Anishyou:
> Let me intervene early here. We have a token concept, it's not used widely but has the potential to allow integrating third parties like Artemis. The current approach of this PR does not fit the project well imo. Have a look at this: https://github.com/TUM-Dev/gocast/blob/dev/api/token.go
> 
> The way I think this could be done is:
> 
> * Allow token scopes to be per course
> * Add the possibility for lecturers to create a course-scoped token in the course settings
> * Adjust the auth middleware to accept that token as a mean of authentication for the existing endpoints.
> 
> If the endpoints don't do what you need, please open an issue or discussion to walk through design of a potential new one. Those should also probably be defined as gRPCs (which can be accessed via REST through our gateway) in `api_v2` .


Thanks for the feedback! I understand the current approach doesn't fit well. I will work on as you described and update my PR. Thanks a lot :)

kmk142789 (dismissed):

joschahenningsen:
Closing as stale

Issue #1701, opened 2025-11-24
Add config option to disable gin request logs


joschahenningsen:
In production, the gin request logs for non ok (>=400) status codes are very noisy. We need a config option to disable them.

e.g.
```
{"service": "GIN", "time": 2025-11-24 16:57:47, "status": 403, "client": "127.0.0.1", "path": "/api/progressReport", "agent": Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36}
```

PR #1730, opened 2026-01-22
disable gin request logs for non ok (>=400) status codes in production

vannsoko:
### Motivation and Context

- disable gin request logs for non ok (>=400) status codes in production
- issue: https://github.com/TUM-Dev/gocast/issues/1701

### Description

- Disable the logs from gin when VersionTag != "development"


joschahenningsen (approved):
Thanks!

PR #1791, opened 2026-04-11
migrating from git to https

Mazi02:
### Motivation and Context
Migrating from git to https in pre-commit-config


kordianbruck:
But git should also work, what was the error message you saw?

Mazi02:
Problem is solved. Don't need https.

PR #1798, opened 2026-04-13
chore(git): update .gitignore

Mazi02:
### Motivation and Context
The go.imports.xml is IDE-specific configuration and should be in .gitignore.

### Description
Adding go.imports.xml to the gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 # Binaries for programs and plugins
 main
 *.exe
-*.exe~
+ .exe~
 *.dll
 *.so
 *.dylib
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,7 +2,7 @@
 # Binaries for programs and plugins
 main
 *.exe
-*.exe~
```
joschahenningsen, line 5: This we should keep

kordianbruck (approved):

.gitignore
joschahenningsen: Ah I missed that it's duplicated!

joschahenningsen (approved):

.gitignore
joschahenningsen: Ah I missed that it's duplicated!

Mazi02: No you are right this one is for backup files, I fixed it.

PR #1803, opened 2026-04-18
fix(bookmarks): prevent event propagation for bookmark update and del…

coolgirl19:
For description and reproduction steps, please see issue https://github.com/TUM-Dev/gocast/issues/1739.


alexanderstephan (approved):
Looks good! I tried it out and it works as expected. 👍

PR #1826, opened 2026-05-08
fix: rtmp-proxy-URL

Mazi02:
### Motivation and Context
Fixing rtmp proxy url.

### Steps for Testing
1. login to admin
2. go to token management and create a token
3. you have to see in example url: https://proxy.example.com/live
grafik


joschahenningsen (approved):
Great! For deployment well also need to adapt the real config, let's do that after Sebi checked out the gh action

kordianbruck (approved):

PR #1872, opened 2026-07-03
fix(ffmpeg): vod download

### LINKED ISSUE #1869: Downloaded videos are not formated right
**Describe the bug**
The duration of the videos downloaded are always about 5 seconds, the videos are running though. It the vod conversion has a problem.

**To Reproduce**
1. Go any VoD of your choosing
2. Click on download
3. Play the video on any video player

**Expected behavior**
Correct timestamp

**Screenshots**

Image

Mazi02:
### Motivation and Context
Fixes vod download problem and optimizes it for streaming.

closes #1869


coolgirl19 (approved):
works again. video does not work in the player though, but probably some other issue.

PR #1879, opened 2026-07-08
Enh(docker): local docker services updated

Mazi02:
### Motivation and Context
Removing the duplicate docker compose file and updating the new one. This way we don't have two separate docker compose files and less confusion.

### Changes
 The complete docker compose file is now called "docker-compose.yml" and additionally has the Prometheus for local development setup. 
Also removed the worker from the new compose file, as we have migrated to runners now. 

### Testing

You can start the compose file in the terminal with `docker compose -f docker-compose.yml up --build` and go to `localhost:8081`.


docker-compose.yml
```diff
@@ -70,6 +88,81 @@ services:
...
+    environment:
+      - Token=abc
+      - Host=worker
+      - MainBase=tum-live
+      - VodURLTemplate=http://localhost:8089/vod/%s.mp4/playlist.m3u8
+      - LrzUploadUrl=http://vod-service:8089
+      - DEBUG-MODE=true
+    volumes:
+      - recordings:/recordings
+      - mass:/mass
```
joschahenningsen: This one we can drop, it's no longer in use

docker-compose.yml
```diff
@@ -18,44 +21,59 @@ services:
...
       - DBHOST=db
       - TZ=Europe/Berlin
-  worker:
-    container_name: tum-live-worker
+    ports:
+      - "8081:8081"
+    # extra_hosts:       # Only needed if you use wsl
+    #   - "localhost:host-gateway"
+    healthcheck:
+      test: [ "CMD-SHELL", "nc -z localhost 8081 || exit 1" ]
```
joschahenningsen, line 29: nice

docker-compose.yml
Mazi02: Got it. Will do. I was also working a little with runners. I am working on implementing grpcui into it for the development. Opinions?

joschahenningsen: What does that do? I'd like to keep the compose file close to what's required for prod deployment

Mazi02: OK then we are scratching that one.

kordianbruck (approved):

kordianbruck:
Why no nicer PR description? 🙈

Mazi02:
Thanks for pointing that out! Done. Description has been updated.

PR #1168, opened 2023-09-21
Feature/add lecture time change

mono424:
### Motivation and Context
No possibility to change date and time of a lecture.

### Description
- Added Timing Fields to Lectures
- Added Logs to changeSet
- Added Computed Properties

### Steps for Testing

Prerequisites:
- 1 Lecturer

1. Log in
2. Navigate to a Course Management Page
3. Change Date Time of a Lecture
4. Reload Page
5. Check if Time Date is still changed
6. Check if other settings can be changed too.

### Screenshots

https://github.com/TUM-Dev/gocast/assets/17506411/8aeb079d-7fa6-4130-b1b3-127d210f9fa1

Bildschirmfoto 2023-09-27 um 22 03 55

PR #1891, opened 2026-07-19
feat(admin): Edit lecture/series times

joschahenningsen:
Allow updating lecture and lecture series start times 

supersedes #1168


mock_dao/runner.go
```diff
@@ -99,19 +99,19 @@ func (mr *MockRunnerDaoMockRecorder) GetAll(arg0 any) *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetAll", reflect.TypeOf((*MockRunnerDao)(nil).GetAll), arg0)
 }
 
-// GetAvailable mocks base method.
```
joschahenningsen, line 102: This was re-generated, missed that previously

PR #1910, opened 2026-08-07
fixes: Runner stream state inconsistencies

joschahenningsen:
See individual commits for details.


cmd/tumlive/main.go
```diff
@@ -285,6 +285,8 @@ func initCron(logger *slog.Logger, m *runner_manager.Manager) {
 	_ = tools.Cron.AddFunc("exportToMeili", tools.NewMeiliExporter(daoWrapper).Export, "30 4 * * *")
 	// fetch live stream previews
 	_ = tools.Cron.AddFunc("fetchLivePreviews", api.FetchLivePreviews(daoWrapper), "*/1 * * * *")
+	// reap streams stuck in live state due to runner crash or lost notifications
+	_ = tools.Cron.AddFunc("reapStaleStreams", m.ReapStaleStreams, "*/1 * * * *")
```
kordianbruck: Can we reduce the frequency here? should suffice to run this every 30 minutes only during the day, right?

kordianbruck (approved):
Overall looks good. I'm worried that this is risky in the sense of stopping properly working streams, but you added a decent amount of logging. The entire setup with job count is a rough abstraction for "is this stuff working properly", but it might get us there. IDK how much of this change I actually understand, but I would've hoped there was an easy path to not use mutex'es (but I assume that's fundamentally how one runner can process multiple streams).

No integration tests of course means you've properly smoke tested this yourself, right? 😉

cmd/tumlive/main.go
joschahenningsen: for simplicity let's run this at night still, increased the frequency to run every 30 mins (at 01 and 31).

joschahenningsen:
> you've properly smoke tested this yourself, right 

Yessir :saluting_face: 
I'll ask my good friend claude for an integrations-test approach sometime. For now that's out of scope.

joschahenningsen:
Thanks for the review @kordianbruck :heart:

PR #1921, opened 2026-08-13
fix(runner): Handling the early end of a stream

Mazi02:
### Motivation and Context
If a lecturer ends the stream early and not discard it, it will not create a VoD.

### Steps for Testing

1. Create a stream and end it early.
2. The runner should create the VoD.

### DIFF
--- a/pkg/runner_manager/manager.go
+++ b/pkg/runner_manager/manager.go
@@ -763,6 +763,11 @@ func (m *Manager) endRunnerJob(ctx context.Context, job model.StreamRunnerJob, d
 		DiscardVod: ptr.Take(discardVoD),
 	})
 	if err != nil {
+		// a job the runner doesn't know is already over, which is what we want anyway
+		if status.Code(err) == codes.NotFound {
+			m.logger.Info("runner does not know job, considering it ended", "runner", job.RunnerHostname, "job", job.JobID)
+			return nil
+		}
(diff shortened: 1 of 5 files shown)


kordianbruck:
Closing this in favour of #1970, which covers both halves of what this branch set out to fix.

The VoD-on-early-end bug you describe here is the same one #1970 fixes: `RequestStreamEnd` cancels the job context, the VoD actions inherited it, and `exec.CommandContext` returns `context.Canceled` straight from `Start`, so `MkVOD`'s ffmpeg never ran. The actions after the stream now run under a context of their own.

Your `DiscardRecording` approach is carried over as-is — the `.del-` marker that `livestreamCleanup` already looks for, passed to `RunAction` as its own action list so the choice stays visible at the call site in `RequestStream`. That design is yours and it was the right call; #1970 adds a two-hour grace period on the marker (a discard clicked by mistake stays recoverable, since skipping VoD creation leaves the recording as the only copy) plus tests in `discard_test.go`, `cleanup_test.go` and `runner_test.go`.

The reason I'm not rebasing this one instead: #1920 has since landed the surrounding discard logic on `dev` with the recording deliberately left on disk, so the rebase conflicts in `runner/runner.go` and `runner/handlers.go` aren't mechanical — resolving them would mean rebuilding #1970's change by hand. Easier to land #1970 and keep the credit in the history.

Thanks for digging this out — please do review #1970 if you have a moment, you know this path better than anyone.

PR #1929, opened 2026-08-29
perf(courses): load and derive only what a course listing shows

coolgirl19:
The public course listing was the slowest page we serve, for two reasons that compound:
it loaded every lecture of every course in the semester, and it signed a playlist URL
for each of the two lectures it ends up showing.

### Load only the lectures a listing shows

A listing renders a course, the date of its most recent recording and the date of its
next lecture — two rows out of a semester's worth. `GetPublicCourses` and
`GetPublicAndLoggedInCourses` now preload only those: the earliest lecture still to
finish, which the next lecture is picked from, and the latest recording.

Both are looked up per `(course_id, private)`, correlated against the course being
filtered rather than grouped over the table. A course administrator is shown its private
lectures and everyone else is not, so the latest public recording has to outlive a
private one recorded after it, and a private lecture earlier than the next public one
must not take the row and leave everyone else with no next lecture at all; which of the
two a caller is given stays where it was, in `model.Course.GetLastRecording` and
`GetNextLecture`. A lecture that has started but not ended is still the next lecture
there is, so the upcoming side selects on `end` and takes the earliest `start` among
them.

`dao/courses_test.go` asserts the SQL this builds against a gorm dry-run connection.
The conditions here fail quietly — a course simply loses its recording — so they are
worth pinning; `frontend/e2e/visibility.spec.ts` covers the rendered page.

### Don't derive the playback payload for a listing

`ParseCourseSummaryToProto` answers with everything `ParseCourseToProto` does except
the signed playlist URLs and the downloadable files of the two derived lectures. Signing
a playlist is an RSA signature per URL, paid once per course in the response, and no
listing consumer reads either field — the player and the course page fetch the course
themselves.

The anonymous listing holds nothing derived for a caller, so it is parsed once per
semester and served from the cache after that. It is held for as long as the course rows
underneath it, so a lecture that has just been recorded is not kept out of the listing
for any longer than it already is.

### DIFF
--- a/apiv2/helpers/parser.go
+++ b/apiv2/helpers/parser.go
@@ -55,12 +55,23 @@ func ParseBookmarkToProto(b model.Bookmark) *protobuf.Bookmark {
 	}
 }
 
+// ParseCourseSummaryToProto converts a course to the reduced representation used
+// on list pages. It intentionally skips signed playlist URLs and download payloads,
+// because those are only needed on player/detail pages.
+func ParseCourseSummaryToProto(c model.Course, u *model.User) *protobuf.Course {
+	return parseCourseToProto(c, u, false, false)
+}
+
 // ParseCourseToProto converts a Course model to its protobuf representation.
 //
 // Everything derived here is derived for u: the private lectures of a course the
 // caller does not administer are left out of the last recording and the next lecture,
 // and the pin is the caller's own.
 func ParseCourseToProto(c model.Course, u *model.User) *protobuf.Course {
+	return parseCourseToProto(c, u, true, true)
+}
+
+func parseCourseToProto(c model.Course, u *model.User, signPlaylists bool, includeDownloads bool) *protobuf.Course {
 	course := &protobuf.Course{
 		Id:   uint32(c.ID),
 		Name: c.Name,
@@ -87,10 +98,10 @@ func ParseCourseToProto(c model.Course, u *model.User) *protobuf.Course {
 	// absent rather than sent as a stream with id 0, which every caller would then
 	// have to know to test for — as the Alpine start page did.
 	if last := c.GetLastRecording(u); last.ID != 0 {
-		course.LastRecording = ParseStreamToProto(*last, c, u)
+		course.LastRecording = parseStreamToProto(*last, c, u, signPlaylists, includeDownloads)
 	}
 	if next := c.GetNextLecture(u); next.ID != 0 {
-		course.NextLecture = ParseStreamToProto(*next, c, u)
+		course.NextLecture = parseStreamToProto(*next, c, u, signPlaylists, includeDownloads)
 	}
 
 	return course
@@ -122,9 +133,19 @@ func ParseSemesterToProto(semester model.Semester) *protobuf.Semester {
 
 // ParseStreamToProto converts a Stream model to its protobuf representation.
 func ParseStreamToProto(stream model.Stream, course model.Course, user *model.User) *protobuf.Stream {
+	return parseStreamToProto(stream, course, user, true, true)
+}
+
+func ParseStreamSummaryToProto(stream model.Stream, course model.Course, user *model.User) *protobuf.Stream {
+	return parseStreamToProto(stream, course, user, false, false)
+}
+
+func parseStreamToProto(stream model.Stream, course model.Course, user *model.User, signPlaylists bool, includeDownloads bool) *protobuf.Stream {
 	liveNow := stream.LiveNowTimestamp.After(time.Now())
 
-	_ = tools.SetSignedPlaylists(&stream, user, course.DownloadsEnabled)
+	if signPlaylists {
+		_ = tools.SetSignedPlaylists(&stream, user, course.DownloadsEnabled)
+	}
 
 	s := &protobuf.Stream{
 		Id:               uint32(stream.ID),
@@ -169,7 +190,7 @@ func ParseStreamToProto(stream model.Stream, course model.Course, user *model.Us
 		s.Duration = uint32(duration)
 	}
 
-	if course.DownloadsEnabled {
+	if includeDownloads && course.DownloadsEnabled {
 		for _, download := range stream.GetVodFiles() {
 			s.Downloads = append(s.Downloads, ParseDownloadToProto(download))
 		}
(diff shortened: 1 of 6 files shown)


apiv2/server/course.go
```diff
@@ -119,7 +129,15 @@ func (a *API) GetPublicCourses(ctx context.Context, req *protobuf.GetPublicCours
...
-		resp[i] = h.ParseCourseToProto(course, user)
+		if user == nil {
+			resp[i] = h.ParseCourseSummaryToProto(course, user)
+		} else {
+			resp[i] = h.ParseCourseToProto(course, user)
+		}
+	}
+
+	if user == nil {
+		dao.Cache.SetWithTTL(fmt.Sprintf("publicCoursesSummary-%d-%s", year, term), resp, 1, 6*time.Hour)
```
joschahenningsen, line 140: here I'd prefer a minute just to shield us from spikes during hot hours when large streams start.

dao/courses.go
```diff
@@ -50,6 +50,17 @@ type CoursesDaoImpl struct {
 	usersDao UsersDao
 }
 
+func publicCourseStreamFilter() func(*gorm.DB) *gorm.DB {
+	latestRecording := DB.Model(&model.Stream{}).
+		Select("MAX(start)").
+		Where("course_id = streams.course_id AND recording = ?", true)
+
+	return func(db *gorm.DB) *gorm.DB {
+		return db.Where("(recording = ? AND start = (?)) OR start > NOW()", true, latestRecording).
```
joschahenningsen, line 59: This will yield all upcoming streams, no?

dao/courses_test.go
```diff
@@ -0,0 +1 @@
+package dao
```
joschahenningsen, line 1: What's that file? Either add tests or delete

dao/courses.go
joschahenningsen: Maybe something like this works?

```
// Subquery 1: Latest past recording per course
latestRecording := DB.Model(&model.Stream{}).
    Select("MAX(start)").
    Where("course_id = streams.course_id AND recording = ? AND start <= NOW()", true)

// Subquery 2: Next upcoming stream per course
nextUpcoming := DB.Model(&model.Stream{}).
    Select("MIN(start)").
    Where("course_id = streams.course_id AND start > NOW()")

// Combined Filter
func publicCourseStreamFilter() func(*gorm.DB) *gorm.DB {
    return func(db *gorm.DB) *gorm.DB {
        return db.Where("start = (?) OR start = (?)", latestRecording, nextUpcoming).
            Order("start asc")
    }
}
```

joschahenningsen:
@coolgirl19 any progress on this?

dao/courses.go
kordianbruck: I think its addressed now as well?

apiv2/helpers/parser.go
```diff
@@ -122,9 +134,18 @@ func ParseSemesterToProto(semester model.Semester) *protobuf.Semester {
...
+// parseStreamToProto fills in the playback payload -- the signed playlist URLs and the
+// downloadable files -- only when includePlayback is set. Both are derived work a
+// listing pays for and never reads; see ParseCourseSummaryToProto.
+func parseStreamToProto(stream model.Stream, course model.Course, user *model.User, includePlayback bool) *protobuf.Stream {
 	liveNow := stream.LiveNowTimestamp.After(time.Now())
 
-	_ = tools.SetSignedPlaylists(&stream, user, course.DownloadsEnabled)
+	if includePlayback {
+		_ = tools.SetSignedPlaylists(&stream, user, course.DownloadsEnabled)
+	}
```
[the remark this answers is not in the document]
Mazi02: Fixed. The URLs now come from locals filled in on the same branch that signs them, so a summary can't answer with unsigned ones. Test covers all four fields and checks the full parse populates them first.

dao/courses.go
```diff
@@ -50,6 +50,30 @@ type CoursesDaoImpl struct {
...
+// has to survive a private one recorded after it. Which of the two a caller is given is
+// decided in model.Course.GetLastRecording, once the rows are in memory.
+func publicCourseStreamFilter(db *gorm.DB) *gorm.DB {
+	// Grouped rather than correlated: a subquery over `streams` cannot name the
+	// `streams` row being filtered, so a correlated condition here would compare the
+	// inner table against itself and yield the latest recording of any course.
+	latestRecordings := DB.Model(&model.Stream{}).
+		Select("course_id, private, MAX(start)").
+		Where("recording = ?", true).
+		Group("course_id, private")
```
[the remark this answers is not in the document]
Mazi02: Fixed by the indexed per-course lookup rather than a new index.

dao/courses.go
Mazi02: > This will yield all upcoming streams, no?

Yes it did. It now takes only the earliest lecture still to finish, per (course_id, private). Same fix as the recording side.

Mazi02: > Maybe something like this works?
> 
> ```
> // Subquery 1: Latest past recording per course
> latestRecording := DB.Model(&model.Stream{}).
>     Select("MAX(start)").
>     Where("course_id = streams.course_id AND recording = ? AND start <= NOW()", true)
> 
> // Subquery 2: Next upcoming stream per course
> nextUpcoming := DB.Model(&model.Stream{}).
>     Select("MIN(start)").
>     Where("course_id = streams.course_id AND start > NOW()")
> 
> // Combined Filter
> func publicCourseStreamFilter() func(*gorm.DB) *gorm.DB {
>     return func(db *gorm.DB) *gorm.DB {
>         return db.Where("start = (?) OR start = (?)", latestRecording, nextUpcoming).
>             Order("start asc")
>     }
> }
> ```

Went with your correlated version in the end. It needed the inner table aliased — unaliased, streams in the subquery names the subquery's own table rather than the outer row, which is what my earlier comment mistook for correlation not being possible. Both subqueries now hit fk_courses_streams.

Mazi02:
@joschahenningsen I didn't find anything else. Do you want to take another look at it?

coolgirl19:
sorry for the delay. thanks @Mazi02 for fixing this and it looks good from my perspective. @joschahenningsen can  you have another look please

PR #1940, opened 2026-09-06
feat(admin): Add external application registration and API keys

bassner:
### Motivation and Context

This PR starts the implementation discussed in [discussion #1752](https://github.com/TUM-Dev/gocast/discussions/1752). It lets instance administrators register external applications and manage their API keys.

The discussion proposed a new `ServiceType` account using the existing account and login mechanisms. This implementation instead uses a separate `Integration` record with an API key. An application does not need a human user account, browser session or user role. Keeping these identities separate avoids adding application-specific behavior to the existing human login flow.

Artemis is the first intended application. The interface is deliberately generic so other applications can use it too. GoCast makes no assumptions about Artemis or how another application organizes its content.

This is the first of six small PRs. Follow-up PRs add API authentication, course approval, approval-code exchange, course metadata and revocation through the API. Course access will use integration grants rather than making the application a course administrator. This PR does not grant access to courses or videos.

### Description

- Add an **Integrations** page for instance administrators.
- Register an application with a name and the return URL for the later approval flow.
- Generate an API key and show it once. Store only its SHA-256 hash.
- Copy, regenerate and revoke keys, and show whether a key is active or revoked.
- Accept HTTPS return URLs, with HTTP allowed for loopback hosts during local development.

### Steps for Testing

Prerequisites:

- A GoCast instance running this PR.
- An instance-administrator account and a lecturer or student account.
- **No Artemis instance or working callback server is required.** The return URL is only stored in this PR.

1. Sign in as an instance administrator and open **Administration → Integrations**.
2. Register `Example application` with `https://example.org/callback`. Check that it appears with an active key and that the generated key can be copied.
3. Reload. The application and active status should remain, but the key must no longer be shown.
4. Regenerate the key. Check that a different key is displayed and can be copied.
5. Revoke the key. Check that the displayed key is removed and the revoked status survives a reload. Regenerating a key should make it active again.
6. Try an empty name and `http://example.org/callback`. Both should be rejected. `http://localhost:3000/callback` should be accepted for local development.
7. Sign in as the lecturer or student. The Integrations menu entry should be hidden, and opening `/admin/integrations` directly should be forbidden.

API-key authentication and its validation endpoint come in PR 2. There is no course-approval or course-connection flow to test in PR 1.

Automated checks:

```sh
go test ./...
go vet ./...
npm --prefix web run lint
npm --prefix web run build
```

### Related Artemis work

The [Artemis draft PR](https://github.com/ls1intum/Artemis/pull/13693) uses the complete six-PR GoCast stack and must wait until all six are merged. It is linked for context only. Artemis is not required to review, test or merge this GoCast PR.


joschahenningsen:
Hey Patrick, thanks a lot for the contribution, I am on board with the changes and they look good. However, we're moving away from `api` in favor of our gRPC `apiv2`. It's proxied to REST with [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) so the client implementation shouldn't change a lot. On top of that you can generate your client from https://tum.live/api/v2/docs/.

If you want me to, I can port your changes to the new API, otherwise feel free to update the PR.

web/template/admin/admin_tabs/integrations.gohtml
```diff
@@ -0,0 +1,77 @@
+{{define "integrations"}}
```
joschahenningsen: This frontend page we should likely implement in our new vue SPA instead. [This PR](https://github.com/TUM-Dev/gocast/pull/1981) is a good example to highlight how to migrate a page from `/web` to `frontend`

joschahenningsen (commented):
Feature lgtm, one comment on the legacy frontend. Sorry you caught us in the midst of some migrations/clean up efforts 🙈

web/template/admin/admin_tabs/integrations.gohtml
bassner: Thanks! See 0f950eb -- does that work? :)

bassner:
@joschahenningsen I added the changes you requested. Can we proceed? :)

PR #1949, opened 2026-09-12
build(deps): migrate both packages to Tailwind CSS 4

kordianbruck:
Split out of #1797. Of the six npm majors Renovate grouped there, this is the only one that is not a version bump — it needs a CSS migration in both packages and a look at the rendered pages — so it is separated to keep #1797 mergeable on CI alone.

**Stacked on #1797.** Review that one first; this branch contains its two commits as its base.

## What the migration involves

- The PostCSS plugin and the CLI moved to their own packages, and the CLI lost its `build` subcommand. `autoprefixer` goes with them: v4 runs its output through Lightning CSS, which prefixes for us.
- The entry directives collapse to a single `@import "tailwindcss"`. Both JS configs are kept and reached through `@config`, which still honours `darkMode: "class"`, so the theme stays defined in one place per package.
- `@import "tailwindcss/custom-forms"` is dropped. It never resolved — v3 passes it straight through into the dist — so nothing has been loading it.
- v4 refuses `@apply` of a project-defined class, so the handful other rules build on (`btn`, `text-3/4/5`, `tum-live-bg*`, `tum-live-button*`) become `@utility`. `@apply … !important` becomes `@apply …!`.
- `watch.css` supplements `home.css` rather than standing alone, so it takes `@reference` and emits no preflight of its own.
- `mode: "jit"` and the `variants` block leave `web/tailwind.config.js` — both have been dead since v3, and v4 ignores them outright.

## Four things that would otherwise have changed the rendering silently

`@tailwindcss/upgrade` does not catch any of these, so they are done by hand:

1. **`theme.container`.** The `center` and `padding` options are no longer theme keys in v4, so the config block is reproduced as a `@utility container`. `@utility` extends the built-in container rather than replacing it, so the responsive max-widths still apply.
2. **Five scale names were reused at different values.** `shadow-sm`, `rounded-sm`, `backdrop-blur-sm` and `outline-none` are renamed to the `-xs` and `outline-hidden` spellings that keep the v3 rendering — verified by diffing compiled v3 and v4 output, not from the changelog. `outline-hidden` keeps the `forced-colors` fallback that made `outline-none` safe to use.
3. **`ring-opacity-5`** is gone in favour of the slash modifier.
4. **Preflight defaults.** Bare borders and placeholders now inherit the text colour instead of defaulting to gray-200 and gray-400. Around 150 elements use a bare `border`, so the v3 defaults are restored in a base layer. That block is commented for removal once those colours are explicit in the markup — it is a deliberate stopgap, not a permanent shim.

The remaining diff is `@tailwindcss/upgrade` output: `flex-grow` → `grow`, `overflow-ellipsis` → `text-ellipsis`, `!ml-0` → `ml-0!`, and bracket values that now have a scale entry. Those are pure spelling changes.

## Testing

`web` builds and lints; the frontend typechecks, builds and passes all 166 tests.

**Rendering is unverified.** Everything above is reasoning about compiled CSS — it wants someone to click through the server-rendered pages, in both light and dark mode, before this goes in. Worth a close look at the focus rings (`outline-hidden`), anything with a bare `border`, and the two pages that use `container` (`watch.gohtml`, `edit-course-by-token.gohtml`).


kordianbruck:
Retargeted back to `dev`: the workflows are gated on `pull_request: branches: [dev]`, so pointing this at the #1797 branch stopped CI from running on it entirely.

That means the diff here currently also contains #1797's commits. It will shrink to just the Tailwind commit once that one merges — review #1797 first either way.

kordianbruck:
Ran the stack locally and diffed every page against `dev`, screenshot by screenshot, light and dark. That turned up three real regressions — none of which CI can see, and one of which is much more serious than it looks.

### Cascade layers (the important one)

`@import "tailwindcss"` puts every utility in `@layer utilities`, and **a layered rule loses to an unlayered one regardless of specificity or source order**. Every vendor stylesheet this app links — video.js, videojs-seek-buttons, silvermine-airplay, fullcalendar, flatpickr, nouislider — is unlayered. So any Tailwind utility on an element those libraries also style was silently losing.

The visible symptom was the watch page: video.js sizes its player through `.video-comb-dimensions { width: 1920px }`, which beat `w-full` and blew the player 1920px wide inside a 1440px viewport, producing horizontal scrollbars. Confirmed from the matched-rule list rather than guessed:

```
.w-full                  width:100%    layer=utilities
.video-js                width:300px   layer=(none)
.video-comb-dimensions   width:1920px  layer=(none)   <- wins
```

The fix composes the entry by hand — theme and preflight layered, utilities not — which restores the v3 cascade exactly: utilities and vendor CSS compete on specificity and order again, while preflight still yields to project CSS. `#watchWrapper` goes back to 1438px, matching v3 to the pixel.

Worth flagging for its own sake: this is a general hazard of v4 in any app with unlayered third-party CSS, and the visible breakage was only the loudest instance.

### `min-w-screen` was never a real class

Not a Tailwind 3 utility, so it emitted nothing and the four admin templates using it rendered as though it were absent. v4 *does* define it, so it started applying `min-width: 100vw` and pushed those pages 320px past the viewport. Removed rather than translated — those pages have never actually had it.

### Default font stack

v4 drops `ui-sans-serif` and `system-ui` from the front of its default sans stack, which on Linux resolves to a different face and reflows text everywhere (long course titles stopped wrapping). The v3 stack is now pinned in both configs with a comment on how to adopt v4's instead. That felt like a visual decision worth making deliberately rather than inheriting from a dependency bump — happy to drop the pin if you'd rather take v4's default.

### Where it stands

Every page is now pixel-identical to `dev` except for Tailwind's built-in palette, which v4 moves from sRGB hex to OKLCH — blue-500 goes `rgb(65,133,246)` → `rgb(49,131,255)`, other built-ins shift comparably. The project's own configured colours are unchanged to the pixel. That one is v4 working as intended, so it is left alone.

Pixel diff against `dev`, after the fixes:

| Page | Before | After |
|---|---|---|
| admin-course (dark) | 18.44% | **0.00%** |
| admin (light/dark) | 5.58% | **0.00%** |
| watch / watch2 | size mismatch (1921px vs 1440px) | 0.22% (palette only) |
| home, course, login | 1.9–4.2% | 0.2–2.6% (palette only) |

Still unverified: mobile breakpoints, and the pages I could not reach with the starter fixture (lecture-cut, lecture-units, course-import, the schedule calendar — fullcalendar is unlayered too, so it is the most likely place for another instance of the first bug).

PR #1950, opened 2026-09-12
docs: add AGENTS.md and a local-testing skill

kordianbruck:
Writes down how to bring the stack up locally and how to prove a change in a browser, so the next person doesn't rediscover it.

The second half is the reason this exists. The Tailwind 4 regressions in #1949 — a 1920px video player forcing scrollbars, four admin pages 320px past the viewport, every line of text reflowed — were invisible to lint, unit tests, typecheck and all eleven CI jobs. They were only found by screenshotting each page against `dev` and diffing.

### Scripts

Three, runnable from anywhere in the tree:

- **`capture.mjs`** — logs in, shoots 14 pages in light and dark, and reports the two things a screenshot alone hides: horizontal overflow, and error pages. The 404 and permission templates render with **HTTP 200**, so a screenshot of one is indistinguishable from a pass. This bit me while writing the skill: `/schedule` isn't a route (`/admin` *is* the schedule page) and my first version cheerfully captured the 404 as a clean result.
- **`compare.mjs`** — pixel-diffs two runs, separating size mismatches (the layout moved) from colour drift.
- **`matched-rules.mjs`** — lists every rule setting a property *with its cascade layer*, which is what actually explains a wrong size when the utility class looks correct.

### What the prose records

The expensive lessons: layered utilities lose to unlayered vendor CSS regardless of specificity or order; a class that was dead in one major can come alive in the next (`min-w-screen`); default theme tokens move between majors (the font stack); utility scale names get reused at new values (`shadow-sm`, `rounded-sm`, `outline-none`) — and `@tailwindcss/upgrade` skips those codemods precisely when the new version is already installed.

Plus the mechanical traps: `npm ci` in `web/` is required to *compile* (`router.go` embeds `node_modules`), the SPA build is required or migrated routes silently fall back to their old templates, `make e2e_db` exists because three clocks have to agree on what day it is, the login field is the users table's `email` column, `@playwright/test` not `playwright`, and `go run` leaves a process that `pkill -f cmd/tumlive/main.go` does not match.

Every command and all three scripts were run against a live stack while writing this.

### DIFF
--- a/.claude/skills/local-testing/SKILL.md
+++ b/.claude/skills/local-testing/SKILL.md
@@ -1,0 +1,194 @@
+---
+name: local-testing
+description: Run GoCast's checks and bring the stack up locally to verify a change in a real browser — Go tests, frontend unit tests, lint, Playwright e2e, and screenshot-based visual regression against a baseline branch. Use when asked to test, run, or visually verify GoCast, when a change touches CSS, Tailwind config or templates, or when a dependency bump needs proving beyond CI.
+---
+
+# Testing GoCast locally
+
+CI covers compile-and-assert. It cannot see layout, colour, or anything that only
+appears once the app is in a browser — which is where the expensive regressions
+live. This skill covers both halves.
+
+## The fast checks
+
+Run these first; they need no database and catch most things.
+
+```bash
+go test -race ./...                  # or: make test  (also runs frontend unit tests)
+npm --prefix frontend test           # 166 vitest tests, ~2s
+npm --prefix frontend run typecheck  # vue-tsc
+npm --prefix web run lint            # eslint, flat config
+make lint                            # golangci-lint + the two above
+```
+
+`npm --prefix web run lint` lints `ts/` only — deliberately. `eslint .` also walks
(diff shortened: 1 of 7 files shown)


kordianbruck:
Added `AGENTS.md` to this branch — it and the skill are the same piece of work, and AGENTS.md points at the skill rather than repeating it.

AGENTS.md covers what isn't visible from the file tree: the five modules under `go.work`, which code is generated and from what, how migrations happen, and the conventions the history already follows. The section that earns its place is **the two frontends** — the template-to-SPA migration is half done by design, `spaRoutes` in `web/router.go` decides which half serves a URL, and moving a page needs an entry there *and* in the client router. The failure mode is quiet: skip the SPA build and every migrated route falls back to its old template, so you test the frontend being replaced and conclude it works.

Also recorded as traps: `npm ci` in `web/` being required to *compile* (not just for icons), the deliberate Tailwind layering from #1949, and `go run` leaving a process `pkill -f` doesn't match.

Every path, make target and factual claim was checked against the repo — which caught two of my own: there is no `main` branch (`dev` is the default), and only a marker file in `web/spa/` is tracked rather than the directory's contents.

PR #1954, opened 2026-09-12
removing pointer passing in AddDocumentsInBatches

Mazi02:
### Motivation and Context
meilisearch-go calls reflect.Value.Len on the argument, which panics on a
pointer, so the courses export crashed the server. The cron has no recover,
so the nightly exportToMeili job took the whole instance down with it.

### To Reproduce
Start the services and in admin, maintenance under the cron jobs choose the exportToMeili.  After this fix it shouldn't crash.


joschahenningsen (approved):
huh, did this break when we upgraded the meili client? I don't recall any crashes. 
Thanks!

kordianbruck (approved):

Mazi02:
It was from the go major update bump. I was working on that but it got merged, so I checked it and this one bug was still there.

PR #1956, opened 2026-09-12
fix(camera): don't panic on camera credentials without a colon

kordianbruck:
## Root cause

`protocol.MakeAuthenticatedRequest` did:

```go
userPassword := strings.Split(*auth, ":")
// ...
Password: userPassword[1],
```

`strings.Split` on a string with no colon returns a 1-element slice, so `userPassword[1]` is an index-out-of-range panic.

## How configuration reaches it

This is not reachable only from malformed input — it is reachable from ordinary configuration:

- `camera.Service.For` (`pkg/camera/controler.go`) sets `auth = ""` when the camera type is missing from the `auths` map.
- Every driver (`axis`, `panasonic`, `sony`) passes `&c.Auth`, never `nil`, so the `auth != nil` guard never helped.

So a lecture hall whose camera type has no configured credentials panicked on its first PTZ / snapshot / preset request. That code runs in a handler goroutine, and an **unrecovered panic in a goroutine takes the whole process down** — this was an availability bug, not a 500 on one request.

## Decision: error vs. unauthenticated

Two distinct cases, handled differently:

- **Empty credential → proceed unauthenticated.** This is the state the config path actually produces, and it is a legitimate one: a camera with no credentials configured may genuinely need an unauthenticated request. Erroring here would break those halls, and it is not a typo — it is the absence of a setting.
- **Non-empty credential without a colon → return an error.** There is no sensible interpretation of `"somesecret"` as a user/password pair. Silently downgrading it to an unauthenticated request would paper over a real misconfiguration and surface later as a confusing auth failure against the camera, so it is reported as a `400` with a clear message.

## Colon in the password

The old code also got this wrong in a second way: `strings.Split` yields more than two fields for `"user:pa:ss"`, and `userPassword[1]` silently truncated the password to `"pa"`. Now `strings.SplitN(*auth, ":", 2)` keeps everything after the first colon as the password. A colon is a legal password character, so this was a real (silent) auth failure for anyone using one.

## Tests

The two tests that pinned the buggy behaviour were rewritten to assert the correct behaviour rather than deleted:

- `pkg/camera/protocol/protocol_test.go`: the `recover()`-based "a credential without a colon panics" test is replaced with cases for an empty credential (unauthenticated request, no `Authorization` header), a credential with no colon (400 + error, no panic), an empty password (`"user:"`, accepted), a password containing a colon (`"user:pa:ss"`, verified end-to-end through a real digest challenge so truncation fails the test), and a normal credential.
- `pkg/camera/controler_test.go`: the matching factory test now also drives a real request with the empty auth the factory hands out, proving the config path no longer panics.

The unrelated "non-200 treated as success" pin in the same package is deliberately left untouched — that is a separate PR.

## Verification

`go test -race -count=2 ./pkg/...`, `go vet ./pkg/...` and `gofmt -l pkg/camera/` are all clean. The colon-in-password test was mutation-checked: reverting `SplitN` to `Split` makes it fail.

## Base branch

This targets #1955 (`test/cover-model-tools-camera`), whose tests pinned this bug. GitHub will retarget it to `dev` automatically once #1955 merges.

### DIFF
--- a/model/stream_test.go
+++ b/model/stream_test.go
@@ -1,0 +1,1058 @@
+package model
+
+import (
+	"database/sql"
+	"encoding/json"
+	"math"
+	"strings"
+	"testing"
+	"time"
+
+	"gorm.io/gorm"
+)
+
+// streamAt builds a stream spanning [now+startOffset, now+endOffset). Every time
+// assertion below is relative so the suite cannot rot or fail on a given date.
+func streamAt(startOffset, endOffset time.Duration) Stream {
+	now := time.Now()
+	return Stream{Start: now.Add(startOffset), End: now.Add(endOffset)}
+}
+
+func TestStreamIsPast(t *testing.T) {
+	tests := []struct {
+		name   string
+		stream Stream
+		want   bool
+	}{
+		{
+			name:   "a stream whose end time has passed is past",
+			stream: streamAt(-2*time.Hour, -time.Hour),
+			want:   true,
+		},
+		{
+			name:   "a stream ending just moments ago is past",
+			stream: streamAt(-2*time.Hour, -time.Second),
+			want:   true,
+		},
+		{
+			name:   "a stream still running is not past",
+			stream: streamAt(-time.Hour, time.Hour),
+			want:   false,
+		},
+		{
+			name:   "a stream yet to start is not past",
+			stream: streamAt(time.Hour, 2*time.Hour),
+			want:   false,
+		},
+		{
+			// Ended is set by the worker when a live stream stops early; it must win
+			// over the scheduled end time or the UI keeps showing a dead player.
+			name: "a stream flagged as ended is past even though its slot is still running",
+			stream: func() Stream {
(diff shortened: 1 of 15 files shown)


joschahenningsen (commented):
My take here: our configuration mechanism for cameras is garbage as is. We assume every camera of a specific brand has the same credentials. That made sense when we owned all cameras and configured them ourselves. Going forward I can see us using i.e. an axis camera installed in a lecture hall we don't manage.

Imo, we can avoid this class of error by 
- configuring user and password as separate columns for the cameras in either the `lecture_halls` table or a new `cameras` table with a fk on `lecture_halls`.
- dropping the credentials for cameras from `config`.

kordianbruck:
Superseded by #1965 (auth panic) and #1966 (non-200 handling). This branch was cut from the pre-squash base of #1955, so it carried that PR's whole test diff and conflicted with `dev`. Both commits are rebased onto current `dev` in the replacements; no content changed.

PR #1958, opened 2026-09-12
fix(model): stop decodeHash from mutating the global argon2 parameters

kordianbruck:
## Root cause

`decodeHash` in `model/user.go` parsed the cost parameters out of the stored hash it was
given and wrote them into the **package-level** `p`:

```go
_, err = fmt.Sscanf(vals[3], "m=%d,t=%d,p=%d", &p.memory, &p.iterations, &p.parallelism)
...
p.saltLength = uint32(len(salt))
p.keyLength  = uint32(len(hash))
```

`p` is also the policy `GenerateFromPassword` reads when hashing a **new** password, so
per-hash verification state was being written into global generation policy.

## Consequences

1. **Cost downgrade.** Verifying one user's login permanently changed the argon2
   parameters used for every password hashed afterwards by that process. A single legacy
   or deliberately weak hash in the `users.password` column downgraded new password
   hashes process-wide, and it persisted until restart.
2. **Data race.** An unsynchronised write to a global from every concurrent login.
   `go test -race` flags it once two logins decode at the same time.

## Fix

`decodeHash` now returns the parameters it parsed alongside the salt and derived key, and
`ComparePasswordAndHash` passes those to `argon2.IDKey`. The global `p` is never written
and stays purely the generation policy. Minimal, in the file's existing style.

## Existing stored hashes still verify

This is the regression a careless fix introduces: comparing a stored hash **requires** that
hash's own parameters, not the current policy — using `p` instead would lock out every user
whose hash was made with different ones. That is exactly why the values are parsed, and the
fix keeps it.

`TestComparePasswordAndHashUsesTheStoredParameters` proves it: it builds a hash with
deliberately weaker-than-policy parameters (m=32768, t=1, p=1), asserts it still verifies,
asserts a wrong password still fails against it, and asserts that a password hashed
afterwards is encoded with the configured policy rather than the legacy one. Reverting the
fix to use the global `p` for the comparison makes this test fail.

## Constant-time comparison

Checked: yes, already constant-time. `ComparePasswordAndHash` uses
`subtle.ConstantTimeCompare(hash, otherHash) == 1` for the final comparison; unchanged by
this PR.

## Tests

- `TestDecodeHashMutatesGlobalParameters` (which pinned the bug) is rewritten as
  `TestDecodeHashLeavesGlobalParametersUntouched`: decoding a weak hash must leave `p`
  identical and must return the hash's own parameters to the caller.
- `TestComparePasswordAndHashUsesTheStoredParameters` — the regression guard described above.
- `TestComparePasswordAndHashIsConcurrencySafe` — verifies three hashes with differing
  parameters from 24 goroutines, asserting every verification succeeds and `p` is unchanged.
  Under `-race` this pins the race itself.
- The `restoreArgonParams` helper existed only because `decodeHash` overwrote the global; it
  is now unused and removed, along with its call sites.

Verified clean: `go test -race -count=2 ./model/...`, `go vet ./model/...`, `gofmt -l model/`.

## Base branch

This was written against #1955 (`test/cover-model-tools-camera`), which added the test that
pinned this behaviour. #1955 has since merged, so it targets `dev` directly — no retarget
needed.

### DIFF
--- a/model/user.go
+++ b/model/user.go
@@ -395,14 +395,15 @@ func (u *User) ComparePasswordAndHash(password string) (match bool, err error) {
 		return false, nil
 	}
 	// Extract the parameters, salt and derived key from the encoded password
-	// hash.
-	salt, hash, err := decodeHash(u.Password)
+	// hash. The parameters are the ones this hash was created with, which may
+	// differ from the current generation policy in `p`.
+	hashParams, salt, hash, err := decodeHash(u.Password)
 	if err != nil {
 		return false, err
 	}
 
 	// Derive the key from the other password using the same parameters.
-	otherHash := argon2.IDKey([]byte(password), salt, p.iterations, p.memory, p.parallelism, p.keyLength)
+	otherHash := argon2.IDKey([]byte(password), salt, hashParams.iterations, hashParams.memory, hashParams.parallelism, hashParams.keyLength)
 
 	// Check that the contents of the hashed passwords are identical. Note
 	// that we are using the subtle.ConstantTimeCompare() function for this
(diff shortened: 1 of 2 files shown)


joschahenningsen (approved):
lgtm, I hope there's no users still using password based login, maybe worth exposing a metric?

PR #1961, opened 2026-09-12
fix(camera): treat a non-200 camera response as an error

kordianbruck:
## What was broken

`protocol.MakeAuthenticatedRequest` returned `(body, status, nil)` for **any** response it could read — 401 and 500 included. It handed the status back, and all three drivers dropped it on the floor (`_, _, err := ...; return err`).

Operator-visible consequences on real lecture-hall hardware:

- **`SetPreset` silently lied.** A camera that was unauthorized, erroring, or powered down but still answering reported a *successful preset change*. The operator saw success in the UI; the camera never moved, and nothing landed in the logs.
- **`TakeSnapshot` was worse.** The 401/500 HTML error page was written to disk as a `.jpg` and its filename returned as a valid snapshot, so a corrupt "image" flowed onward to everything that consumes lecture-hall snapshots.

## Where the check went, and why

**Centrally, in `MakeAuthenticatedRequest`.** Every caller gets it for free, there is exactly one definition of "the camera did not do what was asked", and no future driver can forget it. Per-driver checks would have meant three copies of the same three lines with three chances to diverge — and the original bug was precisely that each driver independently ignored the status.

Anything other than 200 now returns an error wrapping an exported sentinel:

```
camera 10.0.0.5:80: unexpected camera response status: 401 Unauthorized
```

The message names **the camera and the status**, because the operational question is always "which camera, and is this auth or is it down". `errors.Is(err, protocol.ErrUnexpectedStatus)` distinguishes "answered, but badly" from a transport failure. The error is built with `fmt.Errorf` + `%w`, matching the rest of the package.

The status is checked *before* the body is read, and **no body is returned alongside the error** — an error page is not a snapshot, and returning it is exactly how it ended up on disk.

## Signature

**Unchanged:** `(*bytes.Buffer, int, error)`. The status is still useful to callers that want to tell 401 from 404 without matching on a message, and it is asserted in existing tests in `pkg/camera`. Removing it would have churned every call site for no behavioural gain.

Sony's `GetPresets` was the only place that checked the status (`err != nil || status != http.StatusOK`); that check is now redundant and simplifies to `if err != nil { break }`. Its stop-at-the-first-missing-thumbnail behaviour, and therefore the number of presets a lecture hall shows, is unchanged.

## Other changes

- `SaveResponseBuffer` removes a partially written file when the write or close fails — a truncated snapshot on disk is indistinguishable from a whole one downstream — and rejects a `nil` buffer with an error instead of panicking.
- No ordering fix was needed in the drivers: all three already check the error before saving, so the central error is enough to stop the file being created at all.
- **The 200 path is byte-for-byte unchanged.** This hardware cannot be tested here, so the success path was deliberately left alone; a test pins that a 200 carrying a non-image body is still stored exactly as before.

## Tests

The tests that *pinned* the buggy behaviour were rewritten rather than deleted: "a failing camera is reported as success" is now "a failing camera returns an error", and the protocol test asserting a non-200 returns no error now asserts the error, the sentinel, the status code and the host in the message.

Per driver there is now coverage for 500, 401, an empty body, a malformed/HTML body, and a connection failure — plus the regression most worth locking down: **`TakeSnapshot` leaves no file on disk at all** when the camera errors (asserted by reading the output directory, not just by the returned filename), including for an offline camera.

`go test -race -count=2 ./pkg/...`, `go vet ./pkg/...` and `gofmt -l pkg/camera/` are all clean.

## Stacking

This is stacked on **#1956** (`fix/camera-auth-panic`), which is itself stacked on **#1955** (`test/cover-model-tools-camera`). It is based there because #1956 already edits `protocol.go` and its tests. GitHub will retarget this PR automatically as the stack merges down. The auth-parsing tests from #1956 are untouched.

### DIFF
--- a/pkg/camera/axis/axis_test.go
+++ b/pkg/camera/axis/axis_test.go
@@ -1,13 +1,17 @@
 package axis
 
 import (
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
+
+	"github.com/TUM-Dev/gocast/pkg/camera/protocol"
 )
 
 // recorder captures what the camera would have received, so the vendor CGI string can be
@@ -42,6 +46,24 @@ func offlineCam(t *testing.T) *AxisCam {
 	return NewAxisCam(strings.TrimPrefix(srv.URL, "http://"), "user:password")
 }
 
+// assertEmptyDir fails if anything was written into dir. A snapshot that failed must
+// leave nothing behind: whatever consumes lecture-hall snapshots cannot tell a stored
+// error page from a real JPEG.
+func assertEmptyDir(t *testing.T, dir string) {
+	t.Helper()
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		t.Fatalf("reading output directory: %v", err)
+	}
+	if len(entries) != 0 {
+		names := make([]string, 0, len(entries))
+		for _, e := range entries {
+			names = append(names, e.Name())
+		}
+		t.Errorf("output directory contains %v, want no file at all", names)
+	}
+}
+
 func TestSetPreset(t *testing.T) {
 	// Negative and out-of-range ids are included deliberately: the driver does no
 	// validation, it formats whatever it is given straight into the CGI query.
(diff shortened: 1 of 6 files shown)


joschahenningsen (approved):
lgtm!

PR #1964, opened 2026-09-12
fix(tools): stop Truncate blanking descriptions and over-promising with "..."

kordianbruck:
`tools/truncate.go` is the helper that shortens lecture and course descriptions for cards and list views. It had three correctness defects; this PR fixes all three.

## 1. `Truncate` blanked the entire text on unbalanced markup

The error branch read `_ = []byte("")` — a no-op assignment to the blank identifier — and then fell through to `return string(tr)` with `tr` still nil. So any input `TruncateHtml` rejected came back as `""`:

```go
Truncate("a</b>", 3) == ""
```

**User-visible impact:** one stray closing tag anywhere within the visible budget — routine in descriptions pasted out of Word or a mail client — silently erased the whole description. The card rendered empty, with no error anywhere to explain it.

**Chosen fallback: truncate the input as plain text with the markup stripped** (`TagExpr.ReplaceAll(input, nil)`, then truncate again). Rationale:

- Unbalanced markup means no *valid* truncated HTML can be produced from it, so returning the raw input is not an option: it would ship the broken markup straight into the page.
- The text itself is not in doubt, only the tags are. Dropping formatting loses styling; dropping the text loses the content. Losing formatting is strictly the smaller harm.
- The fallback re-enters the same function, so the length budget, rune-safety and entity handling are all still honoured — a stripped description is still truncated, not dumped whole.
- The second call cannot fail the same way: with every tag removed there is no closing tag left to mismatch. The `err != nil` guard on it returns `""` only as a belt-and-braces path.

`TruncateHtml` itself still returns `ErrUnbalancedTags` unchanged — callers that want to distinguish the cases still can.

## 2. The ellipsis was appended unconditionally

```go
Truncate("hello", 10) == "hello..."   // before
Truncate("hello", 10) == "hello"      // after
```

**User-visible impact:** every description shorter than the budget — most of them — advertised text that does not exist, inviting a click through to nothing more.

The subtlety is telling *"stopped because the budget ran out"* apart from *"stopped because the input ended"*, since both are true at once when the input is exactly `maxlen` visible characters. Rather than tracking why the scanner stopped, the fix asks the question that actually matters after the copy offset is final: **is there any input left beyond what was copied?**

```go
if bufPtr < len(buf) {
    output = append(output, []byte(ellipsis)...)
}
```

That is exact at the boundary: input of exactly `maxlen` visible characters ends with `bufPtr == len(buf)` and gets no ellipsis, while one character more leaves a byte over and does. It also stays correct for multi-byte trailing runes, entities and trailing whitespace, all of which move `bufPtr` in ways a "why did we stop" flag would have had to special-case.

## 3. Void elements and closing tags were matched case-sensitively

`voidElementTags` is lowercase and was compared with `==`, so `<BR>` was not recognised as a void element, went onto the close stack, and produced the invalid `</BR>` in the output. The same `!=` comparison matched a closing tag against its opener, so **`<B>text</b>` was wrongly rejected as `ErrUnbalancedTags`** — which, combined with bug 1, blanked the description outright. Both comparisons now use `strings.EqualFold`; HTML tag names are case-insensitive, so this is simply the correct rule.

**User-visible impact:** uppercase and mixed-case markup is routine in pasted rich text, and it produced either browser-confusing invalid HTML or (with bug 1) a vanished description.

## Tests

The existing assertions pinned the buggy behaviour, so they were rewritten to the corrected behaviour rather than removed, and coverage was added for:

- input shorter than / exactly equal to / one character longer than the budget — the equal case being the boundary bug 2 turns on;
- `<BR>`, ``, mixed-case `<Br/>`, and closing tags matching their opener in either case;
- unbalanced input now degrading to truncated plain text instead of vanishing, including a case where the degraded text is itself still truncated;
- normal truncation, entities, multi-byte runes and whitespace, all unchanged.

The parent PR's panic-regression tables (`TestTruncateHtmlStrayLessThan`, `TestTruncateHtmlTrailingMultiByteRune`) are kept intact — their expectations only lose the `"..."` that bug 2 was wrongly adding, since every one of those inputs fits inside its budget; what they regression-test (no panic, whole input returned, valid UTF-8) is untouched.

The property test is likewise kept and *strengthened*: instead of requiring an ellipsis on every non-empty output, it now requires the ellipsis to be present **exactly when** text was dropped — no ellipsis implies the output is the whole input plus closing tags, and an ellipsis implies the kept text is a strict prefix.

Verified clean: `go test -race -count=2 ./tools/`, `go vet ./tools/`, `gofmt -l tools/`.

## Stacking

This started from `fix/truncate-panics` (the two index-out-of-range panics in the same file) and touches the same two files. That branch carried an orphaned copy of `test: cover model, tools and the camera drivers`, whose content is already on `dev` as c8fe4634; it was dropped with `git rebase --onto origin/dev`, so this branch now contains exactly the parent's panic fix plus this commit and **targets `dev` directly**. The panic fixes are therefore included here; if `fix/truncate-panics` merges first, this rebases down to a single commit.

### DIFF
--- a/tools/truncate.go
+++ b/tools/truncate.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 	"unicode"
 	"unicode/utf8"
 )
@@ -129,7 +130,7 @@ func TruncateHtml(buf []byte, maxlen int, ellipsis string) ([]byte, error) {
 		// If this is a void element, do not count it as a start tag
 		isVoidElement := false
 		for _, voidElementTagName := range voidElementTags {
-			if tagName == voidElementTagName {
+			if strings.EqualFold(tagName, voidElementTagName) {
 				isVoidElement = true
 				break
 			}
@@ -146,7 +147,8 @@ func TruncateHtml(buf []byte, maxlen int, ellipsis string) ([]byte, error) {
 		} else {
 			// This is an end tag. First, check to make sure the end tag is
 			// matches what's on top of the stack.
-			if len(tagStack) == 0 || tagStack[len(tagStack)-1] != tagName {
+			// HTML tag names are case-insensitive, so <B>...</b> is balanced.
+			if len(tagStack) == 0 || !strings.EqualFold(tagStack[len(tagStack)-1], tagName) {
 				return nil, ErrUnbalancedTags
 			}
 
@@ -164,8 +166,13 @@ func TruncateHtml(buf []byte, maxlen int, ellipsis string) ([]byte, error) {
 	// Copy the desired input to the output buffer.
 	output := buf[0:bufPtr]
 
-	// Copy ellipsis
-	output = append(output, []byte(ellipsis)...)
+	// Copy the ellipsis, but only if something was actually cut. The scan stops
+	// either because the visible budget ran out or because the input ended, and at
+	// the boundary both happen at once -- so ask the only question that matters:
+	// is there any input left beyond what we copied?
+	if bufPtr < len(buf) {
+		output = append(output, []byte(ellipsis)...)
+	}
 
 	// Finally, create a closing tag for each tag in the stack.
 	for i := len(tagStack) - 1; i >= 0; i-- {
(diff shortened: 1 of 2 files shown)


kordianbruck:
duplicate of https://github.com/TUM-Dev/gocast/pull/1962

PR #1965, opened 2026-09-12
fix(camera): don't panic on camera credentials without a colon

kordianbruck:
## Root cause

`protocol.MakeAuthenticatedRequest` did:

```go
userPassword := strings.Split(*auth, ":")
// ...
Password: userPassword[1],
```

`strings.Split` on a string with no colon returns a 1-element slice, so `userPassword[1]` is an index-out-of-range panic.

## How configuration reaches it

This is not reachable only from malformed input — it is reachable from ordinary configuration:

- `camera.Service.For` (`pkg/camera/controler.go`) sets `auth = ""` when the camera type is missing from the `auths` map.
- Every driver (`axis`, `panasonic`, `sony`) passes `&c.Auth`, never `nil`, so the `auth != nil` guard never helped.

So a lecture hall whose camera type has no configured credentials panicked on its first PTZ / snapshot / preset request. That code runs in a handler goroutine, and an **unrecovered panic in a goroutine takes the whole process down** — this was an availability bug, not a 500 on one request.

## Decision: error vs. unauthenticated

Two distinct cases, handled differently:

- **Empty credential → proceed unauthenticated.** This is the state the config path actually produces, and it is a legitimate one: a camera with no credentials configured may genuinely need an unauthenticated request. Erroring here would break those halls, and it is not a typo — it is the absence of a setting.
- **Non-empty credential without a colon → return an error.** There is no sensible interpretation of `"somesecret"` as a user/password pair. Silently downgrading it to an unauthenticated request would paper over a real misconfiguration and surface later as a confusing auth failure against the camera, so it is reported as a `400` with a clear message.

## Colon in the password

The old code also got this wrong in a second way: `strings.Split` yields more than two fields for `"user:pa:ss"`, and `userPassword[1]` silently truncated the password to `"pa"`. Now `strings.SplitN(*auth, ":", 2)` keeps everything after the first colon as the password. A colon is a legal password character, so this was a real (silent) auth failure for anyone using one.

## Tests

The two tests that pinned the buggy behaviour were rewritten to assert the correct behaviour rather than deleted:

- `pkg/camera/protocol/protocol_test.go`: the `recover()`-based "a credential without a colon panics" test is replaced with cases for an empty credential (unauthenticated request, no `Authorization` header), a credential with no colon (400 + error, no panic), an empty password (`"user:"`, accepted), a password containing a colon (`"user:pa:ss"`, verified end-to-end through a real digest challenge so truncation fails the test), and a normal credential.
- `pkg/camera/controler_test.go`: the matching factory test now also drives a real request with the empty auth the factory hands out, proving the config path no longer panics.

The unrelated "non-200 treated as success" pin in the same package is deliberately left untouched — that is a separate PR.

## Verification

`go test -race -count=2 ./pkg/...`, `go vet ./pkg/...` and `gofmt -l pkg/camera/` are all clean. The colon-in-password test was mutation-checked: reverting `SplitN` to `Split` makes it fail.

## Base branch

This targets #1955 (`test/cover-model-tools-camera`), whose tests pinned this bug. GitHub will retarget it to `dev` automatically once #1955 merges.

---

Replaces #1956, which was branched off the pre-squash base of #1955 and so conflicted with `dev`. Same commit, rebased onto current `dev`.

### DIFF
--- a/pkg/camera/controler_test.go
+++ b/pkg/camera/controler_test.go
@@ -1,11 +1,14 @@
 package camera
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/TUM-Dev/gocast/model"
 	"github.com/TUM-Dev/gocast/pkg/camera/axis"
 	"github.com/TUM-Dev/gocast/pkg/camera/panasonic"
+	"github.com/TUM-Dev/gocast/pkg/camera/protocol"
 	"github.com/TUM-Dev/gocast/pkg/camera/sony"
 )
 
@@ -89,9 +92,10 @@ func TestServiceFor(t *testing.T) {
 		}
 	})
 
-	// BUG: a camera type with no configured credentials yields an empty auth string,
-	// which protocol.MakeAuthenticatedRequest then splits on ":" and indexes at 1 --
-	// panicking on the first request. The factory itself still succeeds.
+	// A camera type with no configured credentials yields an empty auth string. The factory
+	// deliberately succeeds: protocol.MakeAuthenticatedRequest treats an empty credential as
+	// "no credentials configured" and sends the request unauthenticated, which some halls
+	// legitimately need. It used to index the ":"-split at 1 and panic on the first request.
 	t.Run("a camera type without credentials yields an empty auth rather than an error", func(t *testing.T) {
 		c, err := NewService(map[model.CameraType]string{}).For("10.0.0.1", model.Axis)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


kordianbruck:
Carrying over @joschahenningsen's review from #1956, which this PR replaces (that branch was cut from the pre-squash base of #1955 and conflicted with `dev`; same commit, rebased):

> My take here: our configuration mechanism for cameras is garbage as is. We assume every camera of a specific brand has the same credentials. That made sense when we owned all cameras and configured them ourselves. Going forward I can see us using i.e. an axis camera installed in a lecture hall we don't manage.
>
> Imo, we can avoid this class of error by
> - configuring user and password as separate columns for the cameras in either the `lecture_halls` table or a new `cameras` table with a fk on `lecture_halls`.
> - dropping the credentials for cameras from `config`.

Agreed on the diagnosis, and it is actually a notch worse than "one credential per brand": `cmd/tumlive/main.go:141-149` maps **one** `Auths.CamAuth` onto both Panasonic and Axis, so those two brands share a single credential across every hall. There is already a `// todo: per-camera auths` on line 143.

That shape is also what makes the crash reachable. When `CamAuth` is empty the map entry is never written, `Service.For` falls back to `auth = ""`, and every driver passes `&c.Auth` — never nil — into `MakeAuthenticatedRequest`, which split on `:` and indexed `[1]` unconditionally. So a hall whose brand has no configured credential panicked a goroutine on the first PTZ request, which takes the process down rather than returning a 500.

**On sequencing:** I'd rather not fold the schema change into this PR. The guard is worth having on its own merits — per-camera credentials are operator-entered strings, so a value with no colon stays reachable after the refactor, and the parser should not panic on it either way. The split also caught a second, silent bug in the same two lines: `strings.Split(auth, ":")` + `[1]` truncated `user:pa:ss` to password `pa`, so any password containing a colon was being sent wrong. That is a live correctness bug independent of where the credentials are stored, and it is pinned by a test that issues a real Digest challenge and recomputes the expected MD5.

**On the schema, for whoever picks it up:** columns on `lecture_halls` look like the better first move over a new `cameras` table. `CameraIP` and `CameraType` already live there and already assume one camera per hall, so `camera_user` / `camera_password` sit next to them for a single additive migration with no call sites to move. A `cameras` table with an FK is the more flexible end state, but it means migrating `CameraIP`/`CameraType` out of `lecture_halls` too and touching every reader of them — worth doing when there is an actual multi-camera hall to justify it, rather than pre-emptively.

The change itself would be roughly: additive migration in `dao/migrations/`, the two fields on `model.LectureHall`, admin UI fields on the lecture-hall edit form, `Service.For` taking the credential from the hall instead of `map[model.CameraType]string`, and finally dropping `Auths.CamAuth` / `Auths.CamAuthSony` from `tools/config.go` once nothing reads them. Happy to open that as a follow-up if you want it queued.

joschahenningsen (approved):
:shipit:

PR #1968, opened 2026-09-12
fix(security): stop course names from steering filesystem paths

kordianbruck:
Reported by Noah Mathern: course names are unsanitized and are used to build filesystem paths, so a course named `../../escaped` writes attachments outside the mass storage directory.

Confirmed on `dev`, and found the same pattern in two more places.

## What was wrong

`course.Name` is free text — only `Slug` is validated (`Course.BeforeSave`). Three places used the raw name as a directory component:

| Site | Effect |
| --- | --- |
| `api/stream.go` `newAttachment` | upload lands outside `Paths.Mass` (the reported issue) |
| `api/stream.go` `putCustomLiveThumbnail` | same, for custom thumbnails |
| `worker/api/api_server.go` `GenerateSectionImages` | builds `<StorageDir>/<CourseName>/<Year>.<Term>/sections` and calls `os.RemoveAll` on it — a recursive **delete** outside the storage root |

Impact is limited: it needs `PermLecture` or course-admin rights, the basename is a UUID (`filepath.Ext` stops at separators, so no targeted overwrite), and the `RemoveAll` target always ends in `<year>.<S|W>/sections`. Still a write and a delete escaping the storage root, plus `os.ModePerm` creating 0777 directories wherever they land.

## The fix

New `safepath` package:

- `Component` neutralizes separators, control characters, and the `.`/`..` segments, and never returns an empty string.
- `JoinInRoot` sanitizes each component and then verifies the joined path is still inside the root — a backstop so a gap in one half doesn't become a write outside the root.

Used at all three sites. `MkdirAll` on those lines goes from `os.ModePerm` to `0o755`.

The worker is a separate Go module and can't import the main module, so it carries a copy (`worker/safepath`) with a note to keep them in sync.

## On validating the name

`Course.BeforeSave` now rejects empty, control-character and `.`/`..` names. It deliberately **does not** reject path separators: course names come from TUMOnline and legitimately contain them (`Analysis I/II`), so rejecting `/` would block valid imports and make existing courses unsaveable. Separators are handled at path construction instead, where such a name becomes one directory level (`Analysis I_II`).

Existing on-disk layout is unchanged for benign names, and stored `file.Path` values are absolute, so already-uploaded files keep resolving.

## Tests

- `tools/safepath` and `worker/safepath` unit tests.
- `TestValidateCourseName`.
- `TestAttachmentCourseNameTraversal` and `TestCustomThumbnailCourseNameTraversal` drive the real handlers with a malicious course name. Both were verified to fail on the unpatched code.

Full `go test ./...` passes in both modules.

### DIFF
--- a/model/course.go
+++ b/model/course.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"regexp"
+	"strings"
 	"time"
 
 	"gorm.io/gorm"
@@ -353,6 +354,34 @@ var courseSlugRegex = regexp.MustCompile(`^[a-zA-Z0-9äöüÄÖÜ\-_]{1,150}$`)
 func (c *Course) BeforeSave(tx *gorm.DB) (err error) {
 	if !courseSlugRegex.MatchString(c.Slug) {
 		return errors.New("invalid course slug")
+	}
+	if err := validateCourseName(c.Name); err != nil {
+		return err
+	}
+	return nil
+}
+
+// validateCourseName rejects course names that cannot be meant seriously.
+// Names are free text (they come from TUMOnline as well as from the create
+// course form) and may legitimately contain characters such as "/", so this
+// only rules out what no real course name contains. Everything that builds a
+// filesystem path from the name sanitizes it separately, see
+// tools/safepath.
+func validateCourseName(name string) error {
+	trimmed := strings.TrimSpace(name)
+	if trimmed == "" {
+		return errors.New("course name must not be empty")
+	}
+	if len(name) > 255 {
+		return errors.New("course name is too long")
+	}
+	if trimmed == "." || trimmed == ".." {
+		return errors.New("invalid course name")
+	}
+	for _, r := range name {
+		if r == 0 || r < 0x20 || r == 0x7f {
+			return errors.New("course name must not contain control characters")
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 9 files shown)


kordianbruck:
Kudos to Noah Mathern for reporting this!

worker/api/api_server.go
```diff
@@ -98,14 +99,22 @@ func (s server) GenerateLivePreview(ctx context.Context, request *pb.LivePreview
 }
 
 func (s server) GenerateSectionImages(ctx context.Context, request *pb.GenerateSectionImageRequest) (*pb.GenerateSectionImageResponse, error) {
```
joschahenningsen, line 101: Instead of touching worker code, we should add the missing functionality to `runner` and deprecate/delete this handler, the worker is no longer being worked on and most of its features are migrated to the runners.

model/course.go
```diff
@@ -354,6 +355,31 @@ func (c *Course) BeforeSave(tx *gorm.DB) (err error) {
 	if !courseSlugRegex.MatchString(c.Slug) {
 		return errors.New("invalid course slug")
 	}
+	return validateCourseName(c.Name)
```
joschahenningsen, line 358: This will cause all existing courses with names that are invalid to not be updatable going forward. Did you check our production database for that? (you can query it in grafana->explore)

Mazi02: ran it in granfana, it didn't regect anything. Sent you the query to check.

joschahenningsen: thanks, lgtm then!

joschahenningsen (approved):

worker/api/api_server.go
joschahenningsen: We can do this down the line though

PR #1969, opened 2026-09-12
feat(runner): generate video section images on runners

Mazi02:
Follows up on @joschahenningsen's review comment on #1968: section image generation should move to the runners rather than the worker being patched.

Section images are the last piece served only by the workers, and the one that built a filesystem path out of the course name. #1968 sanitizes that in the worker; this moves the feature to where the problem cannot occur.

## Why this is the better fix

The runners never build a storage path. An action produces bytes, sends them to gocast in a notification, and **gocast writes the file where it wants it**, the pattern `ThumbnailReady` already uses. Section images now work the same way:

```
/mass/sections/2025/10/500/1024_42.jpg
        └ YYYY/MM  └ courseID  └ streamID_sectionID
```

Built from ids only. The course name never reaches a path, so there is nothing to sanitize here.

## What is in this PR

| Piece | Location |
| --- | --- |
| `RequestSectionImages` RPC; sections nested in the request, start as `google.protobuf.Duration` | `runner/runner.proto` |
| `SectionImagesReadyNotification` | `runner/notifications.proto` |
| ffmpeg frame extraction, piped out as jpeg | `runner/pkg/actions/mksectionimages.go` |
| RPC handler | `runner/handlers.go` |
| Runner-first dispatch with worker fallback, writing the images, updating the sections | `pkg/runner_manager/sectionimages.go` |

**apiv1 footprint:** the two v1 handlers that trigger generation (`POST /api/stream/:id/sections`, `GET /api/stream/:id/thumb`) call `runner_manager.GenerateSectionImages` and pass the legacy worker call in as the fallback. That is the whole v1 change, per AGENTS.md; apiv2 has no write endpoints for sections yet, so there is no v2 caller to wire up instead.

Three deliberate differences from the worker implementation:

- **A section whose timestamp cannot be read is skipped** instead of failing the batch. The worker aborted everything on the first ffmpeg error, losing the sections that would have worked.
- **Filenames are deterministic**, so regenerating overwrites in place (the worker used UUIDs and `RemoveAll`'d the folder). The file row a section previously pointed at is deleted, otherwise every regeneration orphans rows.
- **Legacy files on disk are left alone.** Cleaning them up means acting on stored paths that, because of the bug #1968 fixed, may point outside mass storage. That belongs in a one-off migration with a containment check, not in this path.

## Rollout

Both workers and runners are in production, so gocast asks a runner first and falls back to the worker when none takes the job. `GenerateSectionImages` in the worker is untouched and stays until the migration is done; the path traversal in that fallback is covered by #1968.

## Tests

- `MkSectionImages`: context validation, the empty case, and two tests that run real ffmpeg against a generated test video and assert one jpeg per section (skipped when ffmpeg is absent).
- `saveSectionImages`: path layout and section wiring, replacement of the previous file row, and a check that nothing a runner sends can land outside mass storage.
- `GenerateSectionImages`: fallback without a runner manager, no-op for an empty request.

`go test ./...` and `golangci-lint run` pass in both modules.

Note for regenerating the protos: this used protoc 35.1 to match the version already recorded in the generated files, and the pre-commit hook reformats the output afterwards.

### DIFF
--- a/api/worker_grpc.go
+++ b/api/worker_grpc.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/TUM-Dev/gocast/dao"
 	"github.com/TUM-Dev/gocast/model"
+	"github.com/TUM-Dev/gocast/pkg/runner_manager"
 	"github.com/TUM-Dev/gocast/tools"
 	subtitle_proto "github.com/TUM-Dev/gocast/voice-service/pb"
 	"github.com/TUM-Dev/gocast/worker/pb"
@@ -1070,4 +1071,28 @@ func ServeWorkerGRPC(subtitleClient subtitle_proto.SubtitleGeneratorClient, subt
 			logger.Error("Can't serve grpc", "err", err)
 		}
 	}()
+}
+
+// generateVideoSectionImages generates the thumbnails for the given video sections.
+//
+// Section image generation is being moved from the workers to the runners: a runner
+// returns the images as bytes and gocast writes them itself, which keeps course
+// metadata out of the storage path entirely. Until every deployment runs runners,
+// this falls back to the worker when no runner can take the job.
+func generateVideoSectionImages(manager *runner_manager.Manager, daoWrapper dao.DaoWrapper, parameters *generateVideoSectionImagesParameters) error {
+	if len(parameters.sections) == 0 {
+		return nil
+	}
+	if manager != nil {
+		err := manager.RequestSectionImages(context.Background(), runner_manager.SectionImageRequest{
+			StreamID:    parameters.sections[0].StreamID,
+			PlaylistURL: parameters.playlistUrl,
+			Sections:    parameters.sections,
+		})
+		if err == nil {
+			return nil
+		}
+		logger.Warn("no runner took the section images job, falling back to a worker", "err", err)
+	}
+	return GenerateVideoSectionImages(daoWrapper, parameters)
 }
--- a/runner/runner.proto
+++ b/runner/runner.proto
@@ -12,6 +12,9 @@ service RunnerService {
   // Requests a stream from a lecture hall
   rpc RequestStream (StreamRequest) returns (StreamResponse) {}
   rpc RequestStreamEnd (StreamEndRequest) returns (StreamEndResponse) {}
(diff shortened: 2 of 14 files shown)


api/worker_grpc.go
```diff
@@ -30,6 +30,7 @@ import (
```
kordianbruck, line 30: Should we really still be adding new code to the apiv1?

runner/runner.proto
```diff
@@ -36,6 +39,24 @@ message StreamEndRequest {
 message StreamEndResponse {
 }
 
+// SectionTimestamp is the offset into a recording at which a section starts.
+message SectionTimestamp {
```
kordianbruck: Move this into a local message inside of SectionImageRequest

runner/runner.proto
```diff
@@ -36,6 +39,24 @@ message StreamEndRequest {
...
+  uint64 section_id = 1;
+  uint32 hours = 2;
+  uint32 minutes = 3;
+  uint32 seconds = 4;
```
kordianbruck: nit: Why not use the timestamp proto instead?

kordianbruck (approved):

api/worker_grpc.go
Mazi02: The v1 change is now just the two call sites. The only endpoints that trigger generation are `POST /api/stream/:id/sections` and `GET /api/stream/:id/thumb`, and apiv2 only has the read-only `getVideoSections`, so there is no v2 caller to use instead. The runner-first logic and the worker fallback now live in `runner_manager.GenerateSectionImages`, and `api/worker_grpc.go` is not touched anymore.

runner/runner.proto
Mazi02: Done, it is `SectionImageRequest.Section` now.

runner/runner.proto
Mazi02: Switched to a well-known type, but `google.protobuf.Duration` rather than `Timestamp`: the section start is an offset into the recording, not a point in time. The runner passes it to ffmpeg's `-ss` in seconds.

joschahenningsen:
pls do a rebase, then we're ready to go

PR #1970, opened 2026-09-12
fix(runner): create the VoD and clean up the recording after an ended stream

kordianbruck:
### Motivation and Context
Both bugs were found while reviewing #1920 and are older than it. They are what happens to
the recording after a stream was ended early, which #1920 deliberately left out of scope.
This was stacked on that branch and is rebased onto `dev` now that #1920 is merged.

### Description
**Ending a stream without discard never produced a VoD.** `RequestStreamEnd` cancels the job
context and the VoD actions inherited it. `exec.CommandContext` returns `context.Canceled`
from `Start` when the context is already done, so `MkVOD`'s ffmpeg never ran, `CheckVoD`
failed right after it, and the `.keep-` marker it writes on failure filed the recording
under `ERROR_PATH`. Ending a stream stops the capture; it does not mean the segments
captured until then should be thrown away. The actions after the stream now run under a
context of their own, the stream actions keep the cancellable one.

**Discarded recordings were never deleted.** A discarded recording carries neither a `.del-`
nor a `.keep-` marker and is not empty either, so `livestreamCleanup` never looks at it and
it stays on `SEGMENT_PATH` forever — the marker was only ever written by `CheckVoD`, which
the discard path skips. `DiscardRecording` writes the `.del-` marker that `livestreamCleanup`
looks for, dated two hours out. That grace period is the point: once VoD creation is skipped
the recording is the only copy left, so a discard clicked by mistake stays recoverable
instead of being deleted immediately. It is passed to `RunAction` as its own action list and
runs in place of the VoD actions, so the choice stays visible at the call site in
`RequestStream`.

### Steps for Testing
Prerequisites:
- 1 Lecturer
- 1 Livestream
- 1 Runner

1. Start a livestream and open `/api/stream/<streamID>/end`. A VoD is created from what was
   recorded until then and shows up on the stream page. On `dev` no VoD appears and the
   recording turns up under `ERROR_PATH` instead.
2. Start a second livestream and open `/api/stream/<streamID>/end?discard=true`. The runner
   logs `discarding recording, skipping VoD creation` and `marking discarded recording for
   deletion`, no VoD is created, and a `.del-<timestamp>` file appears in the recording
   directory under `SEGMENT_PATH`.
3. Set that timestamp into the past (`mv .del-<ts> .del-1`) → `livestreamCleanup` removes the
   directory within a minute. Without the edit it is kept for two hours.

Covered by unit tests as well: `TestRunActionDoesNotCancelActionsAfterTheStream` fails on the
old behaviour, and `TestDiscardedRecordingIsCleanedUpAfterTheGracePeriod` runs the marker
through `determineCleanableDirectories`.

### DIFF
--- a/runner/runner.go
+++ b/runner/runner.go
@@ -178,21 +178,30 @@ func (r *Runner) InitApiGrpc() {
 	}
 }
 
-// RunAction runs the actions in a in the background and returns the id of the created job.
-// The actions in a keep running after the job's context was cancelled, which is what lets
-// StreamEnd report the end of a stream that was stopped early. The VoD actions are skipped
-// entirely when the stream was ended with discardVod.
-func (r *Runner) RunAction(a, vod []actions.Action, data map[string]any, logger *slog.Logger) string {
-	// create new context to avoid cancellation on grpc request termination
-	c, cancel := context.WithCancel(context.Background())
+// RunAction runs the actions of a stream job in the background and returns the id of the
+// created job.
+//
+// The stream actions run under a context that RequestStreamEnd cancels to stop the capture
+// early. They keep running after that cancellation, which is what lets StreamEnd report the
+// end of a stream that was stopped early.
+//
+// Afterwards either the vod or the discard actions run, depending on whether the stream was
+// ended with discardVod. Both run under a context of their own: turning the segments that
+// were captured until the cancellation into a VoD is precisely what still has to happen
+// after a stream was ended early, so they must not inherit the cancelled stream context.
+func (r *Runner) RunAction(stream, vod, discard []actions.Action, data map[string]any, logger *slog.Logger) string {
+	// create new contexts to avoid cancellation on grpc request termination
+	streamCtx, endStream := context.WithCancel(context.Background())
+	afterCtx, endAfter := context.WithCancel(context.Background())
 	job := uuid.New().String()
 	r.JobCount <- 1
 	r.jobsMu.Lock()
-	r.jobs[job] = cancel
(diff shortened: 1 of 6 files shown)


runner/runner.go
```diff
@@ -178,34 +178,43 @@ func (r *Runner) InitApiGrpc() {
...
+// end of a stream that was stopped early.
+//
+// Afterwards either the vod or the discard actions run, depending on whether the stream was
+// ended with discardVod. Both run under a context of their own: turning the segments that
+// were captured until the cancellation into a VoD is precisely what still has to happen
+// after a stream was ended early, so they must not inherit the cancelled stream context.
+func (r *Runner) RunAction(stream, vod, discard []actions.Action, data map[string]any, logger *slog.Logger) string {
+	// create new contexts to avoid cancellation on grpc request termination
+	streamCtx, endStream := context.WithCancel(context.Background())
+	afterCtx, endAfter := context.WithCancel(context.Background())
```
Mazi02, line 195: Regression 2: forced shutdown doesn't stop post-processing

runner/runner.go
```diff
@@ -178,34 +178,43 @@ func (r *Runner) InitApiGrpc() {
...
 	job := uuid.New().String()
 	r.JobCount <- 1
 	r.jobsMu.Lock()
-	r.jobs[job] = cancel
+	r.jobs[job] = endStream
```
Mazi02, line 199: Regression 1: late discard is ignored and the VoD gets published

PR #1974, opened 2026-09-14
fix(web): end the proxy server in /live and add the missing slash bfore the slug

Mazi02:
### Description
Fixes two bugs: 
1. URL shown in the Token Management tab, when you create a new token, the server address is incomplete. 
2. URL shown in the lecture Management tab, the server url as you start a new stream.

### Steps for Testing
Look at the server URLs in the given tabs respectively.


web/template/admin/admin_tabs/token.gohtml
```diff
@@ -101,7 +101,7 @@
         <p class="border-t border-gray-100 dark:border-gray-500 my-4" x-show="scope === 'lecturer'"></span>
         <div class="bg-gray-100 dark:bg-gray-800 p-4 rounded-lg" x-show="scope === 'lecturer'">
                 <div class="mb-4">
-                    <div x-data="{ tab: 'OBS' }">
+                    <div x-data="{ tab: 'OBS', server: '{{ .Tokens.RtmpProxyURL }}'.replace(/\/+$/, '').replace(/\/live$/, '') + '/live' }">
```
joschahenningsen, line 104: This should be part of the config, check out https://github.com/TUM-Dev/TUM-Live-Ansible/blob/c34ae779b52e1b95c32b53745cbaaf6a7d824f35/roles/deploy/templates/configs/config.yaml.j2#L32

joschahenningsen (changes requested):

Mazi02:
Fix is in Ansible repo, not here.

PR #1987, opened 2026-09-17
feat(frontend): migrate token management to the SPA

kordianbruck:
### Motivation and Context
Part of the ongoing migration of admin pages from server-rendered Go templates to the Vue SPA (see AGENTS.md). This PR migrates `/admin/token`.

### Description
- **apiv2**: adds `listTokens`, `createToken`, `deleteToken` to `AdminService` in `apiv2/server/apiv2.proto`, implemented in `apiv2/server/token_admin.go`. All three are gated on `users.manage` (not `server.administer`, unlike most of `AdminService`) — this page is the only place a token can be issued or revoked at all, so it inherits the account pages' permission rather than the server one.
- New proto messages are prefixed `Token` (`Token`, `TokenSecret`, `ListTokensResponse`, `CreateTokenRequest`, `DeleteTokenRequest`) to avoid name collisions with the other pages migrating in parallel.
- **Security-sensitive behaviour preserved exactly**: a freshly created token's secret is returned once, in the response to `createToken` (a `TokenSecret` message, deliberately distinct from `Token` so no future field on `Token` can leak a secret into the listing). `listTokens` never returns a secret. The dao-level scoping in `GetAllTokens` (only your own tokens without `users.manage`) and the ownership check in `DeleteToken` (creator or `users.manage`) are both kept as defense-in-depth even though every caller of these RPCs already holds `users.manage` by policy — same as v1's checks, not assumed dead code.
- **Go tests**: `apiv2/server/token_admin_test.go` covers listing (including the rtmp proxy URL bundled into the response, matching the old `TokensData` struct), scope validation, the admin-scope permission check, secret-returned-once, and the delete ownership/not-found paths.
- **TS client**: `frontend/src/lib/tokens.ts` (+ `tokens.test.ts`), thin wrapper over the generated client, with the same "never expose a secret twice" contract documented at the top.
- **Vue view**: `frontend/src/views/admin/TokensView.vue`, wrapped in `AdminLayout`. Same fields as the old page (user/scope/last used/expires table, create form, one-time secret reveal with copy buttons, OBS/Zoom/Teams self-streaming instructions). The expiration date input is a native `datetime-local` field instead of the old page's flatpickr dependency — same optional-expiry behaviour, no added JS dependency in the SPA.
- **Routing**: `/admin/token` added to `spaRoutes` in `web/router.go`, and its route switched from `routes.AdminPage` to `registerPage(userAdminGroup, ...)`, matching the `users.manage` permission group it was already in. Added to `frontend/src/router/index.ts`. `AdminLayout.vue`'s "Token Management" entry flipped to `migrated: true`.
- **Deleted (legacy)**: `web/template/admin/admin_tabs/token.gohtml`, `web/ts/token-management.ts` (and its entry-point export in `web/ts/entry/admins.ts`), the `{{template "token" ...}}` block in `admin.gohtml`, the `TokensData` struct and its wiring in `web/admin.go`, and the v1 `createToken`/`deleteToken` handlers (+ their tests) in `api/token.go`.
- **Kept (legacy)**: `api/token.go`'s `fetchStreamKey` and its `/api/token/proxy/:token` route — that's called directly by the RTMP proxy, unrelated to this admin page, and had no other caller among create/delete.
- **e2e**: `frontend/e2e/tokens.spec.ts`, covering the SPA shell, listing the seeded token, the nav entry, permission refusals (student/lecturer/anonymous) on both the page and the API, that the list never carries a `token` field, and a full create → reveal-once → delete flow. Added a seeded token row (`tum-live-starter.sql`) and its description in `frontend/e2e/seed.ts`.

**Not run**: per the coordinating instructions for this parallel migration, I did not run the Playwright e2e suite or start the dev server (nine agents running concurrently would collide on ports/DB). The spec above is written to match the reference specs (`runners.spec.ts`, `info-pages-admin.spec.ts`) but is unverified against a live server.

### Steps for Testing
1. `pnpm install` in `web/`, `make spa` (or `pnpm --dir frontend build`) to build the SPA.
2. Run the server, sign in as an administrator (`users.manage`).
3. Navigate to `/admin/token`.
4. Confirm the token list renders, create a lecturer-scoped token, confirm the secret shows once with the OBS/Zoom/Teams instructions, reload the page and confirm the secret is not shown again, then delete a token and confirm it disappears from the list.
5. Sign in as a lecturer or student and confirm `/admin/token` and `/api/v2/admin/tokens` both refuse with 403.

### Screenshots
Not yet taken — see the note above about not starting the dev server in this environment. This PR still needs a screenshot of the migrated page before merging.


Mazi02 (commented):
Tested locally and the page works: token creation, the secret showing exactly once, the row never carrying it, all three OBS / Zoom / Teams tabs, the copy buttons and delete. The route is wired through `registerPage` with `nil` and the old template and `web/ts/token-management.ts` are gone. `rtmp_proxy_url` survives the move intact, which was my main worry given how that field has behaved before.

One thing to fix: `frontend/e2e/tokens.spec.ts` contains a literal NUL byte at line 141.

```js
await expect(row).not.toContainText(secret ?? "\x00");
```

Git classifies the file as binary because of it, so the whole new e2e test is unreviewable here. GitHub's API returns no patch for it at all, and `git diff --numstat` gives `-` `-` instead of line counts. Every other spec in `frontend/e2e/` is plain text.

The fallback is also unreachable: line 136 already asserts `expect(secret).toBeTruthy()`, so `??` never fires. It looks like it is only there to satisfy the `string | null` from `textContent()`, which a narrowing check does without costing the file its text status:

```js
if (!secret) throw new Error("no secret was shown");
await expect(row).not.toContainText(secret);
```

PR #1991, opened 2026-09-19
fix(docker):  repair the compose stack and add an opt-in SPA dev server

Mazi02:
### Motivation and Context

The local `docker compose` stack had three defects that made it unreliable for
development:

- **Meilisearch was never reachable.** `config.yaml` points at
  `http://localhost:7700`, which inside the container is the container itself.
  Every boot logged `dial tcp [::1]:7700: connect: connection refused` and search
  silently did nothing.
- **The database reseeded on every restart.** `--init-file` re-runs the starter
  dump each time the container starts, and the dump drops its own tables first,
  so any local state was lost on restart.
- **`ingest.tum.live` did not resolve**, so the runner could not pull
  self-streams from the ingest container. The workaround was a commented-out
  `extra_hosts` block that only worked on WSL.

Separately, there was no way to iterate on the Vue SPA without a full image
rebuild, because the built app is `//go:embed`-ed into the Go binary.

### Description

Fixes:

- Add a `MEILIHOST` env override in `tools/config.go`, mirroring the existing
  `DBHOST` one, and set it to `http://meili-internal:7700` in compose.
- Seed the database from `/docker-entrypoint-initdb.d/` instead of `--init-file`,
  so the dump is loaded once into an empty volume. Reset with
  `docker compose down -v`. Adds a healthcheck and gates `tum-live` on
  `service_healthy`.
- Give the ingest container an `ingest.tum.live` network alias, so the stock
  `ingestbase` resolves inside the network. Removes the `extra_hosts` workaround.
- Ignore `docs_v2/.pnpm-store`, which the docs service writes into the bind mount.

Feature:

- Add an opt-in `frontend-dev` service running Vite with hot module replacement,
  behind a `dev` profile so a plain `docker compose up` is unchanged.
  `vite.config.ts` takes the backend URL from `GOCAST_BACKEND` (inside the
  network `localhost` is the wrong host), listens on all interfaces, and uses
  `base: "/"` in dev — the router uses `createWebHistory()` with no base, so
  `/admin/users` has to resolve at exactly that path rather than under
  `/spa-assets/`.

### Steps for Testing

Prerequisites: none beyond Docker. The starter dump provides the users and courses.

1. `docker compose down -v && docker compose up -d --build`
   (the `-v` is required once: seeding now only runs against an empty volume)
2. `docker compose ps` — every service healthy, including `mariadb_container`
3. `docker compose logs tum-live | grep 7700` — no `connection refused` lines
4. `docker exec tum-live-runner getent hosts ingest.tum.live` — resolves
5. Log in at http://localhost:8081 as `admin` / `password` and open a course —
   data is present, and still present after `docker compose restart db`
6. Hot reload: `docker compose --profile dev up -d frontend-dev`, open
   http://localhost:5173/admin/users, edit a file under `frontend/src` and watch
   it update without a rebuild


docker-compose.yml
```diff
@@ -109,6 +108,10 @@ services:
     platform: linux/amd64
     ports:
       - "1935:1935"
+    networks:
+      default:
+        # The stock config's ingestbase, so the runner pulls self-streams from this container.
+        aliases: [ ingest.tum.live ]
```
joschahenningsen: likewise you should be able to just change this to `ingest` in the config

tools/config.go
```diff
@@ -95,6 +95,10 @@ func initConfig() {
 	if os.Getenv("DBHOST") != "" {
 		Cfg.Db.Host = os.Getenv("DBHOST")
 	}
+	// Same for meilisearch, which inside docker-compose is a service name, not localhost.
+	if host := os.Getenv("MEILIHOST"); host != "" && Cfg.Meili != nil {
+		Cfg.Meili.Host = host
```
joschahenningsen, line 100: Why not adjust the config for local development?

docker-compose.yml
Mazi02: My bad, thanks. Done  `ingestbase` is `rtmp://ingest/` now and the alias is gone. Turned out `Dockerfile-local` already sed'd it to exactly that, but the bind mount over `config.yaml` meant it never applied, so that line is gone too.

tools/config.go
Mazi02: Because `config.yaml` is read both inside the container and by `go run cmd/tumlive/main.go` on the host, and the two need different hosts  `meili-internal:7700` vs `localhost:7700` on the published port. It mirrors the `DBHOST` override right above it, where `db: host: localhost` stays in the config and compose sets the service name.
I can  just put `meili-internal` in the config if you prefer.

docker-compose.yml
```diff
@@ -128,14 +127,20 @@ services:
     container_name: mariadb_container
     image: mariadb
     environment:
-      MYSQL_ROOT_USER: root
-      MYSQL_ROOT_PASSWORD: example
+      MARIADB_ROOT_PASSWORD: example
       TZ: Europe/Berlin
-    command: --init-file /data/application/init.sql
```
kordianbruck, line 134: I think this was intentional, so to have a clean slate and motivate folks to add proper fixtures for local testing

docker-compose.yml
```diff
@@ -181,6 +186,36 @@ services:
     command:
       - "--config.file=/etc/prometheus/prometheus.yml"
       - "--storage.tsdb.path=/prometheus"
+  # Vite dev server with hot module replacement for the SPA.
+  #   docker compose --profile dev up -d frontend-dev
+  frontend-dev:
```
coolgirl19, line 191: if this one is frontend/src/views/admin/UsersView.vue edited in line 228, I am unable to see live update in the browser even after reloading.

coolgirl19 (changes requested):
Looks good except the remark below

